PR #2, opened 2016-01-21
Use newer Sisu Inject

jglick:
Seems to fix an error originally noted in https://github.com/jenkinsci/jenkins/pull/1982. (Appears reproducible only in plugins with dependencies on other plugins, I think.)

@reviewbybees esp. @andresrc

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -120,6 +120,11 @@ THE SOFTWARE.
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency> <!-- lib-jenkins-maven-embedder uses sisuInjectVersion=0.0.0.M2a which has uses ASM 3.0 which clashes with 4.0 on ClassVisitor -->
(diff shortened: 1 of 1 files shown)


andresrc:
To reproduce: get hello-word plugin, add dependency to scm-api and during tests:

```
org.jvnet.hudson.test.JellyTestSuiteBuilder$JellyTestSuite(org.jvnet.hudson.test.junit.FailedTest)  Time elapsed: 0.026 sec  <<< ERROR!
java.lang.IncompatibleClassChangeError: Implementing class
        at java.lang.ClassLoader.defineClass1(Native Method)
        at java.lang.ClassLoader.defineClass(ClassLoader.java:760)
        at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
```

Tested introducing the new dependency and ok

:bee:

recena:
:bee:

PR #5, opened 2016-01-22
Fixing 'unmappable character for encoding ASCII' ('...')

pjanouse:
I'm fiddling with this module for the first time due to JENKINS-32571 but building failed (use OpenJDK - fedora-2.5.3.0.fc20-x86_64 u71-b14) due to:
error: unmappable character for encoding ASCII

It seems the special character '...' is the root cause, so I've changed its occurrence to "...".


pjanouse:
PR#4 should fix the build issue.

jglick:
See #7.

PR #6, opened 2016-01-22
[FIXED JENKINS-32571] Preserve original security settings after calling CommandInvoker.invoke()

pjanouse:
[JENKINS-32571](https://issues.jenkins-ci.org/browse/JENKINS-32571)

Please review. Thx Pavel

### DIFF
--- a/src/main/java/hudson/cli/CLICommandInvoker.java
+++ b/src/main/java/hudson/cli/CLICommandInvoker.java
@@ -29,7 +29,11 @@ import hudson.model.User;
 import hudson.security.ACL;
 import hudson.security.AuthorizationStrategy;
 import hudson.security.Permission;
+import hudson.security.SecurityRealm;
 import hudson.security.SidACL;
+
+import org.acegisecurity.context.SecurityContext;
+import org.acegisecurity.context.SecurityContextHolder;
 
 import java.io.ByteArrayOutputStream;
 import java.io.InputStream;
@@ -44,6 +48,7 @@ import jenkins.model.Jenkins;
 import org.acegisecurity.acls.sid.PrincipalSid;
 import org.acegisecurity.acls.sid.Sid;
 
+import org.apache.maven.plugin.lifecycle.Execution;
 import org.hamcrest.Description;
 import org.hamcrest.TypeSafeMatcher;
 import org.jvnet.hudson.test.JenkinsRule;
@@ -58,6 +63,9 @@ public class CLICommandInvoker {
     private static final String username = "user";
     private final JenkinsRule rule;
     private final CLICommand command;
+    private SecurityRealm originalSecurityRealm = null;
+    private AuthorizationStrategy originalAuthorizationStrategy = null;
+    private SecurityContext originalSecurityContext = null;
 
     private InputStream stdin;
     private List<String> args = Collections.emptyList();
@@ -112,6 +120,8 @@ public class CLICommandInvoker {
 
     public Result invoke() {
 
+        Result result;
+        Error executionError = null;
         setAuth();
 
         final ByteArrayOutputStream out = new ByteArrayOutputStream();
@@ -121,7 +131,11 @@ public class CLICommandInvoker {
                 args, locale, stdin, new PrintStream(out), new PrintStream(err)
         );
 
-        return new Result(returnCode, out, err);
+        result = new Result(returnCode, out, err);
+
+        restoreAuth();
+
+        return result;
     }
 
     private static class GrantPermissions extends AuthorizationStrategy {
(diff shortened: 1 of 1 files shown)


pjanouse:
PR#4 should fix the build issue.

src/main/java/hudson/cli/CLICommandInvoker.java
```diff
@@ -121,7 +128,11 @@ public Result invoke() {
                 args, locale, stdin, new PrintStream(out), new PrintStream(err)
         );
 
-        return new Result(returnCode, out, err);
+        Result result = new Result(returnCode, out, err);
+
+        restoreAuth();
```
jglick: A little safer to do this in a `finally` block. `main` does catch `Exception`, though, so probably good enough (any `Error` is most likely going to fail the whole test case anyway).

jglick:
:+1:

jglick:
`git merge master && git push` to have CI check this.

pjanouse:
I've fixed calling the restoreAuth().

I'd like to explain more what is my intention in the a bit longer code. First - I'd like to distinct an error occurred in command.main(...) from the error occurred during restoreAuth().

If an issue occurred in the finally block (while restoreAuth() performed) and I'll only catch and report that, the original issue from command.main(...) (and probably an inconsistent state of system) is hidden.

The second option is don't catch an exception while performing restoreAuth() at all and left everything on the caller. On the other hand in that case the root cause can be overlooked and interchanged with the dependent one.

WDYT about this approach?

jglick:
> the original issue from `command.main(...)` is hidden

So? If there is an error in `restoreAuth`, the test is busted and needs to be fixed anyway.

(Java 7+ has a suppressed exception facility, though I think it is overkill here.)

src/main/java/hudson/cli/CLICommandInvoker.java
```diff
@@ -112,16 +120,35 @@ public Result invokeWithArgs(final String... args) {
...
+            executionError = e;
+            throw e;
+        } finally {
+            try {
+                restoreAuth();
+            } catch (Exception e) {
+                if (executionError != null) { // Log original Error is occurred
+                    throw new AssertionError("Restore authentization failed and execution failure occurred previously.", executionError);
+                } else {
+                    throw new AssertionError("Restore authentization failed.");
```
jglick: Now (in either case) you are losing `e`.

jglick:
Seems OK as is, though personally I would just simplify to

```java
try {
    setAuth();
    return …;
} finally {
    restoreAuth();
}
```

pjanouse:
Jesse,

I'm on the same page with you on "do a thing as simple as we can" however I can see two different points of view in this case:
- developer PoV
- test executor PoV

Your approach can be the best for the second one - if something went wrong, it doesn't matter so much what/where an issue occurred, sth. is broken and has to be fixed; period. On the other hand a developer can appreciate if he knows the root cause precisely, in the first shot and without any confusion. 

I see the CLICommandInvoker class primary as a helper for a developer - easily and rapidly develop a CLI test so as a developer I don't like the result produced by your version in this scenario:
- called invoke()
- exception is thrown in command.main()
- exception is thrown in restoreAuth()

In above case the root exception is eaten and JUNIT/Surefire reports only the second one. If the first exception is visible/reported it depends on the impl. code/logging, in many common cases it needn't to be reported anywhere and developer can be completely confused why such an exception was thrown.

Do you really think the above example is so rare exception that we don't need to deal it at all? I'm still inclining to solve it. If you have an idea how to implement that in a better way, please share your thought.

pjanouse:
I've thought about this a bit more, what about the original approach - not to envelope the code with try/catch block at all... if exception is thrown in command.main() fine, will be reported and further execution is unclean; it doesn't matter so much if we call restoreAuth() anyway. If the exception is thrown in restoreAuth(), it will be reported as well.

I mean code like: 
```java
public invoke() {
  setAuth();
  command.main();
  restoreAuth();
}
```

I think that this simple approach solves my objection. WDYT?

pjanouse:
I've returned to my original impl. which fixes above issue which I'd like to solve as well.

pjanouse:
Could someone review and if no objection merge this change please?

It effectively block e.g. JENKINS-32535.

Thank you, Pavel

src/main/java/hudson/cli/CLICommandInvoker.java
```diff
@@ -112,6 +120,8 @@ public Result invokeWithArgs(final String... args) {
 
     public Result invoke() {
 
+        Result result;
+        Error executionError = null;
```
jglick: Seems that `executionError` is unused, no?

Also I see no reason why `result` is being defined here. Would seem simpler for end of method to read

```java
restoreAuth();
return new Result(returnCode, out, err);
```

jglick:
> Do you really think the above example is so rare exception that we don't need to deal it at all?

Yes probably it is very rare.

BTW in `-source 7` you could make `setAuth` return some `AutoCloseable` and then use `try`-with-resources, or simply call `Throwable.addSuppressedException` directly. Anyway this is all probably overkill for such a minor bit of test code.

PR #7, opened 2016-01-22
Forgot to specify source encoding

jglick:
Supersedes #5.

@reviewbybees


abayer:
:bee:

rsandell:
:bee:

jtnord:
NBL do you also need the `project.reporting.outputEncoding` property so that javadoc is output in UTF-8 also?

jglick:
> so that javadoc is output in UTF-8 also?

Could, if anyone complains about it I guess. This was the urgent one—uncompilable by some people without it.

PR #8, opened 2016-01-23
Actively look for a plugin that failed to start.

kohsuke:
In particular, I had a situation where a plugin under test failed to start, which isn't obvious to diagnose in this case

@reviewbybees

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/PluginAutomaticTestBuilder.java
+++ b/src/main/java/org/jvnet/hudson/test/PluginAutomaticTestBuilder.java
@@ -23,6 +23,7 @@
  */
 package org.jvnet.hudson.test;
 
+import hudson.PluginManager.FailedPlugin;
 import hudson.PluginWrapper;
 import hudson.cli.CLICommand;
 import java.io.File;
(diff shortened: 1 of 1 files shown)


stephenc:
:bee:

src/main/java/org/jvnet/hudson/test/PluginAutomaticTestBuilder.java
```diff
@@ -86,8 +87,13 @@ public void testCliSanity() {
...
             String plugin = (String) params.get("artifactId");
             if (plugin != null) {
+                // did any plugin fail to start?
+                for (FailedPlugin fp : Jenkins.getInstance().getPluginManager().getFailedPlugins()) {
+                    throw new Error("Plugin "+fp.name+" failed to start", fp.cause);
+                }
+
                 PluginWrapper pw = Jenkins.getInstance().getPluginManager().getPlugin(plugin);
-                assertNotNull(pw);
+                assertNotNull(plugin+" failed to start", pw);
```
recena: `assertNotNull(plugin + " failed to start", pw);` in order to follow the same convention below

recena:
:bee:

oleg-nenashev:
:bee:

PR #9, opened 2016-01-31
Listen to connections on localhost only

jglick:
Otherwise the tested Jenkins instance could be accessed from other hosts doing port scanning.

@reviewbybees


recena:
:bee:

PR #16, opened 2016-03-11
Make dependency declaration less error prone

kohsuke:
If a plugin explicitly declares a dependency on core, war, or some combination of it, but without declaring a dependency to war-for-test, unit tests end up running with one version of the core jar file and another version of the war file. This results in a strange error that's hard to diagnose.

It's very unlikely that plugins using test harness just so happens to be based on 1.580.1, so it's better not to let this module brings in war-for-test, and instead let the caller declare it explicitly, like the plugin parent POM is doing.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -87,6 +87,11 @@ THE SOFTWARE.
           <artifactId>sshd</artifactId>
         </exclusion>
       </exclusions>
+      <!--
+        To ensure consistent set of core artifacts are used, force the users to declare
+        a dependency to war-for-test
+      -->
+      <optional>true</optional>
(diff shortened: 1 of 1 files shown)


kohsuke:
@reviewbybees

jglick:
:bee: though Maven tricks like this scare me a little.

Really I would like to kill `war-for-test` (huge waste of repository space) and just fix the plugin POM to copy the non-classpath dependency into `target/classes/`. I think this change is compatible with such a move.

andresrc:
:bee:

rsandell:
:bee: :+1:

PR #19, opened 2016-05-10
[JENKINS-34701] Allow customizing the WebAppContext

andresrc:
[JENKINS-34701](https://issues.jenkins-ci.org/browse/JENKINS-34701)

@reviewbybees

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -606,6 +606,21 @@ public class JenkinsRule implements TestRule, MethodRule, RootAction {
     }
 
     /**
+     * Prepares a {@link WebAppContext} to be used in thehosting environment to get the
+     * {@link javax.servlet.ServletContext} implementation that we need for testing.
+     */
+    protected WebAppContext createWebAppContext() throws Exception {
+        final WebAppContext context = new WebAppContext(WarExploder.getExplodedDir().getPath(), contextPath);
+        context.setClassLoader(getClass().getClassLoader());
+        context.setConfigurations(new Configuration[]{new WebXmlConfiguration()});
+        context.addBean(new NoListenerConfiguration(context));
+        context.setMimeTypes(MIME_TYPES);
+        context.getSecurityHandler().setLoginService(configureUserRealm());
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -606,6 +606,21 @@ public File getWebAppRoot() throws Exception {
     }
 
     /**
+     * Prepares a {@link WebAppContext} to be used in thehosting environment to get the
```
amuniz: Typo

amuniz:
:bee:

abayer:
:bee:

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -606,6 +606,21 @@ public File getWebAppRoot() throws Exception {
...
+     * {@link javax.servlet.ServletContext} implementation that we need for testing.
+     */
+    protected WebAppContext createWebAppContext() throws Exception {
+        final WebAppContext context = new WebAppContext(WarExploder.getExplodedDir().getPath(), contextPath);
+        context.setClassLoader(getClass().getClassLoader());
```
jglick: Generally speaking the `getClass().getClassLoader()` idiom is unsafe, since it could cause a subclass to behave differently merely because it was defined in a child `ClassLoader`, without even overriding anything at all.

Use `JenkinsRule.class.getClassLoader()` unless you really have a reason to look for the loader of a rule subclass specifically.

jglick:
:bee: in principle but I suppose you have some particular test in mind when you write this, so you should file a werk-in-progress PR for that (with a snapshot dependency on this) and link them in GitHub so that reviewers can see the use case.

andresrc:
The original use case were test cases for [JENKINS-34681](https://issues.jenkins-ci.org/browse/JENKINS-34681) when using context params, but it was move to `SystemProperties`. I'll address the feedback and add a test case there.

batmat:
Given this is stalled since soon 2 years, I'm going to close this.

@andresrc obviously please feel totally free to reopen/comment in case you'd like to get back to this and push it to completion.

Thanks!

PR #20, opened 2016-05-31
Better version of JenkinsRule.getLog

jglick:
Strip console annotations from output. Helpful in https://github.com/jenkinsci/workflow-cps-plugin/pull/14 for example.

@reviewbybees

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -203,6 +203,7 @@ import org.junit.runners.model.Statement;
 import com.gargoylesoftware.htmlunit.html.DomNodeUtil;
 import com.gargoylesoftware.htmlunit.html.HtmlFormUtil;
 import hudson.maven.MavenRequest;
+import java.io.ByteArrayOutputStream;
 import java.net.HttpURLConnection;
 import java.nio.channels.ClosedByInterruptException;
 import org.jvnet.hudson.test.recipes.Recipe;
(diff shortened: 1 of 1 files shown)


jglick:
For example prints

    [Pipeline] node

rather than

    ha:AAAAaB+LCAAAAAAAAP9b85aBtbiIwTG/KF0vKzUvOzOvODlTryCnNB3I0ivPL8pOy8kv18vKT9JLzs8rzs9J1QuHCgaV5jlDhPzyS1IZIICRiYGhoohBKqM0pTg/D64Hh8ICAFt0h+h/AAAA[Pipeline] node

oleg-nenashev:
I'm aware it may cause regressions in plugin test suites, which check annotations using the original implementation. IMHO it's better to deprecate the current method and to create <code>getAnnotatedLog()</code> and <code>getPlainTextLog</code>

oleg-nenashev:
So 🐜  since it breaks the original behavior (undocumented one BTW).

amuniz:
:bee:

oleg-nenashev:
@jglick 
> looking for Base-64 serialized annotations! The only sane test would be to use other APIs which actually list the annotations in a consumable format.

They can still deserialize these strings.
I do not feel so strong regarding the replacement of the current method behavior since it has not been documented in any case. But a annotated log method would be useful.

jglick:
> They can still deserialize these strings.

In principle they _could_, but do you seriously expect that anyone ever _has_?

> a[n] annotated log method would be useful

For whom?

oleg-nenashev:
> For whom?

For an abstract test developer. At least I have 2 plugins where such method would be useful if I finally write tests there :(

jglick:
> I have 2 plugins where such method would be useful if I finally write tests there

Well perhaps you should write those tests first, figuring out as you go what a reasonable way to assert console annotators might be—and then file a PR to add some utilities to this repository when that is settled.

PR #22, opened 2016-06-07
[JENKINS-35445] Custom HtmlUnit Dependency for JTH

andresrc:
See [JENKINS-35445](https://issues.jenkins-ci.org/browse/JENKINS-35445)

Downstream of https://github.com/andresrc/jenkins-test-harness-htmlunit/pull/1

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,14 +60,14 @@ THE SOFTWARE.
   <repositories>
     <repository>
       <id>repo.jenkins-ci.org</id>
-      <url>http://repo.jenkins-ci.org/public/</url>
+      <url>https://repo.jenkins-ci.org/public/</url>
     </repository>
   </repositories>
 
   <pluginRepositories>
     <pluginRepository>
       <id>repo.jenkins-ci.org</id>
-      <url>http://repo.jenkins-ci.org/public/</url>
+      <url>https://repo.jenkins-ci.org/public/</url>
     </pluginRepository>
   </pluginRepositories>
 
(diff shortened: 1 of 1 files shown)


andresrc:
The upstream snapshot was not deployed, closing and reopening.

@reviewbybees

andresrc:
@jenkinsci/code-reviewers Please see [JENKINS-35445](https://issues.jenkins-ci.org/browse/JENKINS-35445) and https://github.com/andresrc/jenkins-test-harness-htmlunit/issues/1

olivergondza:
CCing @tfennelly who worked hard to get us back to upstream version. This seems to be a step back.

jglick:
> back to upstream version. This seems to be a step back.

Well, this is not a fork of the sources. It only shades the classes from an official release.

recena:
@olivergondza 

>  This seems to be a step back.

I don't think so. Now, at least, we are using an official version (ref. [JENKINS-31161](https://issues.jenkins-ci.org/browse/JENKINS-31161)).

👍 with this solution if we create a JIRA issue to solve the real problem.

batmat:
Weak :+1:, IIUC https://github.com/andresrc/jenkins-test-harness-htmlunit is gonna need to get forked under jenkinsci right?
Then, when forked, we need to be careful of what happens there too to not waste Tom's work in the future, as rightly fears @olivergondza IMO.

jglick:
:bee:

andresrc:
@olivergondza @batmat when we need to upgrade `HtmlUnit` we just need to repackage it in the upstream jar. I've left version numbers that make it easy to map to the upstream project.

No `HtmlUnit` classes are transformed, just some of its transitive dependencies, which are the ones causing problems. In a next step I'll try to go for the xml libraries, which are also causing some problems (see https://github.com/jenkinsci/jenkins/pull/2404).

I'm going to merge and cut a release to pick it in the parent pom.

PR #23, opened 2016-06-10
JenkinsRule.getLog should return an empty string if the log file has not yet even been written

jglick:
Fixes a regression from #20 observed to affect `ExecutorStepTest.queueTaskVisibility`, which calls `waitForMessage` immediately after `QueueTaskFuture.waitForStart` and thus has a race condition.

@reviewbybees


andresrc:
:bee:

rsandell:
:bee:

amuniz:
:bee:

oleg-nenashev:
🐝 and @reviewbybees done

PR #24, opened 2016-06-13
Fix wiki link

fungusakafungus:
Make it a protocol-less absolute link instead of a relative link, which didn't work.


andresrc:
:+1:

PR #25, opened 2016-07-01
Generalize assertBuildStatus and related methods

jglick:
- [X] allow `assertBuildStatus` to be called without forcing the test to call `QueueTaskFuture.get`
- [X] allow `buildAndAssertSuccess` to be called on non-`AbstractProject`s

@reviewbybees

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -202,9 +202,13 @@ import org.junit.runners.model.Statement;
 import com.gargoylesoftware.htmlunit.html.DomNodeUtil;
 import com.gargoylesoftware.htmlunit.html.HtmlFormUtil;
 import hudson.maven.MavenRequest;
+import hudson.model.Job;
+import hudson.model.queue.QueueTaskFuture;
 import java.io.ByteArrayOutputStream;
 import java.net.HttpURLConnection;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -1154,12 +1163,19 @@ public void assertGoodStatus(Page page) {
...
-        assertThat("build was actually scheduled", r, Matchers.notNullValue());
-        return assertBuildStatusSuccess(r.get());
+        return assertBuildStatus(Result.SUCCESS, r);
     }
 
-    public <J extends AbstractProject<J,R>,R extends AbstractBuild<J,R>> R buildAndAssertSuccess(J job) throws Exception {
-        return assertBuildStatusSuccess(job.scheduleBuild2(0));
+    public <J extends Job<J,R>,R extends Run<J,R>> R buildAndAssertSuccess(final J job) throws Exception {
+        assertThat(job, IsInstanceOf.instanceOf(ParameterizedJobMixIn.ParameterizedJob.class));
+        QueueTaskFuture f = new ParameterizedJobMixIn() {
```
jglick: Note that there is a new `static` convenience method in 1.621, but

* our baseline is 1.580.x
* I could not get it to compile anyway—generics hell

recena:
👍  and 🐝

PR #27, opened 2016-07-08
LoggerRule

jglick:
Should be self-explanatory.

@reviewbybees

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/LoggerRule.java
+++ b/src/main/java/org/jvnet/hudson/test/LoggerRule.java
@@ -43,22 +43,50 @@ import org.junit.rules.RuleChain;
 /**
  * A test rule which allows you to easily enable one or more loggers for the duration of a test.
  * Call {@link #record(Class, Level)} or another overload for the rule to take effect.
+ * <p>By default messages are merely printed to test output.
+ * If you also want to examine them, call TODO
  * <p>To capture messages during Jenkins startup,
  * you may compose this with a {@link JenkinsRule} using a {@link RuleChain};
  * or use as a {@link ClassRule}.
  */
 public class LoggerRule extends ExternalResource {
 
-    private final RingBufferLogHandler ringHandler = new RingBufferLogHandler();
     private final ConsoleHandler consoleHandler = new ConsoleHandler();
     private final Map<Logger,Level> loggers = new HashMap<Logger,Level>();
+    // initialized iff capture is called:
+    private RingBufferLogHandler ringHandler;
+    private List<String> messages;
 
     /**
      * Initializes the rule, by default not recording anything.
      */
     public LoggerRule() {
-        ringHandler.setLevel(Level.ALL);
         consoleHandler.setLevel(Level.ALL);
+    }
+
+    /**
+     * Initializes log record capture, in addition to merely printing it.
+     * This allows you to call {@link #getRecords} and/or {@link #getRecordsFormatted} later.
+     * @param maximum the maximum number of records to keep (any further will be discarded)
+     * @return this rule, for convenience
+     */
+    public LoggerRule capture(int maximum) {
+        messages = new ArrayList<String>();
+        ringHandler = new RingBufferLogHandler(maximum) {
+            final Formatter f = new SimpleFormatter(); // placeholder instance for what should have been a static method perhaps
+            @Override
+            public synchronized void publish(LogRecord record) {
+                super.publish(record);
+                String message = f.formatMessage(record);
+                Throwable x = record.getThrown();
+                messages.add(message == null && x != null ? x.toString() : message);
+            }
+        };
+        ringHandler.setLevel(Level.ALL);
+        for (Logger logger : loggers.keySet()) {
+            logger.addHandler(ringHandler);
+        }
+        return this;
     }
 
     /**
@@ -76,8 +104,10 @@ public class LoggerRule extends ExternalResource {
     public LoggerRule record(Logger logger, Level level) {
         loggers.put(logger, logger.getLevel());
         logger.setLevel(level);
-        logger.addHandler(ringHandler);
         logger.addHandler(consoleHandler);
+        if (ringHandler != null) {
+            logger.addHandler(ringHandler);
+        }
         return this;
     }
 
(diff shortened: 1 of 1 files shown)


recena:
Amazing! 🐝

andresrc:
:bee:

src/main/java/org/jvnet/hudson/test/LoggerRule.java
```diff
@@ -0,0 +1,129 @@
...
+     * Obtains all log records collected so far during this test case.
+     */
+    public List<LogRecord> getRecords() {
+        return ringHandler.getView();
+    }
+
+    /**
+     * Like {@link #getRecords} but applies {@link SimpleFormatter#format(LogRecord)}.
+     */
+    public List<String> getRecordsFormatted() {
```
jtnord: Note: the formatted may return different results at different times for the same record as this is using async formatting vs sync formatting that Jenkins normally uses. 
Probably worth calling this out.

src/main/java/org/jvnet/hudson/test/LoggerRule.java
```diff
@@ -0,0 +1,129 @@
...
+        Formatter f = new SimpleFormatter();
+        List<String> messages = new ArrayList<String>();
+        for (LogRecord lr : getRecords()) {
+            messages.add(f.format(lr).trim());
```
jtnord: why trim.. that whitespace could be important.  (it shouldn;t be as you should probably put records in quotes and the like, but seems a bit weired)

jglick: For the one demonstrated use case (`StartupTest`), trailing newlines are unwanted. You can always call `getRecords()` if you want to examine precise details. This was more a convenience method to be able to, say, assert that the list of records was empty, or included some substring, etc.

src/main/java/org/jvnet/hudson/test/LoggerRule.java
jglick: Sorry, no idea what you are talking about. How can formatting be “synchronous” or “asynchronous”?

And what formatting in Jenkins? `Functions.printLogRecordHtml`?

jtnord: so with the normal handlers in java.util.logging the message is computed on the call stack in the logger.log(...) method.  Thus if you pass in non idempotent Objects for the formatter (`Log.log(Level.INFO, "{0}, {1}" new Object[mutable1, mutable2]`) then the message is created at a time that represents the state of those mutable as seen in the caller - if you are logging mutable state that is being changed by multiple threads then you are SOOL.

Now - some logging handlers/frameworks allow for asynchronous logging - such that the creation of the message is decoupled from the caller of the log method.  such that the state of the mutable objects will be different at the time the message is created.

e.g.  with a synchronous logging handler...

```java
class Foo {
  public String s;
  public String toString() { return s }
}
// some class
Foo f = new Foo();
f.s = "jesse"
Logger.log(Level.INFO, "Hello from {0}", f);
f.s = "jglick";
```

so that will with synchronous logging (all handlers etc) print  "Hello from jesse".  
However when you then test the method to see if "Hello from jesse" was printed - you will return false (as its now formatted asynchrously from the call to log).

jtnord:
:bee: for the code, think the javadoc could be clarified a little for users, but not a blocker.

jglick:
> the javadoc could be clarified a little

Any particular places?

src/main/java/org/jvnet/hudson/test/LoggerRule.java
jglick: OK, I see what you mean. Might matter for some tests.

FTR Jenkins itself uses asynchronous formatting, in `LogManager`. Synchronous formatting would only happen for the console handler or web container handler outside of Jenkins’ control.

jglick: Suppose I could kill three birds with one stone by making the buffer handler be disabled by default, but could be turned on with a call initializing it with the `ringSize`, and then having the messages be formatted during capture.

jtnord:
>> the javadoc could be clarified a little
> Any particular places?

The `getRecord*()`  around when the formatting happens (although if you are changing that this becomes moot) , and that not "all records collected" is not "all, but ones that have been collected and not expunged".  (commented inline)

src/main/java/org/jvnet/hudson/test/LoggerRule.java
```diff
@@ -43,25 +43,53 @@
 /**
  * A test rule which allows you to easily enable one or more loggers for the duration of a test.
  * Call {@link #record(Class, Level)} or another overload for the rule to take effect.
+ * <p>By default messages are merely printed to test output.
+ * If you also want to examine them, call TODO
```
jtnord: TODO?  (looks like you have implemented this!)
call `#capture(int)`and subsequently `#getRecords` to retreive the last n records?

src/main/java/org/jvnet/hudson/test/LoggerRule.java
```diff
@@ -0,0 +1,129 @@
...
+    
+    /**
+     * Same as {@link #record(String, Level) but calls {@link Class#getName()} for you first.
+     */
+    public LoggerRule record(Class<?> clazz, Level level) {
+        return record(clazz.getName(), level);
+    }
+
+    /**
+     * Obtains all log records collected so far during this test case.
```
jtnord: fixed by subsequent code change so no longer relevant

src/main/java/org/jvnet/hudson/test/LoggerRule.java
jglick: D’oh, will fix in a sec.

src/main/java/org/jvnet/hudson/test/LoggerRule.java
```diff
@@ -0,0 +1,163 @@
...
+    public LoggerRule capture(int maximum) {
+        messages = new ArrayList<String>();
+        ringHandler = new RingBufferLogHandler(maximum) {
+            final Formatter f = new SimpleFormatter(); // placeholder instance for what should have been a static method perhaps
+            @Override
+            public synchronized void publish(LogRecord record) {
+                super.publish(record);
+                String message = f.formatMessage(record);
+                Throwable x = record.getThrown();
+                messages.add(message == null && x != null ? x.toString() : message);
```
jtnord: so what puts logs with a message but no exception into `messages`?

I take it the unbounded nature is what you are expecting here.

jglick: > what puts logs with a message but no exception into `messages`?

Not following the question.

You either have records with a message:

```java
LOGGER.log(Level.WARNING, "Broken thingy: {0}", thingy);
```

```java
} catch (Exception x) {
    LOGGER.log(Level.WARNING, "Failed to do stuff", x);
}
```

or without:

```java

} catch (Exception x) {
    LOGGER.log(Level.WARNING, null, x);
}
```

Both are handled here.

jtnord: yeah - it's Friday.  Sorry about that.

jtnord: > I take it the unbounded nature is what you are expecting here.

Did I miss something that was limiting the number of messages kept in the `ArrayList` rather than the ring buffer?

jglick: No. Do we really care?

jtnord: no - but the doc says a maximum number of records (which whilst correct as messages != records) I was just wondering if this was your intention or an oversight.

jtnord:
:bee:

src/main/java/org/jvnet/hudson/test/LoggerRule.java
jglick: An oversight, though I see little advantage in truncating messages. Really it might be better to not truncate records, either; would just need a slightly more customized `LogHandler`.

PR #28, opened 2016-07-19
Add a WithOS annotation.

abayer:
This is basically a copy of WithOS from the ATH (along with the
RuleAnnotation class it needs) - I've hit cases where I'd like unit
tests to be able to run on some OSes but be skipped on others without
the tests actually failing. The immediate example of this is trying to
build plugins that have unit tests using Docker Pipeline while on OS
X, because ow, lots of problems (like
https://issues.jenkins-ci.org/browse/JENKINS-36802). Fun!

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/junit/RuleAnnotation.java
+++ b/src/main/java/org/jvnet/hudson/test/junit/RuleAnnotation.java
@@ -1,0 +1,67 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.jvnet.hudson.test.junit;
+
+import org.junit.rules.TestRule;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Meta-annotation for annotations that introduces a {@link TestRule} for test.
+ *
+ * <p>
+ * This allows annotations on test class/method to add additional setup/shutdown behaviours.
+ *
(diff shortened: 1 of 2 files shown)


src/main/java/org/jvnet/hudson/test/junit/RuleAnnotation.java
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * <p>
+ * This allows annotations on test class/method to add additional setup/shutdown behaviours.
+ *
+ * @author Kohsuke Kawaguchi
+ */
+@Retention(RUNTIME)
+@Target(ANNOTATION_TYPE)
+@Documented
+public @interface RuleAnnotation {
```
jglick: I do not see anything actually looking for this. Am I missing something?

jglick:
I tend to find it simpler and clearer to explicitly throw `AssumptionViolatedException` from the test. That way it is obvious how to tweak the condition, to look for a particular binary for example.

src/main/java/org/jvnet/hudson/test/junit/WithOS.java
```diff
@@ -0,0 +1,107 @@
...
+    public class RuleImpl implements TestRule {
+        @Override
+        public Statement apply(final Statement base, final Description d) {
+            return new Statement() {
+
+                @Override
+                public void evaluate() throws Throwable {
+                    Class<?> testSuite = d.getTestClass();
+                    check(d.getAnnotation(WithOS.class), testSuite);
+                    check(testSuite.getAnnotation(WithOS.class), testSuite);
```
stephenc: Why is this necessary, the annotation should be present in the `Description` otherwise there is a bug in JUnit IIUC

stephenc:
If we drop the TestSuite hack I think it is OK as that way the annotation is in the expected places

src/main/java/org/jvnet/hudson/test/junit/RuleAnnotation.java
abayer: `@RuleAnnotation` is on `WithOS`.

src/main/java/org/jvnet/hudson/test/junit/WithOS.java
abayer: Honestly, that's copy-pasted from the ATH, so it may not be needed at all. I'll experiment with it.

abayer:
@jglick - so the reason I want this rather than throwing an Exception is that I don't want the tests to fail when their assumptions aren't satisfied - whatever gets me that result (skipped rather than error/failed) is fine by me.

stephenc:
@abayer `assumeThat(os.isWindows()` would get you that...

but a more interesting PoV for me on this annotation is that we could index it to find the list of tests that are OS specific and then mark those as needing to have at least one run on a matrix of OSs...

Now that is not something we can do right now, but having the annotation in place to let us start capturing that information is a big +1 in my view

src/main/java/org/jvnet/hudson/test/junit/RuleAnnotation.java
jglick: I see where the annotation is _applied_ of course. But where is it _consulted_?

jglick: [reference](https://github.com/jenkinsci/acceptance-test-harness/blob/c35e9b940864f2ce76debc45c9712e36c1bbb644/src/main/java/org/jenkinsci/test/acceptance/junit/JenkinsAcceptanceTestRule.java#L99)

src/main/java/org/jvnet/hudson/test/junit/RuleAnnotation.java
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * The rule class that defines the setup/shutdown behaviour.
+     *
+     * The instance is obtained through Guice.
```
jglick: Huh??

src/main/java/org/jvnet/hudson/test/junit/WithOS.java
jglick: This is to support annotations on both the test case (method) or suite (class).

jglick:
> I don't want the tests to fail when their assumptions aren't satisfied - whatever gets me that result (skipped rather than error/failed) is fine by me

That is what throwing `AssumptionViolatedException` from a test accomplishes. Now if @stephenc wants to construct a system like he proposes, great, but for now that remains hypothetical.

At any rate, :bug: since as written it does not do anything.

src/main/java/org/jvnet/hudson/test/junit/RuleAnnotation.java
stephenc: Oh dear me more Guice "magic"... I say rip that out, the original author uses it like goto

PR #30, opened 2016-08-04
Relaxing the rule that an unpacked @LocalData must include root config.xml

jglick:
If you have no need to override Jenkins global configuration in your test data, it is annoying to have to create an empty `theTest/config.xml`

```xml
<hudson/>
```

With this patch, there is no longer any need to do so. Should make it more convenient to have unpacked test data, which is easier to review and version (as in https://github.com/jenkinsci/workflow-job-plugin/pull/14).

@reviewbybees


abayer:
:bee:

PR #31, opened 2016-08-11
use a subdirectory of the temp directory by default.

jtnord:
1) we are Jenkins :-)
2) put all the files in a directory inside temp rather than directly in
temp.

The problem with using the temporary directory directly is that you can not add
exclusions with wildcards in many anti virus products.  By using a
subdirectory you can add an exclusion to your PATH_TO_TEMP\jenkinsTests.tmp
which will then allow you to exclude jenkinsTests.tmp but still allowing
scanning of your temporary directory (which is desired).

The downside of this is that you will always have a jenkinsTests.tmp directory inside your temp directory - but well the cleanup does not work correctly anyway - so this is nothing new (on windows) - and rather than hundreds of directories you have one directory - which is easier to clean...

@reviewbybees

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
+++ b/src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
@@ -50,7 +50,8 @@ public class TemporaryDirectoryAllocator {
     }
 
     public TemporaryDirectoryAllocator() {
-        this(new File(System.getProperty("java.io.tmpdir")));
+        this.base = new File(System.getProperty("java.io.tmpdir"), "jenkinsTests.tmp");
+        base.mkdirs();
     }
 
     /**
@@ -61,7 +62,7 @@ public class TemporaryDirectoryAllocator {
      */
     public synchronized File allocate() throws IOException {
         try {
-            File f = File.createTempFile("hudson", "test", base);
+            File f = File.createTempFile("jenkins", "test", base);
             f.delete();
             f.mkdirs();
             tmpDirectories.add(f);


jglick:
This allocator is historical, but I could not figure out how to deprecate it. What about tests using the recommended `@TemporaryFolder` rule?

jtnord:
@jglick All tests use this - it is used by the @JenkinsRule iteslf to unpack the exploded jars.
[reference](https://github.com/jenkinsci/jenkins-test-harness/blob/84002fb0017cff338bc1735fb20f15ad93edc2bf/src/main/java/org/jvnet/hudson/test/HudsonHomeLoader.java#L54)

Honestly - I don't care as much about {{TemproaryFolder}} as
# plugins and other compressed archives generally do not get put in them that causes hammering of the AV.
# Can't controll that code as it is from third party
# changing {[java.io.tmpdir}} to achieve this would be bad - and hard to control as to when it gets changed - which could be timing dependant and after some resources have already been created pre change and some post change.  (changing surefire config would be an issue as you would have to have somethgin tat would ensure it is created otherwise bad things will happen)

jtnord:
FWIW

```
C:\workarea\source\github\cloudbees\operations-center-context>c:\cygwin\bin\find %TEMP% -maxdepth 1 -name hudson* | wc -l
632
```

src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
```diff
@@ -50,7 +50,8 @@ public TemporaryDirectoryAllocator(File base) {
     }
 
     public TemporaryDirectoryAllocator() {
-        this(new File(System.getProperty("java.io.tmpdir")));
+        this.base = new File(System.getProperty("java.io.tmpdir"), "jenkinsTests.tmp");
```
jglick: The `.tmp` suffix here is not really necessary.

jglick:
:bee:, seems useful enough.

PR #33, opened 2016-09-17
Make APIs public for benchmarking use too

svanoort:
Includes NoListenerConfiguration, probably WarExploder, etc


jglick:
Is there some intended use case for this filed as a downstream PR?

BTW please file PRs from forks rather than origin branches.

svanoort:
Not needed anymore -- supplanted by later changes post-hackfest.

PR #35, opened 2016-09-17
[JENKINS-38304] Support TestExtension for multiple test cases

ikedam:
[JENKINS-38304](https://issues.jenkins-ci.org/browse/JENKINS-38304)

This allows make classes annotated with `TestExtension` activate in multiple test cases:
```
public class Test {
    @Rule public JenkinsRule j = new JenkinsRule();

    @TestExtension({"testForExtension1", "testForExtension2"})
    public static class ExtensionForTest extends SomeExtensionPoint {
        ...
    }

    @Test
    public void testForExtension1() throws Extension {
        // ExtensionForTest is available
    }

    @Test
    public void testForExtension2() throws Extension {
        // ExtensionForTest is available
    }

    @Test
    public void testForAnotherPoint() throws Extension {
        // ExtensionForTest is not available
    }
```

### DIFF
--- a/src/test/java/org/jvnet/hudson/test/TestExtensionTest.java
+++ b/src/test/java/org/jvnet/hudson/test/TestExtensionTest.java
@@ -1,0 +1,105 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 IKEDA Yasuyuki
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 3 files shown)


ikedam:
You need run `mvn clean` to recreate the index file created by `sezpoz`.

src/test/java/org/jvnet/hudson/test/TestExtensionTest.java
```diff
@@ -0,0 +1,105 @@
...
+    public static class MultipleTests extends ItemListener {
+    }
+
+    private List<Class<? extends ItemListener>> getExtensionClasses() {
+        return Lists.transform(
+            j.jenkins.getExtensionList(ItemListener.class),
+            new Function<ItemListener, Class<? extends ItemListener>>() {
+                @Override
+                public Class<? extends ItemListener> apply(ItemListener arg0) {
+                    return arg0.getClass();
```
jglick: More comfortable to

```java
return arg0.getClass().getSimpleName();
```

so that later you can just write

```java
Matchers.hasItems("AllTests", "SingleTests", "MultipleTests")
```

jglick:
Beware that as a result of this change, if you have a plugin A which uses `jenkins-test-harness` 2.15- and creates a test JAR, and has some `@TestExtension`s specifying a test name, and then a plugin B which depends on A’s test JAR but uses `jenkins-test-harness` 2.16+, you will see an (apparently) harmless, but alarming, error in all of B’s tests:

```
… hudson.ExtensionFinder$GuiceFinder$SezpozModule configure
WARNING: Failed to load a.SomeTest$SomeTestExtension
java.lang.annotation.AnnotationTypeMismatchException: Incorrectly typed data found for annotation element public abstract java.lang.String[] org.jvnet.hudson.test.TestExtension.value() (Found data of type class java.lang.String[someTestName])
	at sun.reflect.annotation.AnnotationTypeMismatchExceptionProxy.generateException(AnnotationTypeMismatchExceptionProxy.java:57)
	at sun.reflect.annotation.AnnotationInvocationHandler.invoke(AnnotationInvocationHandler.java:84)
	at com.sun.proxy.$Proxy34.value(Unknown Source)
	at org.jvnet.hudson.test.TestExtensionLoader.isActive(TestExtensionLoader.java:67)
	at …
```

jglick:
In retrospect this PR was a really bad idea; should rather have introduced a new member (or introduced a `@TestExtensions` aggregator annotation, to ultimately be `@Repeatable`), not changed the type of `value`. Unfortunately it is too late to revert it now: plenty of test code has been compiled against the new version and there would be further errors going backwards. So the best we can do is try to push plugins to update their parent POM and thus JTH dependency, and do the same for their test dependencies.

PR #36, opened 2016-09-17
[JENKINS-38306] Add a parameter to LocalData to use instead of method name

ikedam:
[JENKINS-38306](https://issues.jenkins-ci.org/browse/JENKINS-38306)

It's often the case that I want to use a same local data in multiple test cases,
but not in all test cases in the test.
(e.g. some tests want to use other local data)

This change allows to specify the name of the directory to use
instead of method name
```
public class Test {
    @Rule public JenkinsRule j = new JenkinsRule();

    @Test
    @LocalData("config1")
    public void config1_test1() throws Exception {...}

    @Test
    @LocalData("config1")
    public void config1_test2() throws Exception {...}

    @Test
    @LocalData("config2")
    public void config2_test1() throws Exception {...}

    @Test
    @LocalData("config2")
    public void config2_test2() throws Exception {...}
}
```

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/HudsonHomeLoader.java
+++ b/src/main/java/org/jvnet/hudson/test/HudsonHomeLoader.java
@@ -25,12 +25,15 @@ package org.jvnet.hudson.test;
 
 import hudson.FilePath;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang.StringUtils;
 
 import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.Method;
 import java.net.MalformedURLException;
 import java.net.URL;
+
+import javax.annotation.CheckForNull;
 
 /**
  * Controls how a {@link HudsonTestCase} initializes <tt>JENKINS_HOME</tt>.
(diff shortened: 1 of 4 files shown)


src/main/java/org/jvnet/hudson/test/HudsonHomeLoader.java
```diff
@@ -121,11 +126,31 @@ public File allocate() throws Exception {
             return new CopyExisting(home).allocate();
         }
 
+        public static boolean isJavaIdentifier(@CheckForNull String name) {
```
jglick: Can be replaced with `SourceVersion.latestSupported().isIdentifier(name)`.

src/main/java/org/jvnet/hudson/test/HudsonHomeLoader.java
```diff
@@ -121,11 +126,31 @@ public File allocate() throws Exception {
...
+            }
+            return true;
+        }
+
         private URL findDataResource() {
             // first, check method specific resource
             Class<?> clazz = testMethod.getDeclaringClass();
+            String methodName = testMethod.getName();
+            
+            if (isJavaIdentifier(alterName)) {
```
jglick: Why do we care whether it is a Java identifier or not? Should not matter for purposes of resource loading. Anyway, if it is unusable for some reason, we should fail with an error.

ikedam:
The Java identifier checks ensure that `alterName` doesn't contain special paths such as "..". I worry that those special paths might result unexpected behaviors.

PR #37, opened 2016-09-30
Make fewer assumptions about log handling

jglick:
Needed for https://github.com/jenkinsci/workflow-job-plugin/pull/27, but generally should be an improvement.

@reviewbybees

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/BuildWatcher.java
+++ b/src/main/java/org/jvnet/hudson/test/BuildWatcher.java
@@ -25,7 +25,6 @@
 package org.jvnet.hudson.test;
 
 import hudson.Extension;
-import hudson.console.AnnotatedLargeText;
 import hudson.console.LineTransformationOutputStream;
 import hudson.model.Run;
 import hudson.model.TaskListener;
@@ -90,7 +89,7 @@ public final class BuildWatcher extends ExternalResource {
                 return;
             }
             RunningBuild build = new RunningBuild(r);
-            RunningBuild orig = builds.put(r.getLogFile(), build);
+            RunningBuild orig = builds.put(r.getRootDir(), build);
             if (orig != null) {
                 System.err.println(r + " was started twice?!");
             }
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/BuildWatcher.java
```diff
@@ -90,7 +89,7 @@
                 return;
             }
             RunningBuild build = new RunningBuild(r);
-            RunningBuild orig = builds.put(r.getLogFile(), build);
+            RunningBuild orig = builds.put(r.getRootDir(), build);
```
jglick: We just needed a unique identifier of a build × `$JENKINS_HOME`; were not actually using the log file per se.

src/main/java/org/jvnet/hudson/test/BuildWatcher.java
```diff
@@ -112,18 +111,18 @@
...
         RunningBuild(Run<?,?> r) {
-            log = r.getLogText();
+            this.r = r;
             sink = new LogLinePrefixOutputFilter(System.err, "[" + r + "] ");
         }
 
         synchronized void copy() {
             try {
-                pos = log.writeLogTo(pos, sink);
+                pos = r.getLogText().writeLogTo(pos, sink);
```
jglick: See [this precedent](https://github.com/jenkinsci/jenkins/blob/3f5fe7f6b9b8dab37e82b360574102c6388729e8/core/src/main/java/hudson/cli/ConsoleCommand.java#L75-L79). `getLogText` in general needs to be called repeatedly as more log output arrives.

src/main/java/org/jvnet/hudson/test/BuildWatcher.java
oleg-nenashev: Won't help with ExternalTaskLogging story

oleg-nenashev:
🐝

src/main/java/org/jvnet/hudson/test/BuildWatcher.java
jglick: If you are referring to logging from non-`Run`s, that is true, but anyway `BuildWatcher` does not support them currently either.

PR #38, opened 2016-10-06
Avoid calling Run.getLogFile merely to check if the file exists

jglick:
Suppresses a warning in https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/21. Follows up #37.

@reviewbybees esp. @oleg-nenashev


abayer:
:bee:

oleg-nenashev:
belated 🐝

PR #39, opened 2016-10-19
Add way to record and mock http apis

kwhetstone:
Added support for recording & mocking http responses with Wiremock.  This particular rule allows a test to serve up prerendered files which correspond to particular api responses.  If a response is not saved (trying to access an unmocked url), the test will fail.  You can enable a "recording" mode in which you can run your test through on a live server and save the responses for later offline testing.  The recordings are saved to `/src/test/resources/` of whatever project uses this rule.

While the particular constructors I've included are the ones that have been the most useful to me, they too can be expanded.  Other options for the server can be found on [wiremock.org](http://wiremock.org)

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -173,6 +173,11 @@ THE SOFTWARE.
       <artifactId>wiremock</artifactId>
       <version>2.1.12</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>19.0</version>
+    </dependency>
   </dependencies>
 
   <build>
--- a/pom.xml
+++ b/pom.xml
@@ -48,6 +48,7 @@ THE SOFTWARE.
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <jetty.version>9.2.15.v20160210</jetty.version>
+    <recordMocks>false</recordMocks>
   </properties>
 
   <licenses>
@@ -167,6 +168,17 @@ THE SOFTWARE.
       <version>1.0.2</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.github.tomakehurst</groupId>
+      <artifactId>wiremock</artifactId>
+      <version>2.1.12</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>19.0</version>
+    </dependency>
   </dependencies>
 
   <build>
@@ -189,6 +201,7 @@ THE SOFTWARE.
               <hudson.maven.debug>${mavenDebug}</hudson.maven.debug>
               <buildDirectory>${project.build.directory}</buildDirectory>
               <ignore.random.failures>${ignore.random.failures}</ignore.random.failures>
+              <toRecord>${recordMocks}</toRecord>
           </systemPropertyVariables>
         </configuration>
       </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -172,12 +172,6 @@ THE SOFTWARE.
       <groupId>com.github.tomakehurst</groupId>
       <artifactId>wiremock</artifactId>
       <version>2.1.12</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-      <version>19.0</version>
     </dependency>
   </dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,7 @@ THE SOFTWARE.
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <jetty.version>9.2.15.v20160210</jetty.version>
-    <recordMocks>false</recordMocks>
+    <wiremock.record/>
   </properties>
 
   <licenses>
@@ -195,7 +195,7 @@ THE SOFTWARE.
               <hudson.maven.debug>${mavenDebug}</hudson.maven.debug>
               <buildDirectory>${project.build.directory}</buildDirectory>
               <ignore.random.failures>${ignore.random.failures}</ignore.random.failures>
-              <toRecord>${recordMocks}</toRecord>
+              <wiremock.record>${wiremock.record}</wiremock.record>
           </systemPropertyVariables>
         </configuration>
       </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -170,13 +170,8 @@ THE SOFTWARE.
     </dependency>
     <dependency>
       <groupId>com.github.tomakehurst</groupId>
-      <artifactId>wiremock</artifactId>
-      <version>2.1.12</version>
-    </dependency>
-    <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-      <version>19.0</version>
+      <artifactId>wiremock-standalone</artifactId>
+      <version>2.4.1</version>
     </dependency>
   </dependencies>
 
--- a/src/main/java/org/jvnet/hudson/WireMockRuleFactory.java
+++ b/src/main/java/org/jvnet/hudson/WireMockRuleFactory.java
@@ -1,0 +1,60 @@
+package org.jvnet.hudson.test;
+
+import com.github.tomakehurst.wiremock.common.SingleRootFileSource;
+import com.github.tomakehurst.wiremock.core.Options;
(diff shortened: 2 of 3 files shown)


src/main/java/org/jvnet/hudson/WireMockRuleFactory.java
```diff
@@ -0,0 +1,60 @@
...
+ * }
+ * </pre>
+ *
+ * Point your system to "http://localhost:8089/" to see mocked responses.
+ *
+ * To record mocks: {@code mvn clean test -DrecordMocks=true}  Clearing 
+ * previously recorded mocks before generating new ones is recommended. 
+ */
+public class WireMockRuleFactory {
+    private String prop = System.getProperty("toRecord");
```
stephenc: given that this property is completely under your own control, I would have this at least include wiremock, e.g. `wiremock.record` and I would use the same property name both for the system property as for the pom property

src/main/java/org/jvnet/hudson/WireMockRuleFactory.java
```diff
@@ -0,0 +1,60 @@
...
+            return new WireMockRule(options);
+        }
+    }
+
+
+    private class WireMockRecorderRule extends WireMockRule {
+        //needed for WireMockRule file location
+        private String mappingLocation = "src/test/resources";
+
+        public WireMockRecorderRule(Options options, String url){
```
stephenc: it would seem that the logical way to do this is have the `wiremock.record` property value be the URL to proxy... if the property is blank then no recording... if the property is non-blank, set up the recorder for the specified url. That makes running easier...

kwhetstone: That's a good point.  This would also allow you to potentially record a couple different servers over several different iterations of your tests.

kwhetstone:
Thanks for the review @stephenc  I've addressed your comments in a followup PR.  They do make it easier to use this feature.

src/main/java/org/jvnet/hudson/test/WireMockRuleFactory.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * Point your system to "http://localhost:8089/" to see mocked responses.
+ *
+ * To record mocks: {@code mvn test -wiremock.record="http://api_url"}  Clearing 
+ * previously recorded mocks before generating new ones is recommended. 
+ */
+public class WireMockRuleFactory {
+    private String urlToMock = System.getProperty("wiremock.record");
+
+    public WireMockRule getRule(int port){
```
recena: `public WireMockRule getRule(int port) {`

At least, we should use the same rules in the file.

src/main/java/org/jvnet/hudson/test/WireMockRuleFactory.java
```diff
@@ -0,0 +1,52 @@
...
+            return new WireMockRule(options);
+        }
+    }
+
+
+    private class WireMockRecorderRule extends WireMockRule {
+        //needed for WireMockRule file location
+        private String mappingLocation = "src/test/resources";
+
+        public WireMockRecorderRule(Options options, String url){
```
recena: Ditto

kwhetstone:
@reviewbybees done.  If no one objects, this can be merged.

src/main/java/org/jvnet/hudson/test/WireMockRuleFactory.java
```diff
@@ -0,0 +1,52 @@
...
+ * {@code
+ * WireMockRuleFactory wmrf = new WireMockRuleFactory();
+ * @Rule
+ * WireMockRule = wmrf.getRule(8089); //port is configurable  
+ * }
+ * </pre>
+ *
+ * Point your system to "http://localhost:8089/" to see mocked responses.
+ *
+ * To record mocks: {@code mvn test -wiremock.record="http://api_url"}  Clearing
```
jglick: Did you forget the `D` in `-D`?

jglick:
I presume that this was designed to simplify https://github.com/jenkinsci/github-branch-source-plugin/pull/79, right? If so, can we see that PR updated to use a `SNAPSHOT` version of this PR? (`mvn deploy` for the benefit of CI)

src/main/java/org/jvnet/hudson/test/WireMockRuleFactory.java
recena: @kwhetstone space between `(int port)` and `{` in order to apply the same rules in the file

kwhetstone:
@jglick yes, it is.

pom.xml
```diff
@@ -173,6 +173,11 @@ THE SOFTWARE.
       <artifactId>wiremock</artifactId>
       <version>2.1.12</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>19.0</version>
```
jglick: You do not want to do this. It will clash with the Guava build in Jenkins core.

kwhetstone: I need minimally guava-18.0 and the version we're pulling in through Jenkins core is guava-11.0.1.  Not 100% sure how to work around this at the moment; I might have to rethink this.

stephenc:
:bug: with the guava version conflict

stephenc:
```
<dependency>
    <groupId>com.github.tomakehurst</groupId>
    <artifactId>wiremock-standalone</artifactId>
    <version>2.4.1</version>
</dependency>
```

looks to have all dependencies shaded

src/main/java/org/jvnet/hudson/test/WireMockRuleFactory.java
```diff
@@ -0,0 +1,52 @@
...
+
+
+    private class WireMockRecorderRule extends WireMockRule {
+        //needed for WireMockRule file location
+        private String mappingLocation = "src/test/resources";
+
+        public WireMockRecorderRule(Options options, String url) {
+            super(options);
+            this.stubFor(get(urlMatching(".*")).atPriority(10).willReturn(aResponse().proxiedFrom(url)));
+            this.enableRecordMappings(new SingleRootFileSource(mappingLocation + "/mappings"), new SingleRootFileSource(mappingLocation + "/__files"));
```
stephenc: This likely will fail for multi-module projects where `new File(".")` is not the same directory as `pom.xml`

kwhetstone: Sure, that could be added as a new option.

pom.xml
kwhetstone: Removed; using wiremock-standalone which will include all the pieces without conflicts.

kwhetstone:
I've updated the list to cover choosing a custom folder location.  The mocks can be recorded from the command line as before.  Since recording and mocking will read out of the same folder, no extra configuration is needed there.

src/main/java/org/jvnet/hudson/test/WireMockRuleFactory.java
```diff
@@ -0,0 +1,82 @@
...
+            finshWireMockRuleSetup(options, url, this.mappingLocation);
+        }
+
+        public WireMockRecorderRule(WireMockConfiguration options, String url, String mappingLocation) {
+            super(options);
+            finshWireMockRuleSetup(options, url, mappingLocation);
+        }
+
+        private void finshWireMockRuleSetup(WireMockConfiguration options, String url, String mappingLocation) {
+            this.stubFor(get(urlMatching(".*")).atPriority(10).willReturn(aResponse().proxiedFrom(url)));
```
stephenc: Spelling of "Finish"

kwhetstone: Fixed!

kwhetstone:
I've made the requested changes.  Thanks Jesse and Stephen for the feedback.

src/main/java/org/jvnet/hudson/test/WireMockRuleFactory.java
```diff
@@ -0,0 +1,82 @@
...
+    public WireMockRule getRule(String mapLoc) {
+        return getRule(mapLoc, WireMockConfiguration.wireMockConfig());
+    }
+
+    public WireMockRule getRule(String mapLoc, WireMockConfiguration options) {
+        if(urlToMock != null && !urlToMock.isEmpty()) {
+            return new WireMockRecorderRule(options, urlToMock, mapLoc);
+        } else {
+            return new WireMockRule(options
+               .usingFilesUnderClasspath(mapLoc));
```
stephenc: This will not work the way you think if mapLoc is `src/test/resources/mockLoc` as you provide in the example javadoc

jglick:
IIUC with https://github.com/jenkinsci/github-api-plugin/pull/13 the intent is to move away from WireMock.

PR #40, opened 2016-10-20
Better report `java.lang.IllegalArgumentException: URI is not hierarchical`

jglick:
@abayer encountered this in IntelliJ. I have in the past in NetBeans, though I am not sure how to reproduce it offhand. IIRC it happens when you have a snapshot test dependency which you have not properly built (`mvn -DskipTests install`)—though I cannot reproduce that at the moment.

@reviewbybees


abayer:
In my (annoying!) case, it seems that Intellij is insisting on putting the `other-module/target/classes` directory on the classpath for the test execution, but not `other-module/target/other-module.jar`...

abayer:
:bee:

PR #41, opened 2016-11-03
Optional mode to track down soft/weak references that delay collection in assertGC

jglick:
Found useful in https://github.com/jenkinsci/workflow-cps-plugin/pull/83.

@reviewbybees


abayer:
:bee:

PR #44, opened 2016-12-09
[FIXED JENKINS-40351] provide missing config page

mc1arke:
Copies `configuration.jelly` and associated internationalization properties files from `ComputerConnectorTester` to `JenkinsComputerConnectorTester` so that `JenkinsRule`'s `doConfigRoundTrip(ComputerConnector)` can execute in the same way as `HudsonTestCase` currently does, rather than erroring when attempting to load the `.../configure` URL.

### DIFF
--- a/src/main/resources/org/jvnet/hudson/test/JenkinsComputerConnectorTester/configure_es.properties
+++ b/src/main/resources/org/jvnet/hudson/test/JenkinsComputerConnectorTester/configure_es.properties
@@ -1,0 +1,23 @@
+# The MIT License
+#
+# Copyright (c) 2004-, Kohsuke Kawaguchi, Sun Microsystems, Inc., and a number of other of contributers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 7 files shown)


src/main/resources/org/jvnet/hudson/test/JenkinsComputerConnectorTester/configure_es.properties
```diff
@@ -0,0 +1,23 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+Save=Guardar
```
jglick: I think we can delete all the translations. Irrelevant for tests.

PR #46, opened 2017-01-29
Make classes public.

KostyaSha:
Last 2 non-public classes. Blocks ability override `org.jvnet.hudson.test.JenkinsRule#createWebServer`

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jenkins-test-harness/46)

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/NoListenerConfiguration.java
+++ b/src/main/java/org/jvnet/hudson/test/NoListenerConfiguration.java
@@ -38,7 +38,7 @@ import javax.servlet.ServletContextListener;
  *
  * @author Kohsuke Kawaguchi
  */
-final class NoListenerConfiguration extends AbstractLifeCycle {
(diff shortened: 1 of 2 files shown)


KostyaSha:
@jglick @oleg-nenashev ?

jglick:
I do not think you should be overriding this method without calling `super`. This is poking too much into the internals. And there is no downstream PR showing the usage.

KostyaSha:
> I do not think you should be overriding this method without calling super. This is poking too much into the internals. And there is no downstream PR showing the usage.

PR of what? That's the only way how i can unhardcode "localhost" https://github.com/KostyaSha/yet-another-docker-plugin/pull/126/files#diff-607ffdaaf8eed43fca17dbb21eecc980R42

KostyaSha:
@jglick ??

oleg-nenashev:
@jglick ping. Otherwise will merge on Tuesday since there is no explicit -1

PR #47, opened 2017-02-10
If TemporaryDirectoryAllocator() fails, try again

MarkEWaite:
Multi-user system with a shared temporary directory (like many Linux
computers) may have several users running Jenkins unit tests.  The first
user to create /tmp/jenkinsTests.tmp owns it, and other users then are
unable to run their tests.

This change attempts to retain previous behaviors as much as possible.
If directory creation fails on the first attempt (jenkinsTest.tmp) and
then fails on the second attempt (Files.createTemporaryDirectory), the
base directory assignment will still be jenkinsTest.tmp as it was before.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
+++ b/src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
@@ -50,8 +50,15 @@ public class TemporaryDirectoryAllocator {
     }
 
     public TemporaryDirectoryAllocator() {
-        this.base = new File(System.getProperty("java.io.tmpdir"), "jenkinsTests.tmp");
-        base.mkdirs();
+        File tempDir = new File(System.getProperty("java.io.tmpdir"), "jenkinsTests.tmp");
+        if (tempDir.mkdirs() == false) {
+            try {
+                File tempDir2 = java.nio.file.Files.createTempDirectory("jenkinsTests").toFile();
+                tempDir = tempDir2;
+            } catch (IOException ioe) {
+            }
+        }
+        this.base = tempDir;
     }
 
     /**


src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
```diff
@@ -50,8 +50,15 @@ public TemporaryDirectoryAllocator(File base) {
     }
 
     public TemporaryDirectoryAllocator() {
-        this.base = new File(System.getProperty("java.io.tmpdir"), "jenkinsTests.tmp");
-        base.mkdirs();
+        File tempDir = new File(System.getProperty("java.io.tmpdir"), "jenkinsTests.tmp");
+        if (tempDir.mkdirs() == false) {
+            try {
+                File tempDir2 = java.nio.file.Files.createTempDirectory("jenkinsTests").toFile();
```
jglick: Why not just start with `createTempDirectory` to begin with?

MarkEWaite: Because createTempDirectory does not provide a fixed name, and I didn't want to risk impacting tests which inadvertently depended on a fixed name directory.  I assumed (possibly incorrectly) that some test might have wanted the fixed directory name, and if it is available and writable, then I didn't see any harm leaving it that way.

Windows users are typically single user on a box, so they don't mind having a fixed directory name in the temporary directory.

MacOS users are probably single user on a box, so they likely don't mind.

src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
```diff
@@ -50,8 +50,15 @@ public TemporaryDirectoryAllocator(File base) {
     }
 
     public TemporaryDirectoryAllocator() {
-        this.base = new File(System.getProperty("java.io.tmpdir"), "jenkinsTests.tmp");
-        base.mkdirs();
+        File tempDir = new File(System.getProperty("java.io.tmpdir"), "jenkinsTests.tmp");
+        if (tempDir.mkdirs() == false) {
```
jglick: More simply,

```java
if (!tempDir.mkdirs()) {
```

src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
```diff
@@ -50,8 +50,15 @@ public TemporaryDirectoryAllocator(File base) {
...
+            try {
+                File tempDir2 = java.nio.file.Files.createTempDirectory("jenkinsTests").toFile();
+                tempDir = tempDir2;
+            } catch (IOException ioe) {
```
jglick: Should at least be logged.

src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
jglick: Better to use `File.createTempFile` so we can run on Java 6, since currently this library uses `-target 6`. It simply forgot to use Animal Sniffer.

I am fine with updating to 7 or even 8 (does not really harm even use from plugins with old core baselines), but if we are going to do that, we should change the source level and introduce Animal Sniffer at once.

MarkEWaite:
There is already a constructor in the class which will take a File to use as the base.  I'm closing this pull request because I can easily switch to use the other constructor.  Sorry to waste your time reviewing it.  I can use what already exists, and don't need to modify this class at all.

PR #53, opened 2017-03-14
Deprecating authorizedTo in favor of a simpler asUser

jglick:
More flexible, and hardly more difficult, to use `MockAuthorizationStrategy` (introduced more recently than `CLICommandInvoker`).

@reviewbybees & @olivergondza


olivergondza:
I prefer keeping both ways as this can be convenient as a one-liner. +1

jglick:
> convenient as a one-liner

Yes, but the changes to global configuration can just make it harder to understand what the test is really doing. In https://github.com/jenkinsci/jenkins/pull/2790 I had to get rid of `authorizedTo` to avoid a subtle test bug.

PR #58, opened 2017-05-02
Make FakeChangeLogSCM support Run rather than just AbstractBuild.

abayer:
Needed for https://github.com/jenkinsci/workflow-job-plugin/pull/35 specifically but generally useful, I'd say.

cc @reviewbybees


src/main/java/org/jvnet/hudson/test/FakeChangeLogSCM.java
```diff
@@ -28,17 +28,22 @@
 import hudson.model.AbstractBuild;
```
jglick: Unused import.

abayer:
@jglick Updated to actually work and remove the unused imports. =)

PR #60, opened 2017-05-23
[FIXED JENKINS-44453] - JenkinsRule should ensure that Jenkins reaches the COMPLETED milestone.

oleg-nenashev:
It should make issues like https://issues.jenkins-ci.org/browse/JENKINS-37759 much more explicit in functional tests. From what I see there is no valid use-case for proceeding with the test if the initialization fails.

https://issues.jenkins-ci.org/browse/JENKINS-44453

@reviewbybees

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -206,6 +206,7 @@ import org.junit.runners.model.FrameworkMethod;
 import org.junit.runners.model.Statement;
 import com.gargoylesoftware.htmlunit.html.DomNodeUtil;
 import com.gargoylesoftware.htmlunit.html.HtmlFormUtil;
+import hudson.init.InitMilestone;
 import hudson.maven.MavenRequest;
 import hudson.model.Job;
 import hudson.model.queue.QueueTaskFuture;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -368,6 +369,13 @@ public void before() throws Throwable {
 
         try {
             jenkins = hudson = newHudson();
+            // If the initialization graph is corrupted, we cannot expect that Jenkins is in the good shape.
+            // Likely it is an issue in @Initializer() definitions (see JENKINS-37759).
+            // So we just fail the test.
+            if (jenkins.getInitLevel() != InitMilestone.COMPLETED) {
+                throw new Exception("Jenkins initialization has not reached the COMPLETED initialization stage. Current state is " + jenkins.getInitLevel() +
+                        ". Likely there is and issue with the Initialization task graph (e.g. usage of @Initializer(after = InitMilestone.COMPLETED)). See JENKINS-37759 for more info");
```
amuniz: Typo `and`

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -368,6 +369,13 @@ public void before() throws Throwable {
 
         try {
             jenkins = hudson = newHudson();
+            // If the initialization graph is corrupted, we cannot expect that Jenkins is in the good shape.
+            // Likely it is an issue in @Initializer() definitions (see JENKINS-37759).
+            // So we just fail the test.
+            if (jenkins.getInitLevel() != InitMilestone.COMPLETED) {
+                throw new Exception("Jenkins initialization has not reached the COMPLETED initialization stage. Current state is " + jenkins.getInitLevel() +
```
jglick: `AssertionError` would be appropriate.

oleg-nenashev: Maybe. I still want it to pass through the Exception handler, hence I decided to use common Exception

jglick: Which exception handler are you referring to?

jglick:
@reviewbybees done

@oleg-nenashev any particular timeline for merging & releasing, or is this just a back-burner task?

oleg-nenashev:
@jglick No need to merge it immediately, but AFAIK @amuniz hit this issue recently. I would just release it once there is any other ready change

PR #43, opened 2016-11-14
[JENKINS-24064] No more war-for-test

jglick:
Complements https://github.com/jenkinsci/jenkins/pull/2634.

@reviewbybees

PR #61, opened 2017-05-24
[JENKINS-24064] Replace war-for-test classifier with executable-war type

jglick:
Downstream of https://github.com/jenkinsci/maven-hpi-plugin/pull/65. Supersedes #43.

@reviewbybees


jglick:
No immediate reason to release this.

PR #62, opened 2017-06-12
[JENKINS-41631] Deleting the Maven Embedder dependency from the test harness

jglick:
Critical for [JENKINS-41631](https://issues.jenkins-ci.org/browse/JENKINS-41631) since some of the harder problems relate to random libraries pulled in via the `lib-jenkins-maven-embedder` dependency.

Taking a closer look at the harness code using this library, I could not find any realistic cases where it would actually be used. We already look up test dependencies from the classpath. Furthermore, the group IDs covered are old Hudson names no longer used by Jenkins.

@reviewbybees

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -62,8 +62,6 @@ import hudson.Main;
 import hudson.PluginManager;
 import hudson.Util;
 import hudson.WebAppMain;
-import hudson.maven.MavenEmbedder;
-import hudson.maven.MavenEmbedderException;
 import hudson.model.AbstractBuild;
 import hudson.model.AbstractProject;
 import hudson.model.BuildListener;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -2512,5 +2460,4 @@ private NameValuePair getCrumbHeaderNVP() {
         return new NameValuePair(jenkins.getCrumbIssuer().getDescriptor().getCrumbRequestField(),
                         jenkins.getCrumbIssuer().getCrumb( null ));
     }
-    public static final List<String> PLUGIN_GROUPIDS = new ArrayList<String>(Arrays.asList("org.jvnet.hudson.plugins", "org.jvnet.hudson.main"));
```
stephenc: On the basis of this line alone, I argue this is 100% safe to remove

jglick:
@reviewbybees done

oleg-nenashev:
Missed this PR. I am not sure it causes no regressions in plugins like Maven Project, but +1 for removing it taking the mess in the code into account.

PR #63, opened 2017-06-26
upgrade to jetty 9.4

olamy:
Signed-off-by: olivier lamy <olamy@apache.org>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@ THE SOFTWARE.
   <parent>
     <groupId>org.jenkins-ci</groupId>
     <artifactId>jenkins</artifactId>
-    <version>1.36</version>
+    <version>1.37</version>
   </parent>
 
   <groupId>org.jenkins-ci.main</groupId>
@@ -47,7 +47,9 @@ THE SOFTWARE.
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <jetty.version>9.2.15.v20160210</jetty.version>
+    <jetty.version>9.4.5.v20170502</jetty.version>
+    <maven-surefire-plugin.version>2.20</maven-surefire-plugin.version>
+    <maven-compiler-plugin.version>3.6.1</maven-compiler-plugin.version>
   </properties>
 
   <licenses>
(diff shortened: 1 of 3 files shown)


olamy:
@jglick @oleg-nenashev please review guys thanks!

pom.xml
```diff
@@ -47,7 +47,10 @@ THE SOFTWARE.
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <jetty.version>9.2.15.v20160210</jetty.version>
+    <jetty.version>9.4.5.v20170502</jetty.version>
+    <maven-surefire-plugin.version>2.20</maven-surefire-plugin.version>
+    <maven-compiler-plugin.version>3.6.1</maven-compiler-plugin.version>
```
oleg-nenashev: Why would you need the explicit versions here?

jglick: Also curious about this. The parent POM defines these versions IIUC.

jglick:
Not sure why CI build is broken; opened #64 to modernize it.

pom.xml
olamy: well it's really to use more up2date versions :-)

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -685,9 +686,11 @@ public Thread newThread(Runnable r) {
     protected LoginService configureUserRealm() {
         HashLoginService realm = new HashLoginService();
         realm.setName("default");   // this is the magic realm name to make it effective on everywhere
-        realm.update("alice", new Password("alice"), new String[]{"user","female"});
-        realm.update("bob", new Password("bob"), new String[]{"user","male"});
-        realm.update("charlie", new Password("charlie"), new String[]{"user","male"});
+        UserStore userStore = new UserStore();
```
jglick: Forced by Jetty changes as detailed in https://github.com/eclipse/jetty.project/pull/1488#issuecomment-332201285. This makes it awkward to switch Jetty versions in a plugin back to 9.2.x, which @vivek says is currently necessary for Wiremock compatibility: you must use

```java
@Rule
public JenkinsRule r = new JenkinsRule() {
    @Override
    protected LoginService configureUserRealm() {
        HashLoginService realm = new HashLoginService();
        realm.setName("default");
        return realm;
    }
};
```

We cannot simply delete the `UserStore` part from the default implementation; it works for most tests, but not those calling `WebClient.login` while using `LegacySecurityRealm` (perhaps indirectly via `@PresetData(DataSet.NO_ANONYMOUS_READACCESS)`), of which there are a few in core and some plugins. (For example, `JnlpAccessWithSecuredHudsonTest`, using `HudsonTestCase`.)

PR #66, opened 2017-08-01
Adding a new RestartableJenkinsRule method comfortable for use with Java 8 test sources

jglick:
@reviewbybees

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
@@ -10,6 +10,7 @@ import org.junit.runners.model.Statement;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.Callable;
 
 /**
  * Provides a pattern for executing a sequence of steps.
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
```diff
@@ -75,6 +76,27 @@ public void evaluate() throws Throwable {
...
     }
 
+    /**
+     * One step to run, intended to be a SAM for lambdas.
+     * {@link Closure} does not work because it is an abstract class, not an interface.
+     * {@link Callable} of {@link Void} does not work because you have to return null.
+     * {@link Runnable} does not work because it throws no checked exceptions.
+     * {@code Consumer} is the same, and is not present in Java 7.
+     * Other candidates had similar issues.
+     */
```
oleg-nenashev: `@since TODO`

src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
```diff
@@ -75,6 +76,27 @@ public void evaluate() throws Throwable {
...
+    // TODO Java 8: @FunctionalInterface
+    public interface Step {
+        void run(JenkinsRule r) throws Throwable;
+    }
+    public void then(final Step s) {
```
oleg-nenashev: Javadoc?

PR #68, opened 2017-09-13
Make LoggerRule use the nicer formatter developed for support-core

jglick:
Makes lengthy test output more readable. Useful for example in the test added in https://github.com/jenkinsci/jenkins/pull/3019.

@reviewbybees


jglick:
`LoggingSupport.DEFAULT_FORMAT` in the JRE is very visually awkward when you have a lot of messages, due to the two-line layout.

PR #71, opened 2017-09-18
Suppressing warning from slf4j

jglick:
Not really sure what [this](https://www.slf4j.org/codes.html#release) is all about, but not of interest in the context of functional tests, so getting rid of an annoying warning message.

@reviewbybees

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -209,6 +209,7 @@ import java.io.FileNotFoundException;
 import java.net.HttpURLConnection;
 import java.nio.channels.ClosedByInterruptException;
 import jenkins.model.ParameterizedJobMixIn;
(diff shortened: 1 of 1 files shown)


jglick:
@jtnord said he thought the warning might reflect an incorrect classpath, in which case this should possibly be solved differently.

jtnord:
So IIRC you get this warning if you have both `commons-logging` and the `jcl-over-slf4j` on the classpath and the commons-logging one happens to be used first.

I think the solution is to deploy a commons-logging version 0 and then have a strict dependency on that and make sure that jcl-over-slf4j wins every time (which implements the same API).  Will investigate later.  This would also have the added benifit that Jenkins can rely on this and then that would prevent any plugins accidentally depending on and including commons-logging in their hpi.

jtnord:
@jglik where did you see this error - what which repo where you working with?

jglick:
> where did you see this error - what which repo where you working with?

@jtnord visible in pretty much any test in any plugin, IIRC.

PR #70, opened 2017-09-18
More readable timestamps in log messages during tests

jglick:
Amends #68. Before:

```
=== Starting coalescedQueue(org.jenkinsci.plugins.workflow.support.steps.build.BuildTriggerStepTest)
Picking up existing exploded jenkins.war at /…/jenkinsci/pipeline-build-step-plugin/target/jenkins-for-test
2017-09-18 17:29:16.030+0000 [id=15]	INFO	o.jvnet.hudson.test.JenkinsRule#createWebServer: Running on http://localhost:37353/jenkins/
2017-09-18 17:29:16.212+0000 [id=22]	INFO	jenkins.InitReactorRunner$1#onAttained: Started initialization
2017-09-18 17:29:21.983+0000 [id=27]	INFO	jenkins.InitReactorRunner$1#onAttained: Listed all plugins
2017-09-18 17:29:22.040+0000 [id=28]	INFO	jenkins.InitReactorRunner$1#onAttained: Prepared all plugins
2017-09-18 17:29:22.080+0000 [id=22]	INFO	jenkins.InitReactorRunner$1#onAttained: Started all plugins
2017-09-18 17:29:25.884+0000 [id=22]	INFO	jenkins.InitReactorRunner$1#onAttained: Augmented all extensions
2017-09-18 17:29:25.898+0000 [id=26]	INFO	jenkins.InitReactorRunner$1#onAttained: Loaded all jobs
2017-09-18 17:29:26.261+0000 [id=25]	INFO	o.j.main.modules.sshd.SSHD#start: Started SSHD at port 39621
2017-09-18 17:29:26.261+0000 [id=23]	INFO	jenkins.InitReactorRunner$1#onAttained: Completed initialization
2017-09-18 17:29:26.274+0000 [id=47]	INFO	jenkins.InitReactorRunner$1#onAttained: Started initialization
2017-09-18 17:29:26.276+0000 [id=50]	INFO	jenkins.InitReactorRunner$1#onAttained: Listed all plugins
2017-09-18 17:29:26.282+0000 [id=47]	INFO	jenkins.InitReactorRunner$1#onAttained: Prepared all plugins
2017-09-18 17:29:26.283+0000 [id=47]	INFO	jenkins.InitReactorRunner$1#onAttained: Started all plugins
2017-09-18 17:29:26.283+0000 [id=47]	INFO	jenkins.InitReactorRunner$1#onAttained: Augmented all extensions
[...]

PR #73, opened 2017-09-19
[JENKINS-45245] Work around failure of IDEA to add jenkins-war.war to the test CP

jglick:
Possible fix of [JENKINS-45245](https://issues.jenkins-ci.org/browse/JENKINS-45245). Not tested with actual IDEA, since I do not use IDEA and would not even know how to set it up to see the problem to begin with. But does the right thing if `"/winstone.jar"` is temporarily mangled to something nonexistent.

@reviewbybees @jenkinsci/code-reviewers

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/WarExploder.java
+++ b/src/main/java/org/jvnet/hudson/test/WarExploder.java
@@ -23,13 +23,15 @@
  */
 package org.jvnet.hudson.test;
 
-import hudson.remoting.Which;
 import hudson.FilePath;
-
+import hudson.remoting.Which;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.net.URL;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import jenkins.model.Jenkins;
 
 /**
  * Ensures that <tt>jenkins.war</tt> is exploded.
@@ -41,6 +43,8 @@ import java.net.URL;
  * @author Kohsuke Kawaguchi
  */
 public final class WarExploder {
+
+    private static final Logger LOGGER = Logger.getLogger(WarExploder.class.getName());
 
     public static File getExplodedDir() throws Exception {
         // rethrow an exception every time someone tries to do this, so that when explode()
@@ -74,18 +78,31 @@ public final class WarExploder {
             if(new File(d,".jenkins").exists()) {
                 File dir = new File(d,"war/target/jenkins");
                 if(dir.exists()) {
-                    System.out.println("Using jenkins.war resources from "+dir);
+                    LOGGER.log(Level.INFO, "Using jenkins.war resources from {0}", dir);
                     return dir;
                 }
             }
         }
 
         // locate jenkins.war
+        File war;
         URL winstone = WarExploder.class.getResource("/winstone.jar");
-        if(winstone==null)
-            // impossible, since the test harness pulls in jenkins.war
-            throw new AssertionError("jenkins.war is not in the classpath. If you are running this from the core workspace, run 'mvn install' to create the war image in war/target/jenkins. If running from a plugin, try using `mvn clean test` (cf. JENKINS-45245).");
-        File war = Which.jarFile(Class.forName("executable.Executable"));
+        if (winstone != null) {
+            war = Which.jarFile(Class.forName("executable.Executable"));
+        } else {
+            // JENKINS-45245: work around incorrect test classpath in IDEA. Note that this will not correctly handle timestamped snapshots; in that case use `mvn test`.
+            File core = Which.jarFile(Jenkins.class); // will fail with IAE if have neither jenkins-war.war nor jenkins-core.jar in ${java.class.path}
+            String version = core.getParentFile().getName();
+            if (core.getName().equals("jenkins-core-" + version + ".jar") && core.getParentFile().getParentFile().getName().equals("jenkins-core")) {
+                war = new File(new File(new File(core.getParentFile().getParentFile().getParentFile(), "jenkins-war"), version), "jenkins-war-" + version + ".war");
+                if (!war.isFile()) {
+                    throw new AssertionError(war + " does not yet exist. Prime your development environment by running `mvn validate`.");
+                }
+                LOGGER.log(Level.INFO, "{0} is the continuation of the classpath by other means", war);
+            } else {
+                throw new AssertionError(core + " is not in the expected location, and jenkins-war-*.war was not in " + System.getProperty("java.class.path"));
+            }
+        }
 
         // TODO this assumes that the CWD of the Maven process is the plugin ${basedir}, which may not be the case
         File buildDirectory = new File(System.getProperty("buildDirectory", "target"));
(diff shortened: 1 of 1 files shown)


jglick:
BTW to test this locally you should be able to just

```xml
<jenkins-test-harness.version>2.27-20170919.194540-1</jenkins-test-harness.version>
```

assuming IDEA at least supports timestamped snapshots (perhaps after an initial `mvn validate` priming).

src/main/java/org/jvnet/hudson/test/WarExploder.java
```diff
@@ -74,18 +78,31 @@ private static File explode() throws Exception {
...
         URL winstone = WarExploder.class.getResource("/winstone.jar");
-        if(winstone==null)
-            // impossible, since the test harness pulls in jenkins.war
-            throw new AssertionError("jenkins.war is not in the classpath. If you are running this from the core workspace, run 'mvn install' to create the war image in war/target/jenkins. If running from a plugin, try using `mvn clean test` (cf. JENKINS-45245).");
-        File war = Which.jarFile(Class.forName("executable.Executable"));
+        if (winstone != null) {
+            war = Which.jarFile(Class.forName("executable.Executable"));
+        } else {
+            // JENKINS-45245: work around incorrect test classpath in IDEA. Note that this will not correctly handle timestamped snapshots; in that case use `mvn test`.
+            File core = Which.jarFile(Jenkins.class); // will fail with IAE if have neither jenkins-war.war nor jenkins-core.jar in ${java.class.path}
```
abayer: s/IAE/IDEA/

abayer:
Double-checked with a different plugin than I'd been doing experiments with, and yup, looks good.

src/main/java/org/jvnet/hudson/test/WarExploder.java
jglick: Nope, `IllegalArgumentException`.

abayer: Might be worth spelling that out, then, 'cos `IDEA` a line above `IAE` is, as demonstrated here, potentially confusing. =)

jglick:
Let us give some other folks a chance to test/review, but if it still looks good, we should be able to cut a JTH release tomorrow-ish (I would not mind getting #70 & #71 in too) and integrate into `plugin-pom`.

src/main/java/org/jvnet/hudson/test/WarExploder.java
svanoort: Seconded.

src/main/java/org/jvnet/hudson/test/WarExploder.java
```diff
@@ -74,18 +78,31 @@ private static File explode() throws Exception {
...
+        } else {
+            // JENKINS-45245: work around incorrect test classpath in IDEA. Note that this will not correctly handle timestamped snapshots; in that case use `mvn test`.
+            File core = Which.jarFile(Jenkins.class); // will fail with IllegalArgumentException if have neither jenkins-war.war nor jenkins-core.jar in ${java.class.path}
+            String version = core.getParentFile().getName();
+            if (core.getName().equals("jenkins-core-" + version + ".jar") && core.getParentFile().getParentFile().getName().equals("jenkins-core")) {
+                war = new File(new File(new File(core.getParentFile().getParentFile().getParentFile(), "jenkins-war"), version), "jenkins-war-" + version + ".war");
+                if (!war.isFile()) {
+                    throw new AssertionError(war + " does not yet exist. Prime your development environment by running `mvn validate`.");
+                }
+                LOGGER.log(Level.INFO, "{0} is the continuation of the classpath by other means", war);
```
svanoort: 🐜  Not a terribly meaningful log message and probably not necessary to most users.  I think what you're trying to say is something like "Loaded Jenkins core classes from JAR {0}" -- probably at a CONFIG or FINE log level, just for specialty debugging.

src/main/java/org/jvnet/hudson/test/WarExploder.java
```diff
@@ -74,18 +78,31 @@ private static File explode() throws Exception {
...
+        File war;
         URL winstone = WarExploder.class.getResource("/winstone.jar");
-        if(winstone==null)
-            // impossible, since the test harness pulls in jenkins.war
-            throw new AssertionError("jenkins.war is not in the classpath. If you are running this from the core workspace, run 'mvn install' to create the war image in war/target/jenkins. If running from a plugin, try using `mvn clean test` (cf. JENKINS-45245).");
-        File war = Which.jarFile(Class.forName("executable.Executable"));
+        if (winstone != null) {
+            war = Which.jarFile(Class.forName("executable.Executable"));
+        } else {
+            // JENKINS-45245: work around incorrect test classpath in IDEA. Note that this will not correctly handle timestamped snapshots; in that case use `mvn test`.
```
daniel-beck: Unclear what this means. Will a `jenkins.version` snapshot dependency break with this? If so, is there a way to detect it here and output this comment as an explicit error message (perhaps as part of the assertion error on 99?)

hoshsadiq:
Seems to work for me

src/main/java/org/jvnet/hudson/test/WarExploder.java
jglick: It just means that given a `jenkins-core.version` of `2.99-20180101.1234` or whatever and `jenkins-war.version` `2.99-20180101.5678`, from the perspective of `WarExploder` we only know that we are running out of `…/jenkins-core/2.99-SNAPSHOT/jenkins-core-2.99-SNAPSHOT.jar` and thus that we should load `…/jenkins-war/2.99-SNAPSHOT/jenkins-war-2.99-SNAPSHOT.war`. It does not necessarily know the specific version to use.

Basically, a locally built snapshot dep should be fine, but do not expect a checkout of a downstream plugin PR using an upstream core PR to allow you to run tests from IDEA without using `mvn test`. Which you would rarely want to do anyway, since if you were running tests locally, it is probably because you are testing local changes including to core.

jglick: Of course all of this is speculative—I cannot myself try it.

daniel-beck: Perhaps `if (version.contains("SNAPSHOT")) { /* log something saying we might pick up the wrong war *? }`? But this can go ahead without that, could always be added later.

jglick: Yeah I would leave this to an IDEA user actually trying to use snapshot core deps to further test & refine.

src/main/java/org/jvnet/hudson/test/WarExploder.java
```diff
@@ -107,9 +124,12 @@ private static File explode() throws Exception {
...
             new File(explodeDir + ".exploding").delete();
         } else {
-            System.out.println("Picking up existing exploded jenkins.war at "+explodeDir.getAbsolutePath());
+            LOGGER.log(Level.INFO, "Picking up existing exploded jenkins.war at {0}", explodeDir.getAbsolutePath());
         }
 
         return explodeDir;
     }
+
+    private WarExploder() {}
```
oleg-nenashev: Unrelated, it may probably cause issues in some custom implementations which instantiate War Exploder due to whatever reason.

I suggest reverting it before merging

src/main/java/org/jvnet/hudson/test/WarExploder.java
oleg-nenashev: Fine for me.

src/main/java/org/jvnet/hudson/test/WarExploder.java
jglick: It consists only of a `static` method. There is no reason to ever instantiate this class. That is the purpose of the `private` constructor—to clarify this.

Anyway we do not guarantee binary compatibility for test harnesses. Upgrades are done only by developers.

PR #74, opened 2017-09-25
Fixing LoggerRule

jglick:
Corrects a regression in #68 to display `FINE`—`FINEST` messages again; and also improves on it to apply the format to `INFO`+ messages unconditionally.

@reviewbybees


jglick:
@reviewbybees done

PR #79, opened 2017-10-11
[JENKINS-47393] Remove signature and runtime references to CommandLauncher

jglick:
For https://github.com/jenkinsci/jenkins/pull/3076.

@reviewbybees


jglick:
> introducing bridge methods

Oh no.

> introduces a breaking change in the JTH library

Unlikely to be source-incompatible (not that we care greatly), and binary incompatibility is irrelevant.

jglick:
> things like PCT, which probably bump JTH

AFAIK it does not. That was in fact one of the benefits of splitting `jenkins-test-harness` into its own repository and giving it an independent lifecycle: that a plugin POM could declare a specific `jenkins-test-harness.version` which would remain consistent even when running with a newer `jenkins.version` during something like PCT (or `jenkinsVersions` of `buildPlugin`, etc.).

PR #82, opened 2017-11-07
Add a Spy for the SecurityListener

Wadeck:
- as requested in [#3074](https://github.com/jenkinsci/jenkins/pull/3074), the code is proposed in JTH to provide test helper for all plugins

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/security/SpySecurityListener.java
+++ b/src/main/java/jenkins/security/SpySecurityListener.java
@@ -1,0 +1,142 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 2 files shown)


jglick:
Not sure I understand the motivation for filing this. https://github.com/jenkinsci/jenkins/pull/3074 is not using it. If that were reworked to pick this up, _and_ there were some other PR (in a plugin) which could take advantage of it, then it would make sense.

Wadeck:
@jglick each time the `SecurityListener` is used this helper class could be used to ease the related tests.
- [#22 in sshd](https://github.com/jenkinsci/sshd-module/pull/22)
- [#87 in github-oauth](https://github.com/jenkinsci/github-oauth-plugin/pull/87)
- all other plugin that want to test their usage of authentication events.

As we discussed about the `SecurityListener`, concerning the new abstract methods instead of the mini-design, this PR will be modified accordingly.

jglick:
Putting on hold since it is not actually being used anywhere yet, so for now this is more of a prototype.

Wadeck:
> Still something you want to do?

@jglick since that PR I haven't had to deal with this problem. I will close the PR for the moment.

PR #83, opened 2017-11-08
Add multiple methods to ease the use of Basic & ApiToken

Wadeck:
As requested in [#3129 jenkins-core](https://github.com/jenkinsci/jenkins/pull/3129), some helper methods to ease the use of token (Basic) with username ( password or API Token.

This will allow tests to be written with less code (better readability).

@reviewbybees @jglick

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,7 +79,8 @@ THE SOFTWARE.
     <dependency>
       <groupId>org.jenkins-ci.main</groupId>
       <artifactId>jenkins-war</artifactId>
-      <version>1.580.1</version>
+      <!--to have access to User.getById-->
+      <version>1.651.2</version>
       <type>executable-war</type>
       <exclusions>
         <exclusion>
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -49,6 +49,8 @@ import com.gargoylesoftware.htmlunit.javascript.host.xml.XMLHttpRequest;
 import com.gargoylesoftware.htmlunit.util.NameValuePair;
 import com.gargoylesoftware.htmlunit.util.WebResponseWrapper;
 import com.gargoylesoftware.htmlunit.xml.XmlPage;
+import com.google.common.net.HttpHeaders;
+import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.ClassicPluginStrategy;
 import hudson.CloseProofOutputStream;
 import hudson.DNSMultiCast;
@@ -103,6 +105,7 @@ import hudson.tasks.Publisher;
 import hudson.tools.ToolProperty;
 import hudson.util.PersistedList;
 import hudson.util.ReflectionUtils;
+import hudson.util.Scrambler;
 import hudson.util.StreamTaskListener;
 import hudson.util.jna.GNUCLibrary;
 import java.beans.Introspector;
@@ -160,6 +163,7 @@ import javax.servlet.ServletContextEvent;
 import jenkins.model.Jenkins;
 import jenkins.model.JenkinsAdaptor;
 import jenkins.model.JenkinsLocationConfiguration;
+import jenkins.security.ApiTokenProperty;
 import net.sf.json.JSONObject;
 import net.sourceforge.htmlunit.corejs.javascript.Context;
 import net.sourceforge.htmlunit.corejs.javascript.ContextFactory;
@@ -2309,6 +2313,46 @@ public class JenkinsRule implements TestRule, MethodRule, RootAction {
             String crumb = issuer.getCrumb(null);
 
             return new URL(getContextPath()+relativePath+"?"+crumbName+"="+crumb);
+        }
+
+        /**
+         * Add the "Authorization" header with Basic credentials derived from login and password using Base64
+         * @since TODO
+         */
+        public @NonNull WebClient usingBasicCredentials(@NonNull String login, @NonNull String passwordOrToken){
+            addRequestHeader(HttpHeaders.AUTHORIZATION, "Basic " + Scrambler.scramble(login + ":" + passwordOrToken));
+            return this;
+        }
+
+        /**
+         * Retrieve the {@link ApiTokenProperty} from the user, derive credentials from it and place it in Basic authorization header
+         * @see #usingBasicCredentials(String, String)
+         * @since TODO
+         */
+        public @NonNull WebClient usingBasicApiToken(@NonNull User user){
+            return usingBasicCredentials(user.getId(), user.getProperty(ApiTokenProperty.class).getApiToken());
+        }
+
+        /**
+         * Retrieve the {@link ApiTokenProperty} from the associated user, derive credentials from it and place it in Basic authorization header
+         * @see #usingBasicCredentials(String, String)
+         * @since TODO
+         */
+        public @NonNull WebClient usingBasicApiToken(@NonNull String userId){
+            User u = User.getById(userId, false);
+            assertNotNull("The userId must correspond to an already created User", u);
+            return usingBasicApiToken(u);
+        }
+
+        /**
+         * Remove the "Authorization" header from the client
+         * Meant to be used after having configured the authentication using a token
+         * @see #usingBasicCredentials(String, String)
+         * @since TODO
+         */
+        public @NonNull WebClient removeBasicAuthorizationHeader(){
+            this.removeRequestHeader(HttpHeaders.AUTHORIZATION);
+            return this;
         }
 
         /**
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -50,8 +50,6 @@ import com.gargoylesoftware.htmlunit.util.NameValuePair;
 import com.gargoylesoftware.htmlunit.util.WebResponseWrapper;
 import com.gargoylesoftware.htmlunit.xml.XmlPage;
 import com.google.common.net.HttpHeaders;
-import com.trilead.ssh2.crypto.Base64;
-import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.ClassicPluginStrategy;
 import hudson.CloseProofOutputStream;
 import hudson.DNSMultiCast;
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -79,7 +79,8 @@ THE SOFTWARE.
     <dependency>
       <groupId>org.jenkins-ci.main</groupId>
       <artifactId>jenkins-war</artifactId>
-      <version>1.580.1</version>
+      <!--to have access to User.getById-->
+      <version>1.651.2</version>
```
oleg-nenashev: In such case we will need to bump the default baseline in Plugin POM. BTW, Plugin POM 3.x already did it: https://github.com/jenkinsci/plugin-pom/blob/master/CHANGELOG.md#30

oleg-nenashev: IIUC the changelog, it's safe to bump the baseline to 2.7.x at least. CC @batmat

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -2312,6 +2316,76 @@ public URL createCrumbedUrl(String relativePath) throws IOException {
         }
 
         /**
+         * Add the "Authorization" header with Basic credentials derived from login and password using Base64
+         * @since TODO
+         */
+        public @NonNull WebClient usingBasicCredentials(@NonNull String login, @NonNull String passwordOrToken){
```
oleg-nenashev: What is the reason of having both "using..." and "with..."?

src/test/java/org/jvnet/hudson/test/JenkinsRuleTest.java
```diff
@@ -92,4 +182,27 @@ public String getSetterParam() {
             return setterParam;
         }
     }
+
+    @TestExtension
```
oleg-nenashev: Maybe it should be linked to a particular test case. "WhoAmI" name seems to be copy-pasta

Wadeck: Name changed + link added to the new test

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
Wadeck: New design proposed, with an optional closure as last parameter. This will avoid the confusion between the `AutoCloseable` (that remove the header) and the `WebClient` (that is already `AutoCloseable`).
Additionnally it will simplify the readability of the tests.

pom.xml
jglick: `plugin-pom` 3.0 did not really drop support for Jenkins 1; it just stopped enforcing a check on usages on Servlet APIs. Switching _this_ dependency would really make it impossible to use JTH, and thus a `plugin-pom` incorporating it (without overriding JTH version), on a plugin with an older baseline.

I am not opposed to doing that; maybe now is the time. https://github.com/jenkinsci/plugin-pom/pull/83#pullrequestreview-72362389

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -2312,6 +2316,103 @@ public URL createCrumbedUrl(String relativePath) throws IOException {
         }
 
         /**
+         * Add the "Authorization" header with Basic credentials derived from login and password using Base64
+         * @since TODO
+         */
+        public @NonNull WebClient withBasicCredentials(@NonNull String login, @NonNull String passwordOrToken) {
+            addRequestHeader(HttpHeaders.AUTHORIZATION, "Basic " + Scrambler.scramble(login + ":" + passwordOrToken));
```
jglick: Do not use `Scrambler`. Use `Base64`.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -2312,6 +2316,103 @@ public URL createCrumbedUrl(String relativePath) throws IOException {
...
+            return this;
+        }
+
+        /**
+         * Add the "Authorization" header with Basic credentials derived from login and password using Base64
+         * After the closure being called, the header is removed
```
jglick: Seems overly complicated. You can simply add a header to a single request. Or just call `createWebClient` again to get a fresh one without authentication, as we would already have done for `.login` calls.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -50,6 +50,7 @@
 import com.gargoylesoftware.htmlunit.util.WebResponseWrapper;
 import com.gargoylesoftware.htmlunit.xml.XmlPage;
 import com.google.common.net.HttpHeaders;
+import com.trilead.ssh2.crypto.Base64;
```
jglick: Why not use [the standard version](https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html)?

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -110,12 +111,7 @@
 import hudson.util.jna.GNUCLibrary;
 import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
-import java.io.BufferedReader;
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.OutputStream;
+import java.io.*;
```
jglick: please avoid wildcard imports

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -2320,23 +2315,15 @@ public URL createCrumbedUrl(String relativePath) throws IOException {
...
-                closure.call(this);
-            }
-            finally{
-                this.removeBasicAuthorizationHeader();
+            String authCode;
+            try {
+                authCode = new String(Base64.encode((login + ":" + passwordOrToken).getBytes("UTF-8")));
+            } catch (UnsupportedEncodingException e) {
+                // impossible
+                throw new Error(e);
```
jglick: `StandardCharsets.UTF_8`

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -49,6 +49,9 @@
 import com.gargoylesoftware.htmlunit.util.NameValuePair;
 import com.gargoylesoftware.htmlunit.util.WebResponseWrapper;
 import com.gargoylesoftware.htmlunit.xml.XmlPage;
+import com.google.common.net.HttpHeaders;
+import com.trilead.ssh2.crypto.Base64;
+import edu.umd.cs.findbugs.annotations.NonNull;
```
jglick: Use `javax.annotation.Nonnull`, already imported.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -2312,6 +2311,63 @@ public URL createCrumbedUrl(String relativePath) throws IOException {
...
+            return withBasicApiToken(user);
+        }
+
+        /**
+         * Remove the "Authorization" header from the client
+         * Meant to be used after having configured the authentication using a token
+         * @see #withBasicCredentials(String, String)
+         * @since TODO
+         */
+        public @NonNull WebClient removeBasicAuthorizationHeader(){
```
jglick: Probably unnecessary (you can just `createWebClient` again), but OK.

src/test/java/org/jvnet/hudson/test/JenkinsRuleTest.java
```diff
@@ -92,4 +155,27 @@ public String getSetterParam() {
...
+        }
+
+        @Override
+        public String getUrlName() {
+            return "test";
+        }
+
+        public HttpResponse doIndex() {
+            User u = User.current();
+            return HttpResponses.plainText(u!=null ? u.getId() : "anonymous");
```
jglick: BTW there is also a standard `/whoAmI` though the response is HTML so it is probably less convenient for purposes of this test. Or you can use `/me/api/xml?xpath=/*/id/text()`.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
Wadeck: IDE configured :)

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
Wadeck: Miss-import 😞

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
Wadeck: Thank you for the information 👍

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
Wadeck: You convinced me, I was not satisfied with the name btw.

src/test/java/org/jvnet/hudson/test/JenkinsRuleTest.java
Wadeck: The usage of XPath in the url throw me a 403, it requires SecureRequester or something like that. I didn't investigate more, I just used `whoAmI/api/json` and `JSONObject` to do the job.

Wadeck:
All the tests passed on the CI, so the integration of the snapshot is working like a charm

src/test/java/org/jvnet/hudson/test/JenkinsRuleTest.java
```diff
@@ -113,22 +114,23 @@ public void testTokenHelperMethods() throws Exception {
...
+        wc = j.createWebClient();
         wc.withBasicApiToken("charlotte");
         makeRequestAndAssertLogin(wc, "charlotte");
     }
 
     private void makeRequestAndAssertLogin(JenkinsRule.WebClient wc, String expectedLogin) throws IOException, SAXException {
-        WebRequest req = new WebRequest(new URL(j.getURL(),"test"));
-        req.setEncodingType(null);
+        WebRequest req = new WebRequest(new URL(j.getURL(),"whoAmI/api/json"));
         Page p = wc.getPage(req);
```
jglick: or just `goTo` I think

PR #84, opened 2017-11-08
Allow connector to listen in all interfaces

carlossg:
In the kubernetes plugin we need the server to listen in all interfaces to connect external agents.

Currently working by copying the class https://github.com/jenkinsci/kubernetes-plugin/blob/master/src/test/java/org/jvnet/hudson/test/JenkinsRuleNonLocalhost.java#L83

@reviewbybees

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -673,7 +673,7 @@ public class JenkinsRule implements TestRule, MethodRule, RootAction {
         HttpConfiguration config = connector.getConnectionFactory(HttpConnectionFactory.class).getHttpConfiguration();
         // use a bigger buffer as Stapler traces can get pretty large on deeply nested URL
         config.setRequestHeaderSize(12 * 1024);
-        connector.setHost("localhost");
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -673,7 +673,7 @@ public Thread newThread(Runnable r) {
         HttpConfiguration config = connector.getConnectionFactory(HttpConnectionFactory.class).getHttpConfiguration();
         // use a bigger buffer as Stapler traces can get pretty large on deeply nested URL
         config.setRequestHeaderSize(12 * 1024);
-        connector.setHost("localhost");
+        connector.setHost(System.getProperty("host", "localhost"));
```
jtnord: Seems like this would be bad, it opens up a potentially insecure instance on public interfaces, which is a drive by owning of your machine

carlossg:
it is a bad idea if you run it in an open network with 0.0.0.0 
I'm not saying set is as the default but at least make it configurable for anybody that knows what they are doing

oleg-nenashev:
@carlossg please feel free to reopen it if you are still interested

PR #85, opened 2017-11-09
Abstract class for RootAction

Wadeck:
- to ease the inner implementation that often return null in two of the methods
- compared to the `InvisibleAction` present in core, it's just a facility (the methods are not final)
- impacted test cases in jenkins-core:
  - `hudson.model.EnvironnementContributor.EnvVarsHtml`
  - `lib.form.FormTest.`[`AutocompleteOffByDefault`, `AutocompleteOnWhenTrue` and `InputsCanSetAutocomplete`]
  - `jenkins.util.FullDuplexHttpServiceTest.Endpoint`
  - `jenkins.model.JenkinsTest.ProtectedRootActionImpl`
  - (present only in [#3129](https://github.com/jenkinsci/jenkins/pull/3129)) `lib.form.PasswordTest.SecretNotPlainText`
  - `lib.form.RowSetTest.Subject`

Another idea could be to use default method in the interface directly, but this PR could be a solution for plugin that does want to require last core version.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/model/AbstractRootAction.java
+++ b/src/main/java/hudson/model/AbstractRootAction.java
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2004-2009, Sun Microsystems, Inc.
+ * Copyright (c) 2017, CloudBees, Inc.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


Wadeck:
As Jesse is OK with the default implementation of the interface, I proposed this other PR: [#3172 in core](https://github.com/jenkinsci/jenkins/pull/3172).

Wadeck:
@dwnusbaum thank you for the review, I close this one so :)

PR #86, opened 2017-11-10
[JENKINS-47975] Add utility methods to LoggerRule for matching log entries

dwnusbaum:
See [JENKINS-47975](https://issues.jenkins-ci.org/browse/JENKINS-47975).

Adds utility methods to `LoggerRule` so that it can be used naturally with `assertThat`. Code taken from @rsandell and `CaptureLogRule` in jenkinsci/ldap-plugin/pull/27.

@reviewbybees

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/LoggerRule.java
+++ b/src/main/java/org/jvnet/hudson/test/LoggerRule.java
@@ -36,6 +36,10 @@ import java.util.logging.Level;
 import java.util.logging.LogRecord;
 import java.util.logging.Logger;
 import java.util.logging.SimpleFormatter;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeMatcher;
 import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.rules.ExternalResource;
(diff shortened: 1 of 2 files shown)


dwnusbaum:
That makes sense. I made `recorded` a static method so `assertThat(logRule, logRule.recorded(...))` can be simplified to `assertThat(logRule, recorded(...))` with a static import.

jglick:
Not going to review until there is a consumer PR.

dwnusbaum:
Now consumed by jenkinsci/ldap-plugin/pull/27.

src/main/java/org/jvnet/hudson/test/LoggerRule.java
```diff
@@ -162,4 +166,97 @@ protected void after() {
...
     }
 
+    /**
+     * Creates a {@link Matcher} that matches if the {@link LoggerRule} has a {@link LogRecord} at
+     * the specified {@link Level}, with a message matching the specified matcher, and with a
+     * {@link Throwable} matching the specified matcher.
+     *
+     * @param level The {@link Level} of the {@link LoggerRule} to match. Pass {@code null} to match any {@link Level}.
+     * @param message the matcher to match against {@link LogRecord#getMessage}
+     * @param thrown the matcher to match against {@link LogRecord#getThrown()}
```
jglick: or null to match against any throwable, or none (as opposed to passing a matcher that matches no throwable)

src/main/java/org/jvnet/hudson/test/LoggerRule.java
```diff
@@ -162,4 +166,97 @@ protected void after() {
...
+    /**
+     * Creates a {@link Matcher} that matches if the {@link LoggerRule} has a {@link LogRecord} at
+     * the specified {@link Level}, with a message matching the specified matcher, and with a
+     * throwable matching the specified matcher.
+     *
+     * @param level The {@link Level} of the {@link LoggerRule} to match. Pass {@code null} to match any {@link Level}.
+     * @param message the matcher to match against {@link LogRecord#getMessage}
+     * @param thrown the matcher to match against {@link LogRecord#getThrown()}
+     */
+    public static Matcher<LoggerRule> recorded(@CheckForNull Level level, @Nonnull Matcher<String> message, @CheckForNull Matcher<Throwable> thrown) {
```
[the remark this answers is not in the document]
jglick: Agreed.

src/main/java/org/jvnet/hudson/test/LoggerRule.java
```diff
@@ -162,4 +166,97 @@ protected void after() {
         }
     }
 
+    /**
+     * Creates a {@link Matcher} that matches if the {@link LoggerRule} has a {@link LogRecord} at
+     * the specified {@link Level}, with a message matching the specified matcher, and with a
+     * {@link Throwable} matching the specified matcher.
+     *
+     * @param level The {@link Level} of the {@link LoggerRule} to match. Pass {@code null} to match any {@link Level}.
```
jglick: Or I suppose this could be a `Matcher<Level>`, though that seems like overkill.

src/main/java/org/jvnet/hudson/test/LoggerRule.java
```diff
@@ -162,4 +166,97 @@ protected void after() {
         }
     }
 
+    /**
+     * Creates a {@link Matcher} that matches if the {@link LoggerRule} has a {@link LogRecord} at
```
jglick: Should remind the caller to have called `capture` first, as `getRecords` does.

src/main/java/org/jvnet/hudson/test/LoggerRule.java
```diff
@@ -162,4 +166,97 @@ protected void after() {
...
+            }
+            return false;
+        }
+
+        @Override
+        public void describeTo(org.hamcrest.Description description) {
+            description.appendText("has LogRecord");
+            if (level != null) {
+                description.appendText(" with level ");
+                description.appendValue(level.getName());
```
jglick: Awkward repetition. It would be nice if `Matcher` had an API which specified just one method, a `@CheckForNull String` returning null for a match and an explanation of the mismatch otherwise.

jglick:
Suggested some improvements to Javadoc but otherwise seems OK to merge, so going ahead with it.

PR #89, opened 2017-11-29
[RFC] space char in tmp dirs to catch quoting issues

thomasgl-orange:
This is an idea I got after reporting [JENKINS-48264](https://issues.jenkins-ci.org/browse/JENKINS-48264), which is a recent regression in pipeline-maven-plugin with job names (hence workspace paths) containing space characters.  

It's not the first time I come across a quoting issue like that in a plugin, so... why not try to detect them (at least some) sooner, by tweaking JenkinsRule to always put some space characters in workspaces paths?
My initial thought was to make a WorkspaceLocator or something like that, but the patch in this PR is even simpler.

I've verified that it somehow works: it breaks 6 out of 27 tests in the WithMavenStepOnMasterTest class in pipeline-maven-plugin 3.0.3.

I'm not sure how good or stupid this idea is, just wanted to share and get your feedback.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
+++ b/src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
@@ -45,6 +45,14 @@ public class TemporaryDirectoryAllocator {
      */
     private final File base;
 
+    /**
+     * Whether there should be a space character in the allocated temporary directories names.
+     * It forces slaves created from a {@link JenkinsRule} to work inside a hazardous path,
+     * which can help catching shell quoting bugs.<br>
+     * This option is controlled by the <code>jenkins.test.noSpaceInTmpDirs</code> system property.
+     */
+    private final boolean withoutSpace = Boolean.getBoolean("jenkins.test.noSpaceInTmpDirs");
+
     public TemporaryDirectoryAllocator(File base) {
         this.base = base;
     }
@@ -62,7 +70,7 @@ public class TemporaryDirectoryAllocator {
      */
     public synchronized File allocate() throws IOException {
         try {
-            File f = File.createTempFile("jenkins", " test", base);
+            File f = File.createTempFile("jenkins", (withoutSpace ? "test" : " test"), base);
             f.delete();
             f.mkdirs();
             tmpDirectories.add(f);
--- a/src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
+++ b/src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
@@ -62,7 +62,7 @@ public class TemporaryDirectoryAllocator {
      */
     public synchronized File allocate() throws IOException {
         try {
-            File f = File.createTempFile("jenkins", "test", base);
+            File f = File.createTempFile("jenkins", " test", base);
             f.delete();
             f.mkdirs();
             tmpDirectories.add(f);


jglick:
probably a good time to call in @jenkinsci/code-reviewers

MarkEWaite:
If you do something like this, please provide a way to disable it.  Command line git based authentication fails in ugly ways on Windows when there are spaces in the path to the temporary directory.  The temporary directory prefers to be a derivative of the workspace name (to avoid placing credentials into the system temporary directory).  Thus, this change could cause temporary directory names to have spaces in their path.

I believe the git plugin has code that detects spaces in the workspace path.  If spaces are detected in the workspace path, then it switches and uses the system temporary directory instead.  I'd rather not have that variability in the upstream test harness if it can't be easily disabled.

Yes, I accept that command line git should not have authentication problems when the temporary directory includes a space.  I also accept that we should rewrite the command line git authentication to use GIT_SSH_COMMAND for those git installations which have a modern enough implementation of command line git.  Unfortunately, neither of those things is likely to happen soon.

thomasgl-orange:
> If you do something like this, please provide a way to disable it.

@MarkEWaite  OK, I've added a system property to control this behaviour.  If it's good enough like this (without fine-grained control over which test gets spaces or not), then we could also add a corresponding Maven property in Jenkins plugin-pom, which would pass the value to surefire's system properties. This way it would be less a hidden "feature", and plugins which can't work with it would have an easy way to disable it.
I can make a PR for plugin-pom if this approach gets agreed on.

MarkEWaite:
That level of control is great.  Thanks

thomasgl-orange:
I've submitted jenkinsci/plugin-pom#93 to introduce the corresponding Maven property, which would then be passed to the surefire JVM.

src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
```diff
@@ -62,7 +70,7 @@ public TemporaryDirectoryAllocator() {
      */
     public synchronized File allocate() throws IOException {
         try {
-            File f = File.createTempFile("jenkins", "test", base);
+            File f = File.createTempFile("jenkins", (withoutSpace ? "test" : " test"), base);
```
jtnord: Just a nit that starting or ending with space is a bit annoying for listing and copy/paste.
I would prefer something like "a test"

jglick: After #111, I would suggest `jkh` → `j h`.

thomasgl-orange: Thanks @jglick. I've merged master in my branch, and made the change you've suggested (the `"jkh"` prefix becoming `"j h"`).

batmat:
Given there were only approvals, **I'm going to merge this in the next days if nobody objects.** 

I'm only giving it a bit more time because it was approved by others already almost one year ago, so I'm being slightly extra cautious.

thomasgl-orange:
@jglick @batmat 
Just a little status update. So far I've made PR to test JTH 2.45, with this space in tmp dir path change, on two plugins:
- jenkinsci/git-client-plugin#393
- jenkinsci/git-plugin#661
On the later, one test fails because of `Xalan` being in the classpath (depended-on by `jenkins-test-harness-htmlunit`), and messing with file:// URLs:
https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fgit-plugin/detail/PR-661/1/tests
http://svn.apache.org/viewvc/xalan/java/tags/xalan-j_2_7_2/src/org/apache/xalan/transformer/TransformerIdentityImpl.java?view=markup#l273

I think that's an issue which should be addressed in JTH, not sure how (not sure how mandatory is the xalan dependency for htmlunit, etc.).

jglick:
The code in that version of Xalan-J is clearly wrong. The question is

* Does this affect production usages of `ItemGroupMixIn.createProjectFromXML`, or is the Oracle implementation of this JAXP API correct?
* Why is Xalan-J in the test classpath? Can it upgraded, or removed? Or can it be used for explicit API calls while its JAXP registrations are suppressed?

thomasgl-orange:
> Does this affect production usages of `ItemGroupMixIn.createProjectFromXML`, or is the Oracle implementation of this JAXP API correct?

According to [JENKINS-45512](https://issues.jenkins-ci.org/browse/JENKINS-45512?focusedCommentId=312187), at least JobDSL uses this method, and it works just fine in production (with spaces in item names, thus in the file path).  It only fails during unit tests, when `xalan` gets added to the classpath and used as the default xslt provider.  So, no, I assume there is no similar issue with the Oracle default implementation.

> Why is Xalan-J in the test classpath?

`jenkins-test-harness-htmlunit` depends on it, because it is an `htmlunit` compile/runtime dependency.

> Can it upgraded,

Version 2.7.2 is still the latest published version. Not sure it is actively maintained.

> or removed?

`htmlunit` explicitly uses some xalan classes (`org.apache.xml.utils.PrefixResolver` and a few others - and that's still true in latest `htmlunit` version, 2.33), so I would say no, not without modifying `htmlunit`.

> Or can it be used for explicit API calls while its JAXP registrations are suppressed?

I think it is the way to go, yes.  Overriding its `javax.xml.transform.TransformerFactory` registration seems to be enough: jenkinsci/git-plugin@f2da193 ([CI build](https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fgit-plugin/detail/PR-661/4/pipeline) is still pending)
Of course, that's not something which should be done at the plugin level like this, that was just for illustrating the idea. Doing it in `jenkins-test-harness-htmlunit` would make sense I think, or here in JTH. 
One issue with this is that you can't (AFAIK) simply prevent `xalan` from registering itself (not without repackaging its jar). You have to explicitly force the setting back to what is, currently, the JDK 8 default. Which might break at some point in the future (I've not checked if it had changed in JDK 11).

jglick:
> I assume there is no similar issue with the Oracle default implementation.

A fork of Xalan, IIRC. A lot of fixes were not upstreamed.

> not without modifying `htmlunit`

Indeed it seems to bake `org.apache.xml.utils.PrefixResolver` into its own APIs—a bad idea, since it cannot be removed without risking breakage. Xalan-J itself [appears dead](https://github.com/apache/xalan-j/commits/trunk).

> you can't simply prevent `xalan` from registering itself without repackaging its jar

Sounds like repackaging the Xalan JAR is the simplest way to go. We already [shade dependencies](https://github.com/jenkinsci/jenkins-test-harness-htmlunit/blob/b58312305232c2dc27d333b0c0bf0e3cded424a7/pom.xml#L108-L109) so I suppose we could do the same for Xalan-J (and Xerces while we are at it—this also has JAXP registrations).

thomasgl-orange:
@jglick

> Sounds like repackaging the Xalan JAR is the simplest way to go. We already shade dependencies so I suppose we could do the same for Xalan-J (and Xerces while we are at it—this also has JAXP registrations).

OK, sounds good, I will submit a PR for JTH-htmlunit with these changes, hopefully tomorrow. I'm first testing htmlunit itself (running its own extensive tests suite) with a modified Xalan-J (the original jar minus its JAXP registration files), just to check whether NOT having Xalan-J registered breaks anything (because if it does, well... then it would mean this change has potential for randomly breaking some valid plugins unit tests).

jvz:
There's an [effort to reboot Xalan](https://lists.apache.org/thread.html/05ac466caf2dfcac365f8a0b1ef6e4f4593ff6bce43ada59131625c4@%3Cdev.xalan.apache.org%3E), but that's about it right now.

PR #91, opened 2017-12-09
Add utilities to RestartableJenkinsRule to simulate dirty shutdowns

svanoort:
This can be used to verify that data is appropriately persisted and masters can start cleanly even when normal shutdown-time processes don't get a chance to run. 

Note that it isn't perfect -- slightly-delayed periodic or async processes may still get a chance to run, since the JENKINS_HOME isn't atomically/instantly snapshotted (instead it is copied over a piece at a time).  However, it has been useful to identify persistence issues. 

Deployed as:
* 2.33-20171220.165503-3

@reviewbybees

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
@@ -1,6 +1,7 @@
 package org.jvnet.hudson.test;
 
 import groovy.lang.Closure;
+import hudson.FilePath;
 import org.junit.rules.MethodRule;
 import org.junit.rules.TemporaryFolder;
 import org.junit.runner.Description;
@@ -9,14 +10,8 @@ import org.junit.runners.model.Statement;
 
 import java.io.File;
 import java.io.IOException;
-import java.nio.file.FileVisitResult;
-import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
-import java.nio.file.Path;
-import java.nio.file.SimpleFileVisitor;
-import java.nio.file.attribute.BasicFileAttributes;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.Callable;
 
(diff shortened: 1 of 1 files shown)


svanoort:
@jglick I hope you don't mind if I assign you as a reviewer since unless my memory fails you'd expressed interest in seeing this contributed upstream.   @batmat and @oleg-nenashev may find this of interest and value for their current work.

src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
```diff
@@ -76,6 +83,63 @@ public void evaluate() throws Throwable {
...
+        @Override
+        public FileVisitResult visitFile(final Path file,
+                                         final BasicFileAttributes attrs) throws IOException {
+            if (!Files.isSymbolicLink(file)) {
+                // Needed because Jenkins includes invalid lastSuccessful symlinks and otherwise we get a NoSuchFileException
+                Files.copy(file,
+                        targetPath.resolve(sourcePath.relativize(file)));
+            } else if (Files.isSymbolicLink(file) && Files.exists(Files.readSymbolicLink(file))) {
+                Files.copy(file,
+                        targetPath.resolve(sourcePath.relativize(file)));
```
jglick: Unless I am missing something, these two branches do exactly the same thing.

src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
```diff
@@ -76,6 +83,63 @@ public void evaluate() throws Throwable {
...
+     * @throws IOException
+     */
+     void simulateAbruptShutdown() throws IOException {
+        File homeDir = this.home;
+        TemporaryFolder temp = new TemporaryFolder();
+        temp.create();
+        File newHome = temp.newFolder();
+
+        // Copy efficiently
+        Files.walkFileTree(homeDir.toPath(), Collections.EMPTY_SET, 99, new CopyFileVisitor(newHome.toPath()));
```
jglick: Why not just use `FilePath.copyRecursiveTo`, and if that is not as fast as you think it could be, optimize it? Surely we do not need blazing performance for a test utility anyway.

svanoort:
@jglick I'm already using the logic in https://github.com/jenkinsci/workflow-cps-plugin/pull/188/files#diff-e489d687952c586b5f1e0c8d589c7050R491 -- this is just making it usable for others (I'd only need to adapt the test-harness dependency and update code to use this instead of the copied logic).

src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
svanoort: @jglick It's subtle -- think about what happens if you have a symbolic link to something nonexistent. ;)  This is solving a specific bug that occurs with Jenkins.

src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
batmat: +1, also would be indeed an opportunity to actually have a bigger perf impact, if need be. Since `copyRecursiveTo` is used in Pipeline too FWICT by the `libraryAdder`.

svanoort:
@reviewbybees done

PR #92, opened 2018-01-10
[JENKINS-48885] - Fix getPlugin() in JenkinsRule

oleg-nenashev:
WiP for now, just to not forget about it.

https://issues.jenkins-ci.org/browse/JENKINS-48885

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,8 +79,8 @@ THE SOFTWARE.
     <dependency>
       <groupId>org.jenkins-ci.main</groupId>
       <artifactId>jenkins-war</artifactId>
-      <!--to have access to User.getById-->
-      <version>1.651.2</version>
+      <!--to run with plugin unbundling by default-->
+      <version>2.7.3</version>
       <type>executable-war</type>
       <exclusions>
         <exclusion>
--- a/src/main/java/org/jvnet/hudson/test/recipes/WithPlugin.java
+++ b/src/main/java/org/jvnet/hudson/test/recipes/WithPlugin.java
@@ -30,6 +30,7 @@ import org.apache.commons.io.FileUtils;
 
 
 import java.io.File;
+import java.io.IOException;
 import java.lang.annotation.Documented;
 import static java.lang.annotation.ElementType.METHOD;
 import java.lang.annotation.Retention;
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -79,8 +79,8 @@ THE SOFTWARE.
     <dependency>
       <groupId>org.jenkins-ci.main</groupId>
       <artifactId>jenkins-war</artifactId>
-      <!--to have access to User.getById-->
-      <version>1.651.2</version>
+      <!--to run with plugin unbundling by default-->
+      <version>2.7.3</version>
```
jglick: Then you would need to do some cleanup. `TestPluginManager` has distinct modes for 1.x vs. 2.x.

src/main/java/org/jvnet/hudson/test/recipes/WithPlugin.java
```diff
@@ -89,6 +90,9 @@ public void setup(JenkinsRule jenkinsRule, WithPlugin recipe) throws Exception {
         public void decorateHome(JenkinsRule jenkinsRule, File home) throws Exception {
             for (String plugin : a.value()) {
                 URL res = getClass().getClassLoader().getResource("plugins/" + plugin);
+                if (res == null) {
+                    throw new IOException("Cannot find resource for plugin " + plugin);
```
jglick: This only applies to tests actually annotated with `@WithPlugin`. No effect on tests assuming plugin existence due to a `test`-scoped dependency.

oleg-nenashev: Yes. It is not really related to the defect. Just better diagnostics for another issue I hit. Can detach it to a separate PR if you prefer

MRamonLeon: I think I've been impacted by this issue too doing some tests in support-core plugin. Here the folder _plugins_ doesn't exist: https://github.com/jenkinsci/support-core-plugin/blob/54fde93fa28ac1c69adc0a2a6c4962ca2c629b3c/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java#L837

oleg-nenashev: IIRC the plugin registry is just not initialized in the Test plugin manager

pom.xml
jglick: …which I guess did not get cleaned up in #135, so maybe do that here?

oleg-nenashev: I believe we will wipe 1.x entirely after https://groups.google.com/forum/#!topic/jenkinsci-dev/SdMOMKs7XIQ

oleg-nenashev:
I plan to revisit it using the approach suggested in https://github.com/jenkinsci/jenkinsfile-runner/issues/389 and https://github.com/jenkinsci/jenkinsfile-runner/issues/390.

PR #95, opened 2018-02-10
[JENKINS-49491] Adapting JTH to a recent version of HtmlUnit

recena:
[JENKINS-49491](https://issues.jenkins-ci.org/browse/JENKINS-49491)

Downstream of https://github.com/jenkinsci/jenkins-test-harness-htmlunit/pull/1

@jenkinsci/code-reviewers

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
+++ b/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
@@ -1018,7 +1018,7 @@ public abstract class HudsonTestCase extends TestCase implements RootAction {
      * Plain {@link HtmlForm#submit(com.gargoylesoftware.htmlunit.html.SubmittableElement)} doesn't work correctly due to the use of YUI in Hudson.
      */
     public HtmlPage submit(HtmlForm form) throws Exception {
-        return (HtmlPage) HtmlFormUtil.submit(form, last(form.getHtmlElementsByTagName("button")));
+        return (HtmlPage) HtmlFormUtil.submit(form, last(form.getElementsByTagName("button")));
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
```diff
@@ -1397,29 +1397,6 @@ public boolean processSynchron(HtmlPage page, WebRequest settings, boolean async
                 }
             });
 
-            setCssErrorHandler(new ErrorHandler() {
```
recena: I need to replace this one.

recena:
Why do not we remove `HudsonTestCase.java`?

MarkEWaite:
@recena I thought that removing the `HudsonTestCase` class would break those plugins which have not yet migrated from HudsonTestCase to JenkinsRule.  The git plugin only made that transition in 2016, so I assume there are other plugins which still depend on HudsonTestCase.

recena:
@MarkEWaite For those cases, the maintainer has to make a decision. If he wants to update `JTH` then needs to review his tests. Or do you mean about OSS ATH?

jglick:
Yes there are a _lot_ of tests out there still extending `HudsonTestCase`. While we would certainly like to continue to update them, it is labor-intensive, so anything which really breaks it had better have a very solid justification. Merely updating HtmlUnit is not enough reason.

recena:
The motivation is not relared with upgrading HtmlUnit but with keeping the things clearer.

rtyler:
FWIW, without updating HTMLUnit, it's not possible (as far as I can tell) to use ECMAScript 6, or anything moderately considered "contemporary" JavaScript in Jenkins Core.

IMHO that's a strong motivation.

recena:
@rtyler I could continue working here whether it is still interesting for the community.

uhafner:
Fixed with https://github.com/jenkinsci/jenkins-test-harness/pull/103

PR #96, opened 2018-03-29
[JENKINS-50476] offer a way to assert that Jenkins won't start

batmat:
[JENKINS-50476](https://issues.jenkins-ci.org/browse/JENKINS-50476)

Used in https://github.com/jenkinsci/jenkins/pull/3364

@reviewbybees

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
@@ -2,6 +2,7 @@ package org.jvnet.hudson.test;
 
 import groovy.lang.Closure;
 import hudson.FilePath;
+import org.junit.Assert;
 import org.junit.rules.MethodRule;
 import org.junit.rules.TemporaryFolder;
 import org.junit.runner.Description;
@@ -12,7 +13,9 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.file.NoSuchFileException;
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.Callable;
 
 /**
@@ -34,7 +37,11 @@ import java.util.concurrent.Callable;
 public class RestartableJenkinsRule implements MethodRule {
     public JenkinsRule j;
     private Description description;
-    private final List<Statement> steps = new ArrayList<Statement>();
+
+    /**
+     * List of {@link Statement}. For each one, the boolean value says if Jenkins is expected to start or not.
+     */
+    private final Map<Statement, Boolean> steps = new LinkedHashMap<>();
 
     private TemporaryFolder tmp = new TemporaryFolder();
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
```diff
@@ -188,9 +208,20 @@ public File allocate() throws Exception {
...
+            try {
+                j.apply(step, description).evaluate();
+                if (!entry.getValue()) {
+                    Assert.fail("The current JenkinsRule should have failed to start Jenkins.");
+                }
+            } catch (Exception e) {
+                if(entry.getValue()) {
+                    throw e;
+                }
+                // Failure ignored as requested
```
svanoort: Might be worth logging something to indicate we saw expected failure, so we know rule ran okay.

src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
```diff
@@ -188,9 +208,20 @@ public File allocate() throws Exception {
...
         // run each step inside its own JenkinsRule
-        for (Statement step : steps) {
+        for (Map.Entry<Statement, Boolean> entry : steps.entrySet()) {
+            Statement step = entry.getKey();
             j = createJenkinsRule(description).with(loader);
-            j.apply(step,description).evaluate();
+            try {
+                j.apply(step, description).evaluate();
+                if (!entry.getValue()) {
+                    Assert.fail("The current JenkinsRule should have failed to start Jenkins.");
```
svanoort: Super-🐜 - I like to phrase this as "X happened but we saw Y" because that makes it clearer.

src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
```diff
@@ -34,7 +37,11 @@
 public class RestartableJenkinsRule implements MethodRule {
     public JenkinsRule j;
     private Description description;
-    private final List<Statement> steps = new ArrayList<Statement>();
+
+    /**
+     * List of {@link Statement}. For each one, the boolean value says if Jenkins is expected to start or not.
+     */
+    private final Map<Statement, Boolean> steps = new LinkedHashMap<>();
```
svanoort: 🐜 I would declare the type here as *explicitly* LinkedHashmap since no other type could be legally used since we need ordering preserved.   

What we're doing here is a *bit* dodgy (using a LinkedHashmap to simulate a List of Tuple types, with no intent to really use mapping)... but it's also much less verbose than the "proper" alternative and doesn't require adding Apache Commons Lang [with a modern version to get tuple types](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/tuple/Pair.html).

batmat:
@svanoort thanks a lot for the review. I will address your comments in a followup PR. I'm merging and releasing to make the associated core PR ready (instead of using a JTH SNAPSHOT).

PR #97, opened 2018-04-05
[JENKINS-50590] do not remove all request parameters when adding a crumb

jtnord:
If you had a WebRequest with existing Parameters specified then adding a
crumb would remove all of those headers.  THis is somewhat suprising.

I thought about making getCrumbHeaderNVP() public but went with the
option that should anyway cause the least suprise.  The crumb is now
added to the list of parameters rather than replacing them.  It is added
first such that if there where existing crumb fields the newly added one
should be first.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -2300,8 +2300,13 @@ public class JenkinsRule implements TestRule, MethodRule, RootAction {
          * Use {@link #createCrumbedUrl} instead if you intend to call {@link WebRequest#setRequestBody}, typical of a POST request.
          */
         public WebRequest addCrumb(WebRequest req) {
-            NameValuePair crumb = getCrumbHeaderNVP();
-            req.setRequestParameters(Arrays.asList(crumb));
+            ArrayList<NameValuePair> params = new ArrayList<>();
+            params.add(getCrumbHeaderNVP());
+            List<NameValuePair> oldParams = req.getRequestParameters();
+            if (oldParams != null && !oldParams.isEmpty()) {
+                params.addAll(oldParams);
+            }
+            req.setRequestParameters(params);
             return req;
         }
 


src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -2300,8 +2300,13 @@ public String getContextPath() throws IOException {
          * Use {@link #createCrumbedUrl} instead if you intend to call {@link WebRequest#setRequestBody}, typical of a POST request.
```
jglick: You should also fix `createCrumbedUrl` to look for `?` in `relativePath` and switch to `&`.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -2300,8 +2300,13 @@ public String getContextPath() throws IOException {
...
          */
         public WebRequest addCrumb(WebRequest req) {
-            NameValuePair crumb = getCrumbHeaderNVP();
-            req.setRequestParameters(Arrays.asList(crumb));
+            ArrayList<NameValuePair> params = new ArrayList<>();
+            params.add(getCrumbHeaderNVP());
+            List<NameValuePair> oldParams = req.getRequestParameters();
+            if (oldParams != null && !oldParams.isEmpty()) {
```
jglick: Can `getRequestParameters` actually be null? I doubt it. And if it is empty, the `addAll` is a no-op anyway; so probably the `if` statement could be elided.

jtnord:
> BTW this is much less important after jenkinsci/jenkins#3129.

@jglick I am using 2.107 which has  jenkinsci/jenkins#3129 and I needed this otherwise things where blocked.  (using in a `JenkinsRule` where anonymous access is allowed - so I guess no TOKEN is used)

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
jtnord: ok, will take a look

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
jtnord: they can be if you call `setRequestParameters` with `null` (its not checked - but they are initialised to an empty collection)

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
jtnord: @jglick addressed [here](https://github.com/jenkinsci/jenkins-test-harness/pull/97/files#diff-1d6816bd6f9430280c5d8b7853286b1dR2320)

PR #99, opened 2018-04-05
[JENKINS-50598] - Add support of running JTH with a custom WAR

oleg-nenashev:
I am experimenting with running PCT with WARs generated by Custom WAR Packager. I would like to be able to run JTH and PCT tests against WARs instead of the dependency-based mechanism. It would allow to support things like auto-configuration of WARs which is offered by WAR packager.

- [x] - Custom WAR can be passed via system property
- [x] - It is possible to prevent replacing of bundled plugins by older dependencies (opt-in, system property)
- [ ] - Some documentation

https://issues.jenkins-ci.org/browse/JENKINS-50598

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/TestPluginManager.java
+++ b/src/main/java/org/jvnet/hudson/test/TestPluginManager.java
@@ -32,17 +32,26 @@ import hudson.Util;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.lang.reflect.Method;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.Collection;
-import java.util.HashSet;
-import java.util.Set;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.jar.JarEntry;
+import java.util.jar.Manifest;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+
+import hudson.util.VersionNumber;
 import org.apache.commons.io.FileUtils;
 import org.junit.Assert;
+
+import javax.annotation.CheckForNull;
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
@reviewbybees @olivergondza

src/main/java/org/jvnet/hudson/test/TestPluginManager.java
```diff
@@ -57,6 +66,11 @@
  * @see HudsonTestCase#useLocalPluginManager
  */
 public class TestPluginManager extends PluginManager {
+
+    @CheckForNull
+    private static final boolean DO_NOT_OVERRIDE_BUNDLED_PLUGINS_BY_OLDER_TEST_DEPS =
+            Boolean.getBoolean("jth.pluginManager.doNotOverrideBundledPluginsByOlderTestDeps");
```
jglick: Maybe -0 on this: I feel it makes the semantics of functional tests a bit harder to understand, and for no very compelling reason: you can already simply ensure that the WAR under test contains sufficiently new versions of the dependencies.

raul-arabaolaza:
@jglick I believe i have one valid use case, integration testing of shared libraries and Jenkinsfiles.

Now that CWP supports Configuration as code you can easily define a war file that ressembles a real environment, and with this you can use Jenkinsrule to perform the testing of your shared libraries using JTH with minimum hassle, you can even use the docker fixtures to set up auxiliary services like git repos and similar. 

WDYT?

jglick:
> you can easily define a war file that ressembles a real environment, and with this you can use Jenkinsrule to perform the testing of your shared libraries using JTH

Not following. If you are using CWP, why would you not have the versions of plugins in CWP that you wanted?

jglick:
@oleg-nenashev do you believe this is still still valid?

oleg-nenashev:
Might be, but PCT already has a number of solutions applied for this issue.
Closing since I do not plan to work on this PR ATM

PR #100, opened 2018-04-05
[JENKINS-50598] - Add support of passing custom WAR file in WAR exploder using system property

oleg-nenashev:
It is a detached commit from #99. @jglick has a concern about the dependency management patch, so I decided to park that part for a while and to deliver this change. Downstream change in PCT - WiP

@reviewbybees

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/WarExploder.java
+++ b/src/main/java/org/jvnet/hudson/test/WarExploder.java
@@ -33,6 +33,8 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import jenkins.model.Jenkins;
 
+import javax.annotation.CheckForNull;
+
 /**
  * Ensures that <tt>jenkins.war</tt> is exploded.
  *
@@ -45,6 +47,10 @@ import jenkins.model.Jenkins;
 public final class WarExploder {
 
     private static final Logger LOGGER = Logger.getLogger(WarExploder.class.getName());
+
+    public static final String JENKINS_WAR_PATH_PROPERTY_NAME = "jth.jenkins-war.path";
+    @CheckForNull
+    private static final String JENKINS_WAR_PATH = System.getProperty(JENKINS_WAR_PATH_PROPERTY_NAME);
 
     public static File getExplodedDir() throws Exception {
         // rethrow an exception every time someone tries to do this, so that when explode()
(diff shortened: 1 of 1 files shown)


jglick:
> Downstream change in PCT - WiP

Nothing filed yet? Best to keep on hold until that is demonstrated.

src/main/java/org/jvnet/hudson/test/WarExploder.java
```diff
@@ -46,6 +48,10 @@
 
     private static final Logger LOGGER = Logger.getLogger(WarExploder.class.getName());
 
+    public static final String JENKINS_WAR_PATH_PROPERTY_NAME = "jth.jenkins-war.path";
+    @CheckForNull
+    private static final String JENKINS_WAR_PATH = System.getProperty(JENKINS_WAR_PATH_PROPERTY_NAME);
```
jglick: or just `System.getenv("JENKINS_WAR")`?

src/main/java/org/jvnet/hudson/test/WarExploder.java
```diff
@@ -84,23 +90,34 @@ private static File explode() throws Exception {
             }
         }
 
-        // locate jenkins.war
-        File war;
-        URL winstone = WarExploder.class.getResource("/winstone.jar");
```
jglick: A lot of [reindentation](https://github.com/jenkinsci/jenkins-test-harness/pull/100/files?w=1); would be clearer if you could use an `if`-`else if`-`else` series with the same indentation.

src/main/java/org/jvnet/hudson/test/WarExploder.java
```diff
@@ -84,23 +90,34 @@ private static File explode() throws Exception {
...
-        if (winstone != null) {
-            war = Which.jarFile(Class.forName("executable.Executable"));
+        final File war;
+        if (JENKINS_WAR_PATH != null) {
+            war = new File(JENKINS_WAR_PATH).getAbsoluteFile();
+            LOGGER.log(Level.INFO, "Using a predefined WAR file {0} define by the -{1} system property",
```
jglick: Why the `-`?

src/main/java/org/jvnet/hudson/test/WarExploder.java
oleg-nenashev: It can be added later if needed.
The biggest problem with that is that the same env name is used by packaging logic in Jenkins: https://github.com/search?q=org%3Ajenkinsci+JENKINS_WAR&type=Code (e.g. Docker packages). If one runs plugin build on a Jenkins with such system environment variables, JTH may pick it and run tests against a wrong WAR. So I prefer explicit definition

PR #101, opened 2018-04-05
"Dirty" restart enhancements

svanoort:
- Switch back to using native file copier utility so it better handles Symlinks and missing files, i.e. tmpFiles that are created by atomic operations and may be deleted while the copy operation is running
- Added logging for abrupt shutdown processes so you can see what happens *while* snapshot of JENKINS_HOME is being taken... and what happen after it (activity that can be ignored).
- Swallows IOExceptions during abrupt shutdown since they can be thrown in spurious places

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
@@ -1,7 +1,6 @@
 package org.jvnet.hudson.test;
 
 import groovy.lang.Closure;
-import hudson.FilePath;
 import org.junit.Assert;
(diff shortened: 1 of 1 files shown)


batmat:
As there are no tests here, do you have at least an example somewhere, ideally CI, where you used this successfully?

svanoort:
@batmat  Yeah it's used heavily in workflow-cps -- I'm using a SNAPSHOT dep here -- https://github.com/jenkinsci/workflow-cps-plugin/pull/216

Most of the issues here couldn't really be tested for because they only crop up in the 'fuzzing' tests which are essentially hammering on the persistence to find race conditions / timing or order of operations issues where durability guarantees break.

PR #103, opened 2018-05-03
Upgrade to HTMLUnit 2.31

uhafner:
Old HTML version does not support modern JS Libraries like jQuery 3.3.
- Removed old dependency to jenkins-test-harness-htmlunit.
- Fixed calls to deprecated, renamed, or removed methods.


pom.xml
```diff
@@ -115,19 +115,9 @@ THE SOFTWARE.
...
-          <groupId>org.eclipse.jetty</groupId>
-          <artifactId>jetty-io</artifactId>
-        </exclusion>
-        <exclusion>
-          <groupId>org.eclipse.jetty</groupId>
-          <artifactId>jetty-util</artifactId>
-        </exclusion>
-      </exclusions>
+      <groupId>net.sourceforge.htmlunit</groupId>
+      <artifactId>htmlunit</artifactId>
```
jglick: No, I think we rather wanted the `jenkins-test-harness-htmlunit` to be updated to pull in the upstream 2.30 version, then cut a `2.30-1` release of that which would continue to do the shading.

src/main/java/org/jvnet/hudson/test/HudsonPageCreator.java
```diff
@@ -23,40 +23,35 @@
...
-import org.apache.commons.lang3.StringUtils;
-
-import java.io.IOException;
-import java.io.InputStream;
-import java.util.Locale;
 
 /**
  * {@link PageCreator} that understands JNLP file.
- * 
+ *
```
jglick: Can you revert inessential changes such as whitespace and reformatting please? (You should be able to configure your IDE/editor to not modify lines of text unless and until a significant change is made on that line.)

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -24,88 +24,10 @@
...
-import hudson.EnvVars;
-import hudson.Extension;
-import hudson.ExtensionList;
-import hudson.Functions;
-import hudson.Launcher;
-import hudson.Main;
-import hudson.PluginManager;
-import hudson.Util;
-import hudson.WebAppMain;
-import hudson.model.AbstractBuild;
```
jglick: Could this diff be reduced? Surely you are not making many actual changes to imports.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -1981,7 +1983,7 @@ public WebClient createWebClient() {
         public WebClient() {
             // default is IE6, but this causes 'n.doScroll('left')' to fail in event-debug.js:1907 as HtmlUnit doesn't implement such a method,
```
jglick: Is this comment still accurate?

src/main/java/org/jvnet/hudson/test/HudsonPageCreator.java
uhafner: I can do yes. (If the rest of the changes are committed to reduce the overhead.)

Just a side note: This requires that I set up a new IDE configuration just for every Jenkins sub-project. Almost every other project I am contributing to defines a coding style that is also part of the project (i.e. in version control). Your approach by using no coding style at all makes contributing to Jenkins quite a burden. Normally, the IDE would be smart enough to handle those changes as you like them. In your case I need to manually revert such changes a smart IDE automatically takes care of.

jglick: You do not need to manually revert changes if you do not make the changes in the first place, which you should be able to configure your IDE to do.

uhafner: Haha.

uhafner:
Snapshot available as 2.39-SNAPSHOT/jenkins-test-harness-2.39-20180609.202456-3.jar

uhafner:
Snapshot available as 2.40-SNAPSHOT/jenkins-test-harness-2.40-20180609.204658-1.jar.

uhafner:
Ping? Is there anything missing?

uhafner:
Yes, thanks!

I thing you then can close https://github.com/jenkinsci/jenkins-test-harness/pull/95.

PR #105, opened 2018-06-04
Prevent agent processes from stealing focus on OS X

dwnusbaum:
Any test that launches a new java process without passing `-Djava.awt.headless=true` causes focus to switch to the new process, which is mildly infuriating when I want to work on something while running tests in the background. A similar issue was fixed in jenkinsci/plugin-pom#22 to prevent `JenkinsRule` tests from stealing focus by default.

The focus-stealing feature on OS X can be disabled entirely, but it is normally useful and I do not have this issue with any other background processes on my machine, so I'd prefer to fix it here.

I am happy to create a Jira ticket or a screen recording to demonstrate the issue if desired.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -183,7 +183,7 @@ THE SOFTWARE.
         <artifactId>maven-surefire-plugin</artifactId>
         <!-- version specified in grandparent pom -->
         <configuration>
-          <argLine>-Dfile.encoding=UTF-8 -Xmx256m</argLine>
+          <argLine>-Dfile.encoding=UTF-8 -Xmx256m -Djava.awt.headless=true</argLine>
           <systemPropertyVariables>
               <!-- use AntClassLoader that supports predictable file handle release -->
               <hudson.ClassicPluginStrategy.useAntClassLoader>true</hudson.ClassicPluginStrategy.useAntClassLoader>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -183,7 +183,7 @@ THE SOFTWARE.
         <artifactId>maven-surefire-plugin</artifactId>
         <!-- version specified in grandparent pom -->
         <configuration>
-          <argLine>-Dfile.encoding=UTF-8 -Xmx256m</argLine>
+          <argLine>-Dfile.encoding=UTF-8 -Xmx256m -Djava.awt.headless=true</argLine>
```
dwnusbaum: Only relevant when running the test suite of the JTH itself. Plugins get this property from the [plugin pom](https://github.com/jenkinsci/plugin-pom/blob/fe93cf79e83f1a03e4f149b554ceaadd6f8072be/pom.xml#L50).

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -918,9 +918,10 @@ public PretendSlave createPretendSlave(FakeLauncher faker) throws Exception {
     public ComputerLauncher createComputerLauncher(EnvVars env) throws URISyntaxException, IOException {
         int sz = jenkins.getNodes().size();
         return new SimpleCommandLauncher(
-                String.format("\"%s/bin/java\" %s -jar \"%s\"",
+                String.format("\"%s/bin/java\" %s %s -jar \"%s\"",
                         System.getProperty("java.home"),
                         SLAVE_DEBUG_PORT>0 ? " -Xdebug -Xrunjdwp:transport=dt_socket,server=y,address="+(SLAVE_DEBUG_PORT+sz): "",
+                        "-Djava.awt.headless=true",
```
jglick: A constant, so could just inline, but OK.

PR #108, opened 2018-08-20
Avoid serializing Permission.ID_COMPARATOR

jglick:
Follow-up to https://github.com/jenkinsci/remoting/pull/282 from running `com.cloudbees.plugins.credentials.SystemCredentialsProviderTest.given_globalScopeCredential_when_builtAsUserWithoutUseItem_then_credentialNotFound` against a new core. Saw

```
WARNING	o.j.r.u.AnonymousClassWarnings#warn: Attempt to (de-)serialize anonymous class hudson.security.Permission$1 in file:…/jenkins-core-2.138.jar; see: https://jenkins.io/redirect/serialization-of-anonymous-classes/
```

which referred to `Permission.ID_COMPARATOR`, that indeed was not designed to be serialized (in this case by XStream).

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/MockAuthorizationStrategy.java
+++ b/src/main/java/org/jvnet/hudson/test/MockAuthorizationStrategy.java
@@ -41,6 +41,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 import jenkins.model.Jenkins;
 import org.acegisecurity.acls.sid.Sid;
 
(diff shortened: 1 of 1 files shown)


jglick:
> Does something else need to be done for general usage of the Permission class?

Not that I know of; plugins just should not be trying to serialize `Permission`. You can see some examples of serializing the IDs, such as [here](https://github.com/jenkinsci/matrix-auth-plugin/blob/fabddd5ed453edbdc3aa02bc3a026cf609e770c3/src/main/java/org/jenkinsci/plugins/matrixauth/AbstractAuthorizationContainerConverter.java#L66). (Why does it not just retain a `Map<String, Set<String>> grantedPermissions` to begin with and ditch the `Converter`? Settings compatibility I think.)

PR #109, opened 2018-09-11
[JENKINS-53511] Improve discovery and readability of WebClient

Wadeck:
- upstream of https://github.com/jenkinsci/jenkins/pull/3618
- add methods to directly change the throwing exception behavior (and redirect behavior)
- also correct behavior of assertFail that is not taking into account the "throwExceptionOnFailingStatusCode" option

See [JENKINS-53511](https://issues.jenkins-ci.org/browse/JENKINS-53511).

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -31,6 +31,7 @@ import com.gargoylesoftware.htmlunit.ElementNotFoundException;
 import com.gargoylesoftware.htmlunit.FailingHttpStatusCodeException;
 import com.gargoylesoftware.htmlunit.HttpMethod;
 import com.gargoylesoftware.htmlunit.Page;
+import com.gargoylesoftware.htmlunit.WebClientOptions;
 import com.gargoylesoftware.htmlunit.WebClientUtil;
 import com.gargoylesoftware.htmlunit.WebRequest;
 import com.gargoylesoftware.htmlunit.WebResponse;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -2072,10 +2072,71 @@ public boolean isJavaScriptEnabled() {
...
+         * Changes the behavior of this webclient when a script error occurs.
+         * Fluent method to ease discovery of feature + improve readability
+         * @see com.gargoylesoftware.htmlunit.WebClientOptions#setThrowExceptionOnFailingStatusCode(boolean)
+         */
+        public WebClient withThrowExceptionOnFailingStatusCode(boolean enabled) {
+            setThrowExceptionOnFailingStatusCode(enabled);
+            return this;
+        }
+
+        public boolean isRedirectEnabled() {
```
jeffret-b: I don't understand why some of these have JavaDoc and not others.

oleg-nenashev:
Agree with @jeffret-b , also bonus points for `@since TODO` for new methods

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
Wadeck: Simply because currently they are far less popular in the project (only one call of `isThrowExceptionOnFailingStatusCode`) compared to the one that I write JavaDoc for.

Now every added methods have their related JavaDoc, and also the `xxxJavaScriptEnabled` as they are similar.

PR #110, opened 2018-09-27
[JENKINS-53823] - Make DummyCloudImpl compatible with JDK11 + facelift

oleg-nenashev:
- [x] - JTH can now be built with JDK11
- [x] - Remove dependency on javax.xml.ws.Provider which was removed from Java 11

https://issues.jenkins-ci.org/browse/JENKINS-53823

@jenkinsci/sig-platform

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@ THE SOFTWARE.
   <parent>
     <groupId>org.jenkins-ci</groupId>
     <artifactId>jenkins</artifactId>
-    <version>1.37</version>
+    <version>1.49-20180926.104803-1</version>
   </parent>
 
   <groupId>org.jenkins-ci.main</groupId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -48,10 +48,13 @@ THE SOFTWARE.
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <jetty.version>9.4.5.v20170502</jetty.version>
-    <maven-surefire-plugin.version>2.20</maven-surefire-plugin.version>
-    <maven-compiler-plugin.version>3.6.1</maven-compiler-plugin.version>
     <java.level>8</java.level>
     <concurrency>1</concurrency> <!-- may use e.g. 2C for 2 × (number of cores) -->
+    <!--TODO: fix upper bounds -->
```
oleg-nenashev: Didn't touch it

src/main/java/hudson/slaves/DummyCloudImpl.java
```diff
@@ -123,15 +121,8 @@ public Node call() throws Exception {
...
-            Provider<NodeProperty> nodePropertyProvider = new Provider<NodeProperty>() {
-                @Override
-                public NodeProperty invoke(NodeProperty request) {
-                    request.setNode(slave);
-                    return request;
-                }
-            };
             for (NodeProperty nodeProperty : nodeProperties) {
-                slave.getNodeProperties().add(nodePropertyProvider.invoke(nodeProperty));
+                slave.getNodeProperties().add(updateWithNode(nodeProperty, slave));
```
dwnusbaum: nit: Unless I'm missing something it looks like you could just inline the method instead:
    
    nodeProperty.setNode(slave);
    slave.getNodeProperties().add(nodeProperty);

oleg-nenashev: Yeah :( 
Whatever, I doubt we will touch this class till JDK 15 or so

dwnusbaum: Yep definitely not important, just figured I'd mention it :)

jglick: I cannot figure out what the original code here was doing. Introduced in https://github.com/jenkinsci/jenkins/pull/1815 FTR.

PR #111, opened 2018-10-01
Shortening paths

jglick:
https://github.com/jenkinsci/plugin-pom/pull/122 picks a temp dir root inside the project checkout, so we prefer to keep the relative paths short and sweet.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -121,6 +121,12 @@ THE SOFTWARE.
       <groupId>${project.groupId}</groupId>
       <artifactId>jenkins-test-harness-htmlunit</artifactId>
       <version>2.31-1</version>
+      <exclusions>
+        <exclusion>
+          <groupId>commons-io</groupId>
+          <artifactId>commons-io</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.jvnet.hudson</groupId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -121,6 +121,12 @@ THE SOFTWARE.
       <groupId>${project.groupId}</groupId>
       <artifactId>jenkins-test-harness-htmlunit</artifactId>
       <version>2.31-1</version>
+      <exclusions>
+        <exclusion>
+          <groupId>commons-io</groupId>
+          <artifactId>commons-io</artifactId>
```
jglick: Fixes a problem in #103.

uhafner: When running `mvn dependency:tree` I don't see any other reference to commons-io in jenkins-test-harness? In my plug-ins I'm using commons-io as dependency so I'm not affected but what happens to other plug-ins, does html-unit work without commons-io? (Or is Jenkins itself is using commons-io?)

jglick: `jenkins-core` itself uses `commons-io`. Without this `exclusion`, https://github.com/jenkinsci/plugin-pom/pull/125 does not build, because `requireUpperBoundsDeps` (rightly) complains that the version inherited from `jenkins-core` is older than the version required by `jenkins-test-harness-htmlunit`. Since `JenkinsRule` does not support class loader tricks to load test code in a separate namespace, we have to choose which to run, and it is better to pick the version used by Jenkins core at runtime.

PR #118, opened 2019-01-02
Code Improvements

darxriggs:
These are commits to improve the current code.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
+++ b/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
@@ -1815,11 +1815,6 @@ public abstract class HudsonTestCase extends TestCase implements RootAction {
             public BuildWrapper newInstance(StaplerRequest req, JSONObject formData) {
                 throw new UnsupportedOperationException();
             }
-
-            @Override // TODO 1.635+ delete
-            public String getDisplayName() {
-                return "TestBuildWrapper";
-            }
         }
     }
 }
--- a/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
+++ b/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
@@ -1005,8 +1005,8 @@ public abstract class HudsonTestCase extends TestCase implements RootAction {
      * a cancellation.
      */
     private List<String> listProperties(String properties) {
-        List<String> props = new ArrayList<String>(Arrays.asList(properties.split(",")));
-        for (String p : props.toArray(new String[props.size()])) {
+        List<String> props = new CopyOnWriteArrayList<>(properties.split(","));
+        for (String p : props) {
             if (p.startsWith("-")) {
                 props.remove(p);
                 props.remove(p.substring(1));
--- a/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
+++ b/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
@@ -680,7 +680,7 @@ public abstract class HudsonTestCase extends TestCase implements RootAction {
     }
 
     /**
-     * Create a new slave on the local host and wait for it to come onilne
+     * Create a new slave on the local host and wait for it to come online
      * before returning.
      */
     public DumbSlave createOnlineSlave() throws Exception {
@@ -688,7 +688,7 @@ public abstract class HudsonTestCase extends TestCase implements RootAction {
     }
     
     /**
-     * Create a new slave on the local host and wait for it to come onilne
+     * Create a new slave on the local host and wait for it to come online
      * before returning.
      */
     public DumbSlave createOnlineSlave(Label l) throws Exception {
@@ -1024,7 +1024,7 @@ public abstract class HudsonTestCase extends TestCase implements RootAction {
     }
 
     /**
-     * Submits the form by clikcing the submit button of the given name.
+     * Submits the form by clicking the submit button of the given name.
      *
      * @param name
      *      This corresponds to the @name of {@code <f:submit />}
@@ -1444,7 +1444,7 @@ public abstract class HudsonTestCase extends TestCase implements RootAction {
 
             setAlertHandler(new AlertHandler() {
                 public void handleAlert(Page page, String message) {
-                    throw new AssertionError("Alert dialog poped up: "+message);
+                    throw new AssertionError("Alert dialog popped up: "+message);
                 }
             });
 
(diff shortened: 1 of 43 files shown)


batmat:
@darxriggs I just deployed a `2.45-20190207.140250-1` SNAPSHOT of this PR. Can you please file a PR against the core to show this still builds fine there with this change?

I.e. like https://github.com/jenkinsci/jenkins/pull/3053/files but with `2.45-20190207.140250-1` instead.

Thanks!

batmat:
@darxriggs sorry, I forgot about the fact you'd be blocked by ongoing fixes for bumping to 2.46 already in progress in https://github.com/jenkinsci/jenkins/pull/3872.

I would be inclined to wait for https://github.com/jenkinsci/jenkins/pull/3872 to land so we can test your PR here.

Sorry about that

batmat:
@darxriggs finally, I went much faster than anticipated. Given https://github.com/jenkinsci/jenkins/pull/3872 is now merged, you can move forward by checking https://github.com/jenkinsci/jenkins-test-harness/pull/118#issuecomment-461431062 :).

thanks!

darxriggs:
@batmat the verification with Jenkins core passed.

src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
```diff
@@ -1007,7 +1005,7 @@ public Object call() throws Exception {
      */
     private List<String> listProperties(String properties) {
         List<String> props = new ArrayList<String>(Arrays.asList(properties.split(",")));
-        for (String p : props.toArray(new String[props.size()])) {
+        for (String p : props.toArray(new String[0])) {
```
batmat: This looks quite wrong to me, care to explain?

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -847,7 +847,7 @@ public UserDetails loadUserByUsername(String username) throws UsernameNotFoundEx
                     auths.add(new GrantedAuthorityImpl(g));
                 }
             }
-            return new org.acegisecurity.userdetails.User(username,"",true,true,true,true, auths.toArray(new GrantedAuthority[auths.size()]));
+            return new org.acegisecurity.userdetails.User(username,"",true,true,true,true, auths.toArray(new GrantedAuthority[0]));
```
batmat: Same. Why a 0 length array?

src/main/java/org/jvnet/hudson/test/SupportLogFormatter.java
```diff
@@ -21,26 +21,25 @@
...
 import java.util.logging.Formatter;
 import java.util.logging.LogRecord;
-/***********************************************
- * DO NOT INCLUDE ANY NON JDK CLASSES IN HERE.
- * IT CAN DEADLOCK REMOTING - SEE JENKINS-32622
- ***********************************************/
+// *********************************************
+// DO NOT INCLUDE ANY NON JDK CLASSES IN HERE.
+// IT CAN DEADLOCK REMOTING - SEE JENKINS-32622
+// *********************************************
```
batmat: Err, what? :) Why this?

src/main/preset-data/package.groovy
```diff
@@ -21,21 +21,21 @@
...
-def baseDir = project.basedir;
+def baseDir = project.basedir
 
-File targetDir = new File(baseDir,"target/classes");
-targetDir.mkdirs();
+def targetDir = new File(baseDir, "target/classes")
+targetDir.mkdirs()
 
-File dataSetRoot = new File(baseDir,"src/main/preset-data");
+def dataSetRoot = new File(baseDir, "src/main/preset-data")
```
batmat: This definitely looks like a change in the wrong direction to me.

darxriggs: You refert to `def` instead of `File`?
That's reducing type duplication. 
Groovy is a dynamic language.
Having types everywhere doesn't provide much more confidence as a lot of things are just checked as runtime unless `@CompileStatic` is used.
But I can revert this if really desired.
Just let me know.

src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
darxriggs: The commit comment contains the explanation. It also refers to https://shipilev.net/blog/2016/arrays-wisdom-ancients/ with a very detailed analysis.

src/main/java/org/jvnet/hudson/test/SupportLogFormatter.java
darxriggs: A Javadoc describes a class, method or field - also see https://stackoverflow.com/a/43874598/2847174.
IntelliJ IDEA marks everything as "dangling".
Therefore I changed this to "normal" comments.

src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
jvz: That applies to copying arrays. In this line, you are only iterating over them, so you don't need to copy. It should just be:

```suggestion
        for (String p : props) {
```

src/main/preset-data/package.groovy
jvz: I'm not sure on the precise context, but even in Java, this is idiomatic now. For example:

```java
var f = new File(...)
```

versus:

```java
File f = getRoot();
```

Essentially, you should have an obvious type in the line before using `def`/`var`.

src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
darxriggs: @jvz your suggested change is not applicable as it would result in a `ConcurrentModificationException` when removing elements from the list while iterating over it.

darxriggs:
As I answered all review comments I'd like to ask you to have another look and either merge the changes or request specific modifications.

src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
jvz: What if you made props a `CopyOnWriteArrayList` instead?

darxriggs: That's definitely an option. I updated the original commit and simplied it even further - see 288464dffc8f1d3f0708287d9ba6bc3f0f4a7d5a.

batmat:
@darxriggs I just pushed a SNAPSHOT for the current tip `2.47-20190221.102815-1`.

Can you please file a similar on the core with it? Once done and green, I'll merge and release here.

Thanks a lot for the cleanup!

batmat:
(sorry if this looks convoluted, we just had a dumpster fire on the previous changes and bump to JTH 2.46, so I'm being extra careful we can then just release a bump the core with latest JTH without outstanding efforts. Thanks for your understanding)

PR #119, opened 2019-01-02
[JENKINS-55381] Add Configuration-as-Code support to TestCrumbIssuer

darxriggs:


src/main/java/org/jvnet/hudson/test/TestCrumbIssuer.java
```diff
@@ -5,39 +5,39 @@
...
-import org.kohsuke.stapler.StaplerRequest;
-
 import hudson.Extension;
 import hudson.model.ModelObject;
 import hudson.security.csrf.CrumbIssuer;
 import hudson.security.csrf.CrumbIssuerDescriptor;
+import net.sf.json.JSONObject;
+import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.StaplerRequest;
+import javax.servlet.ServletRequest;
```
olivergondza: Please revert unnecessary whitespace changes. Yes, the formatting is odd but commit history spoiled with irrelevant changes is worse.

olivergondza:
The code looks good to me. Can you add a downstream PR that is using this feature to make sure this improves the situation? Thanks

src/main/java/org/jvnet/hudson/test/TestCrumbIssuer.java
olivergondza: Well actually, since it is in separate commit, I am fine with it.

darxriggs:
> Can you add a downstream PR that is using this feature to make sure this improves the situation?

@olivergondza, I didn't get this. Could you please tell a bit in more detail what you mean by this and which is the downstream repository?

darxriggs:
The tests in the gradle plugin don't directly require this change here. They assert on the "gradle" tool property only in the exported configuration (as can be seen [here](https://github.com/jenkinsci/gradle-plugin/blob/efc7421b58fbe1a06704f2a0966dcc8c9760ba42/src/test/groovy/hudson/plugins/gradle/ConfigurationAsCodeTest.groovy#L49-L54)). Therefore ignoring the stacktrace in the exported configuration. I would leave the tests exactly this way as that's what they should test.

But during development of these tests or when an assertions fails, the stacktrace (as reported in the ticket [JENKINS-55381](https://issues.jenkins-ci.org/browse/JENKINS-55381)) catches one's eyes in the exported configuration (and the WARNING log entry in the Jenkins logs) confuses people.

I added pull request #120 containing a test for verification. Is this ok for you?

olivergondza:
@darxriggs, awesome, thanks! I will merge this unless someone objects in coming days.

PR #121, opened 2019-01-05
Fix tests when run under Jenkins 2.x

darxriggs:
When tests were run with jenkins-war >=2.7.4, one failed with this error:
> WARNING jenkins.model.Jenkins$7#runTask: Loading plugin Hudson Task Scanner Plug-in v2.23 (tasks) failed perhaps due to plugin dependency issues
> java.lang.ClassNotFoundException: hudson.maven.MavenReporters
That's because the maven plugin is now under detached-plugins in jenkins-war since Jenkins 2.x.

Therefore for testing the `@WithPlugin` annotation another plugin was selected that
- supports Jenkins 1.x and 2.x
- has a small .jpi file (to keep this repository size small)
- has no dependencies (it would require these .jpi files in this repository)

New plugin downloaded from:
https://updates.jenkins-ci.org/download/plugins/keep-slave-disconnected/1.0/keep-slave-disconnected.hpi

Tests now pass for these Jenkins LTS versions (always latest patchlevel):
2.7.4, 2.19.4, 2.32.3, 2.46.3, 2.60.3, 2.73.3, 2.89.4, 2.107.3, 2.121.3

Beginning with 2.138.1 `JenkinsRuleTest#testTokenHelperMethods` fails, which is not covered in this fix.


jglick:
Trying to [ping CI](https://ci.jenkins.io/job/Core/job/jenkins-test-harness/indexing/events).

PR #122, opened 2019-01-19
bump JTH-htmlunit to 2.31-2

thomasgl-orange:
This is the follow-up after #89 and jenkinsci/jenkins-test-harness-htmlunit#4.


batmat:
Released as 2.46.

PR #126, opened 2019-03-07
Disable assertion on Java 9+

batmat:
cc @jenkinsci/java11-support 

To be used for https://github.com/jenkinsci/workflow-job-plugin/blob/72eecff5f7bd98aa4e73368c31203c50b3e89149/src/test/java/org/jenkinsci/plugins/workflow/job/MemoryCleanupTest.java#L35 fix on Java 11. Fix upcoming.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -151,6 +151,11 @@ THE SOFTWARE.
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci</groupId>
+      <artifactId>version-number</artifactId>
+      <version>1.6</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci</groupId>
       <artifactId>test-annotations</artifactId>
       <version>1.2</version>
       <scope>test</scope>
--- a/src/main/java/org/jvnet/hudson/test/MemoryAssert.java
+++ b/src/main/java/org/jvnet/hudson/test/MemoryAssert.java
@@ -40,6 +40,8 @@ import java.util.concurrent.Callable;
 import javax.swing.BoundedRangeModel;
 import jenkins.model.Jenkins;
 import static org.junit.Assert.*;
+import static org.junit.Assume.assumeTrue;
+
 import org.netbeans.insane.impl.LiveEngine;
 import org.netbeans.insane.live.LiveReferences;
 import org.netbeans.insane.live.Path;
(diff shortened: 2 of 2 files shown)


src/main/java/org/jvnet/hudson/test/MemoryAssert.java
```diff
@@ -146,13 +148,16 @@ public static void assertGC(WeakReference<?> reference) {
...
+     * <strong>Assumes Java runtime is &le; Java 8. I.e. tests will be skipped if Java 9+</strong>
      * Forces GC by causing an OOM and then verifies the given {@link WeakReference} has been garbage collected.
      * @param reference object used to verify garbage collection.
      * @param allowSoft if true, pass even if {@link SoftReference}s apparently needed to be cleared by forcing an {@link OutOfMemoryError};
      *                  if false, fail in such a case (though the failure will be slow)
      */
     @SuppressWarnings("DLS_DEAD_LOCAL_STORE_OF_NULL")
     public static void assertGC(WeakReference<?> reference, boolean allowSoft) {
+        // Disabled on Java 9+, because below will call Netbeans Insane Engine, which in turns tries to call setAccessible
+        assumeTrue(System.getProperty("java.specification.version").startsWith("1."));
```
MarkEWaite: Comment says Java 9 but assertion does not mention 8.  Would it be worth considering:
```suggestion
        assumeTrue(System.getProperty("java.specification.version").startsWith("1.8"));
```

dwnusbaum: I'm not sure if lib-version-number is in scope here, but if so, it has `io.jenkins.lib.versionnumber.JavaSpecificationVersion` which could be used to make this `JavaSpecificationVersion.forCurrentJVM().isOlderThanOrEqualTo(JavaSpecificationVersion.JAVA_8);` as @alecharp did in jenkinsci/script-security-plugin#238.

batmat: Done, good idea

batmat:
@markewaite it is applicable to ≤8 hence why I had written this this way.
But anyway now it should be clearer. Thanks!

pom.xml
```diff
@@ -149,6 +149,11 @@ THE SOFTWARE.
       <artifactId>findbugs-annotations</artifactId>
       <version>1.3.9-1</version>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci</groupId>
+      <artifactId>version-number</artifactId>
```
jglick: Is this not already available from `jenkins-core`?

src/main/java/org/jvnet/hudson/test/MemoryAssert.java
```diff
@@ -146,13 +150,16 @@ public static void assertGC(WeakReference<?> reference) {
     }
 
     /**
+     * <strong>Assumes Java runtime is &le; Java 8. I.e. tests will be skipped if Java 9+</strong>
```
jglick: Wrong location for a summary sentence.

PR #127, opened 2019-04-02
Added JenkinsRule.showAgentLogs utility

jglick:
Developed during JEP-210 work but could be broadly useful.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -991,10 +991,10 @@ public class JenkinsRule implements TestRule, MethodRule, RootAction {
     }
 
     /**
-     * Same as {@link #showSlaveLogs(Slave, Map)} but taking a preconfigured list of loggers as a convenience.
+     * Same as {@link #showAgentLogs(Slave, Map)} but taking a preconfigured list of loggers as a convenience.
      */
-    public void showSlaveLogs(Slave s, LoggerRule loggerRule) throws Exception {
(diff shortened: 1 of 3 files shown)


src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -988,6 +990,53 @@ public void waitOnline(Slave s) throws Exception {
         }
     }
 
+    /**
+     * Same as {@link #showSlaveLogs(Slave, Map)} but taking a preconfigured list of loggers as a convenience.
+     */
+    public void showSlaveLogs(Slave s, LoggerRule loggerRule) throws Exception {
```
batmat: I'm sorry to arrive late, but can you please rename to `showAgentLogs` instead? let's not keep adding new code about this.

jglick: No strong feelings either way, just I was matching

* the type `Slave` used in the method signature
* the existing methods right next to this (`createSlave`, `createOnlineSlave`)

batmat: Yeah, I know we'll be more inconsistent, but I think we need to start moving in that direction.

jglick: OK, can change it.

batmat: So, I won't block this alone if someone else is fine with adding more of these. But again I'd rather see us start removing this phrasing, even if this unfortunately raises a bit the code naming inconsistency.

PR #129, opened 2019-04-10
[HOSTING-742] Using external lib-support-log-formatter

jglick:
Matches https://github.com/jenkinsci/support-core-plugin/pull/168.


src/main/java/org/jvnet/hudson/test/SupportLogFormatter.java
```diff
@@ -1,210 +0,0 @@
...
-
-/**
- * Format log files in a nicer format that is easier to read and search.
- *
- * @author Stephen Connolly
- */
-class SupportLogFormatter extends Formatter {
-
-    private static long start = System.nanoTime();
-    static String elapsedTime() { // modified from original
```
batmat: Ok, I think I get the DeltaSupportLogFormatter class, because this method got added here and hence is not in https://github.com/jenkinsci/lib-support-log-formatter/blob/master/src/main/java/io/jenkins/lib/support_log_formatter/SupportLogFormatter.java

jglick: Correct. For the specialized use case of functional tests, I found it more helpful to show seconds elapsed since start of test, rather than absolute timestamps.

PR #130, opened 2019-04-26
Add Hamcrest matchers for Secret and Pattern

jvz:
I noticed while reviewing https://github.com/jenkinsci/configuration-as-code-plugin/pull/862 that it'd be nice to have matchers for Secret. I also added one for regular expressions as Hamcrest doesn't come with one by default despite being rather generic.

@reviewbybees


casz:
Uhh I have several things I want added was not aware this was the repo to submit those to 😅

jvz:
Another alternative is [AssertJ](https://joel-costigliola.github.io/assertj/), though not many Jenkins plugins use it.

PR #131, opened 2019-04-30
Incrementalified

jglick:
Inspired by https://github.com/jenkinsci/swarm-plugin/pull/81#discussion_r279812868.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,31 +1,21 @@
-pipeline {
-    options {
-        buildDiscarder(logRotator(numToKeepStr: '20'))
-        timeout(time: 1, unit: 'HOURS')
-    }
-    agent {
-        docker {
-            image 'maven:3.5.0-jdk-8'
-            label 'docker'
+properties([buildDiscarder(logRotator(numToKeepStr: '20'))])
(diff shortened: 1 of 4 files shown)


jglick:
Still awaiting a fix or workaround for [JENKINS-54126](https://issues.jenkins-ci.org/browse/JENKINS-54126).

Jenkinsfile
```diff
@@ -13,11 +13,17 @@ pipeline {
         stage('main') {
             // TODO Windows build in parallel
             steps {
-                sh 'mvn -B clean verify'
+                sh 'mvn -B clean verify -Dset.changelist'
             }
             post {
                 success {
                     junit '**/target/surefire-reports/TEST-*.xml'
+                    dir("$HOME/.m2/repository") {
```
jglick: `$HOME` is unset, or just `?`. Need to use the `pwd tmp: true` trick. Not sure if that is even possible in Declarative.

oleg-nenashev:
It looks like there are some issues in this PR: https://github.com/jenkinsci/jenkins-test-harness/pull/135#issuecomment-498553050

PR #132, opened 2019-04-30
Allow port reuse in RestartableJenkinsRule

basil:
Inspired by [jenkinsci/swarm-plugin#81 (comment)](https://github.com/jenkinsci/swarm-plugin/pull/81#discussion_r279723719). Testing the Swarm client requires having Jenkins run on the same port before and after restart (c.f. jenkinsci/swarm-plugin#81). Here we add this functionality to `RestartableJenkinsRule`. The new functionality is opt-in to preserve existing behavior for tests which don't need it and which could be negatively impacted by races for open ports.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
@@ -10,6 +10,8 @@ import org.junit.runners.model.Statement;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.net.ServerSocket;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.LinkOption;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
```diff
@@ -62,8 +64,25 @@
...
+     */
+    private final int port;
+
     private static final Logger LOGGER = Logger.getLogger(HudsonTestCase.class.getName());
 
+    public RestartableJenkinsRule() {
+        this.port = 0;
+    }
+
+    public RestartableJenkinsRule(boolean reusePort) {
```
jglick: To avoid constructor arg pollution, maybe this could be a builder-like method `withReusedPort()` or similar?

basil: Done.

basil:
@jglick I think I've addressed all your comments. Do you have any further concerns, or can we move forward with this?

jglick:
It is in my review queue.

basil:
> It is in my review queue.

Got it, no rush and sorry for being pushy. Thanks for the quick reply!

oleg-nenashev:
Restarting CI

src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
```diff
@@ -62,8 +64,38 @@
      */
     public File home;
 
+    /**
+     * TCP/IP port that the server is listening on.
+     */
+    private final int port;
```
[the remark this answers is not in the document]
basil: > Maybe it is better to default to -1 and set a `@CheckNegativeValue` annotation in the getter logic, but t is rather an improvement

That would probably work, but 0 is already a special value in `JenkinsRule` and `org.eclipse.jetty.server.AbstractNetworkConnector`. I opted to keep the special value as 0 in `RestartableJenkinsRule` for consistency with those classes.

raul-arabaolaza:
I was worried about the port reuse to be the default and break parallel tests, but it seems optional so LGTM

oleg-nenashev:
I hope to cut the new release tomorrow

PR #133, opened 2019-05-25
Require a Stapler verb annotation on web method looking methods

daniel-beck:
The current releases of matrix-auth would behave as follows with this change:

```
[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   PluginAutomaticTestBuilder$OtherTests.testStaplerDispatches:155 Web methods lack verb annotations like @RequirePOST, @GET, @POST, etc.
Expected: is an empty collection
     but: <[hudson.security.GlobalMatrixAuthorizationStrategy$DescriptorImpl#doCheckName, org.jenkinsci.plugins.matrixauth.AuthorizationContainerDescriptor#doCheckName_, org.jenkinsci.plugins.matrixauth.AuthorizationMatrixNodeProperty$DescriptorImpl#doCheckName]>
[INFO] 
[ERROR] Tests run: 54, Failures: 1, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
```

Unsure whether all the rules are 100% matching what Stapler does, but it should be close enough.


Wadeck:
> there's no straightforward migration path there

We can split the method into two, annotate one with `@GET` and the second with `@POST`. One of the difference with `@RequirePOST`, the two previous ones do not fail the request if the method is incorrect.

daniel-beck:
> We can split the method into two

Well, yeah, but that's not exactly straightforward IMO, especially if faces with a few dozen of these in some plugins. The code needs to change nontrivially. We probably need `@WebMethod` too, or achieve different method signatures through creative argument declarations.

> Do you plan to add a blog post explaining the reason of this hardening?

Yes, plus a better redirect target. It should take a while for this to get picked up by plugins anyway.

----

A potential negative side effect of this would be that we cannot change form validation request methods at will anymore, without breaking lots of plugins. Looking at you, https://github.com/jenkinsci/jenkins/commit/09d60462b9edf775f08568601bb3e2cfd8075368.

src/main/java/org/jvnet/hudson/test/PluginAutomaticTestBuilder.java
```diff
@@ -65,6 +84,7 @@ public static TestSuite build(Map<String,?> params) throws Exception {
             String packaging = StringUtils.defaultIfBlank((String)params.get("packaging"), "hpi");
             if ("hpi".equals(packaging)) {
                 inJenkins.addTest(new OtherTests("testPluginActive", params));
+                inJenkins.addTest(new OtherTests("testStaplerDispatches", params));
```
jtnord: I would rather other things did not go into `InjectedTest` (and that it would die).  The reason being is that skipping part of the tests (rather than all of them) is pretty much impossible as it uses old junit Suites rather than newer test classes, and is hard to debug.
This would be much better going forward in a `generate-tests` mojo in maven-hpi-plugin which then also makes it much easier to debug in an IDE.  but otherwise 👍

src/main/java/org/jvnet/hudson/test/PluginAutomaticTestBuilder.java
```diff
@@ -106,5 +126,88 @@ public void testPluginActive() {
...
+            Assert.assertThat("There should be no web methods that lack HTTP verb annotations like @RequirePOST, @GET, @POST, etc. -- see https://jenkins.io/redirect/developer/csrf-protection", methodsFound, is(empty()));
+        }
+
+        private Method findIsRoutableMethod() throws NoSuchMethodException {
+            try {
+                Method method = Class.forName("jenkins.security.stapler.TypedFilter").getDeclaredMethod("isRoutableMethod", Method.class);
+                method.setAccessible(true);
+                return method;
+            } catch (ClassNotFoundException e) {
+                LOGGER.warning("This test requires Jenkins 2.154, Jenkins LTS 2.138.4, or newer to run, use e.g. -Djenkins.version=2.138.4");
```
jvz: Use of `assume` from JUnit might be appropriate here for better integration with error reporting.

oleg-nenashev:
Needs a merge conflict fix + clarity on TODO items in the PR description

daniel-beck:
PR build fails because of https://github.com/jenkinsci/jenkins-test-harness/pull/167#issuecomment-609078264

jglick:
This would be better put on hold until https://github.com/jenkinsci/jenkins/pull/4623 is merged, or exclude `doCheckXXX` methods, or exclude them unless the Jenkins version indicates that https://github.com/jenkinsci/jenkins/pull/4623 is included.

daniel-beck:
> This would be better put on hold until jenkinsci/jenkins#4623 is merged

Unsure. Would be beneficial, but IMO overall doesn't make a big difference if the docs say:

* Use `@POST` for any `doCheck*` and add `checkMethod="post"` or your plugin will break once jenkinsci/jenkins#4623 is merged
* Use `@POST` for any `doCheck*` and add `checkMethod="post"` or your plugin will break on Jenkins 2.2xy and newer

The advice needs to favor `@POST` (or `@RequirePOST`) for most endpoints anyway; only few, such as `doWs`, `doArtifact`, `doIndex` should be annotated `@GET`.

jglick:
Agreed, it is fine so long as the docs guide users to use `@POST` rather than the HTTP method currently being sent.

daniel-beck:
I still need to figure some stuff out.

PR #141, opened 2019-06-17
[JENKINS-58038] Use Annotation-Indexer instead of Reflections

AbhyudayaSharma:
Reflections was causing a dependency conflict with the Guava from
Jenkins Core (see: https://travis-ci.org/jenkinsci/configuration-as-code-plugin/jobs/546110907#L878)

@jenkinsci/gsoc2019-role-strategy

### DIFF
--- a/src/main/java/jenkins/benchmark/jmh/BenchmarkFinder.java
+++ b/src/main/java/jenkins/benchmark/jmh/BenchmarkFinder.java
@@ -1,8 +1,8 @@
 package jenkins.benchmark.jmh;
 
-import io.github.classgraph.ClassGraph;
-import io.github.classgraph.ClassInfo;
-import io.github.classgraph.ScanResult;
+
+import net.java.sezpoz.Index;
+import net.java.sezpoz.IndexItem;
 import org.openjdk.jmh.runner.options.ChainedOptionsBuilder;
 
 /**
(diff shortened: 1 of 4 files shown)


oleg-nenashev:
Thanks for the investigation @AbhyudayaSharma! Maybe it would be even better to use [sezpoz](https://github.com/jglick/sezpoz) since it is already included into JTH. 

A Jenkins JIRA ticket would be also much appreciated since the previous JTH version has been already released

casz:
Green built in downstream PR: https://github.com/jenkinsci/configuration-as-code-plugin/pull/921

oleg-nenashev:
My plan is to merge and release it on the evening if there is no negative feedback

src/main/java/jenkins/benchmark/jmh/BenchmarkFinder.java
```diff
@@ -1,40 +1,26 @@
...
-        Set<Class<?>> benchmarkClasses = reflections.getTypesAnnotatedWith(JmhBenchmark.class);
-        benchmarkClasses.forEach(clazz -> {
-            JmhBenchmark annotation = clazz.getAnnotation(JmhBenchmark.class);
-            if (Objects.nonNull(annotation)) {
-                optionsBuilder.include(clazz.getName() + annotation.value());
-            }
-        });
+    public static void findBenchmarks(ChainedOptionsBuilder optionsBuilder) {
+        for (IndexItem<JmhBenchmark, Object> item : Index.load(JmhBenchmark.class, Object.class)) {
+            optionsBuilder.include(item.className() + item.annotation().value());
```
jglick: If you have to call `IndexItem.className` you are not using SezPoz correctly. If you do not care about lazy loading and just want to enumerate classes marked with a certain annotation, then use `org.jenkins-ci:annotation-indexer` (also already in the classpath).

PR #142, opened 2019-06-17
Add support for running benchmarks with WebClient

AbhyudayaSharma:
Adds support for running benchmarks using `WebClient`.

@jenkinsci/gsoc2019-role-strategy

### DIFF
--- a/src/main/java/jenkins/benchmark/jmh/JmhJenkinsRule.java
+++ b/src/main/java/jenkins/benchmark/jmh/JmhJenkinsRule.java
@@ -1,0 +1,44 @@
+package jenkins.benchmark.jmh;
+
+import com.gargoylesoftware.htmlunit.Cache;
+import jenkins.model.Jenkins;
+import org.jvnet.hudson.test.JenkinsRule;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.Objects;
+
+/**
+ * Extension of {@link JenkinsRule} to allow it to be used from JMH benchmarks.
+ * <p>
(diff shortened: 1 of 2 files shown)


AbhyudayaSharma:
> How closely is `WebClient` tied to JenkinsRule

From a quick look at the code, it is possible to extract `WebClient` from `JenkinsRule`. `WebClient` basically depends on the reference to the Jenkins instance, it's URL and a private method related to CrumbIssuer. @oleg-nenashev Do you recommend to extract it out of `JenkinsRule`? Also, in the above test benchmark, we directly instantiate the WebClient just because it is running on a single thread. In role strategy, when we were running it on two threads, we had to create a separate state because it is not thread-safe. 

https://github.com/jenkinsci/role-strategy-plugin/blob/2ce1b5ad0373387fcfd4b26b688087c261283957/src/test/java/jmh/benchmarks/WebClientBenchmark.java#L21-L29

From our benchmarks, the only thing I see that is different is the URL because the port it runs is different every time. Please let me know what you think about it.

AbhyudayaSharma:
Or we could just have another utility class that creates a `WebClient` for each thread in the benchmark and the overall benchmark class could look something like:
```java
@JmhBenchmark
public class SomeBenchmark {
    public static class State1 extends JmhJenkinsRule {}
    
    public static class ThreadState extends WebClientThreadState {}

    public void benchmark() {
        ...
    }
}
```

oleg-nenashev:
Let's park it for a while I will release #141 without it, so we won't need to rush this PR. Also, a merge conflict

PR #158, opened 2019-07-29
Switch from GMaven to GMavenPlus

basil:
Inspired by [jenkinsci/jenkins-test-harness#156 (comment)](https://github.com/jenkinsci/jenkins-test-harness/pull/156#pullrequestreview-267614033). Supersedes jenkinsci/jenkins-test-harness#156. This repository currently uses the GMavenPlugin to run a `package.groovy` to zip a few files during the build. The GMaven plugin is no longer maintained; active development is now taking place in the GMavenPlus project. Therefore, we switch to GMavenPlus.

After updating the group ID, artifact ID, and version, I got this error:

```
[ERROR] Failed to execute goal org.codehaus.gmavenplus:gmavenplus-plugin:1.7.1:execute (preset-packager) on project jenkins-test-harness: The parameters 'scripts' for goal org.codehaus.gmavenplus:gmavenplus-plugin:1.7.1:execute are missing or invalid -> [Help 1]
```

Looking into it further, I found that the calling conventions had changed. I updated them following the instructions in the [upstream wiki](https://github.com/groovy/GMavenPlus/wiki/Examples#execute-scripts).

With that, the goal passes, using the version of Groovy on the classpath from Jenkins core:

```
[INFO] --- gmavenplus-plugin:1.7.1:execute (preset-packager) @ jenkins-test-harness ---
[INFO] Using Groovy 2.4.8 to perform execute.
[INFO] Running Groovy script from file:////home/basil/work/third-party/jenkinsci/jenkins-test-harness/src/main/preset-data/package.groovy.
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -248,17 +248,6 @@ THE SOFTWARE.
             <artifactId>ant</artifactId>
             <version>1.10.6</version>
           </dependency>
-          <!-- Usually a dependency of ant, but some people seem to have an incomplete ant POM. See JENKINS-11416 -->
-          <dependency>
-            <groupId>org.apache.ant</groupId>
-            <artifactId>ant-launcher</artifactId>
-            <version>1.8.0</version>
-          </dependency>
-          <dependency>
-            <groupId>org.codehaus.gmaven.runtime</groupId>
-            <artifactId>gmaven-runtime-2.0</artifactId>
-            <version>1.5-jenkins-1</version>
-          </dependency>
         </dependencies>
       </plugin>
     </plugins>
--- a/pom.xml
+++ b/pom.xml
@@ -225,9 +225,9 @@ THE SOFTWARE.
         </configuration>
       </plugin>
       <plugin>
-        <groupId>org.codehaus.gmaven</groupId>
-        <artifactId>gmaven-plugin</artifactId>
-        <!-- version specified in grandparent pom -->
+        <groupId>org.codehaus.gmavenplus</groupId>
+        <artifactId>gmavenplus-plugin</artifactId>
+        <version>1.7.1</version>
         <executions>
           <execution>
             <id>preset-packager</id>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -236,7 +236,9 @@ THE SOFTWARE.
               <goal>execute</goal>
             </goals>
             <configuration>
-              <source>${pom.basedir}/src/main/preset-data/package.groovy</source>
+              <scripts>
+                <script>file:///${pom.basedir}/src/main/preset-data/package.groovy</script>
```
jglick: Ouch, hope this works on Windows and in directories with spaces in their names. Constructing `file`-protocol URLs via string concatenation (or, conversely, extracting the file path using string matching) is notoriously unsafe. You need to use proper methods like `File.toUri` or (better, on UNC paths) `Path.toUri`. Oh well, guess it works well enough.

basil: > Ouch, hope this works on Windows and in directories with spaces in their names.

Can't speak to Windows, but it does work in directories with spaces in their names:

```
[INFO] --- gmavenplus-plugin:1.7.1:execute (preset-packager) @ jenkins-test-harness ---
[INFO] Using Groovy 2.4.8 to perform execute.
[INFO] Running Groovy script from file:////home/basil/work/third-party/jenkinsci/jenkins test harness/src/main/preset-data/package.groovy.
```

basil:
> It should work on Windows, but indeed some integration tests for that would not hurt.

Note that this code only executes when _building_ the test harness, not when _consuming_ the test harness. I don't have a Windows developer environment, and the `Jenkinsfile` for this repository doesn't build the test harness on Windows (and I have no write access to the repository in order to try to update the `Jenkinsfile`), so I effectively have no way of testing this on Windows.

pom.xml
jglick: Maybe whatever code is _using_ the URL is similarly lax, and two wrongs make a right!

jglick:
I have a Windows development environment so I can check it soon.

oleg-nenashev:
Will merge once @jglick provides feedback. I do not mind not being able to build JTH in Windows, WSL will help

jglick:
Works fine on Windows 10.

PR #159, opened 2019-08-02
[JENKINS-58571] Fix default value for `isNameEditable()` on MockFolder.

aHenryJard:
I need this modification to be able to merge this PR: https://github.com/jenkinsci/jenkins/pull/4122
Test are failing there because the default value for isNameEditable is false, and the PR add a protection when trying to rename a Item with isNameEditable false. MockFolder used in test does not override isNameEditable to true but it should.

The isNameEditable method is available only on jenkins-core > 2.110, so I upgrade jenkins-war to the first version including this method.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -82,7 +82,7 @@ THE SOFTWARE.
     <dependency>
       <groupId>org.jenkins-ci.main</groupId>
       <artifactId>jenkins-war</artifactId>
-      <version>2.110</version>
+      <version>2.60.3</version>
       <type>executable-war</type>
       <exclusions>
         <exclusion>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -82,7 +82,7 @@ THE SOFTWARE.
     <dependency>
       <groupId>org.jenkins-ci.main</groupId>
       <artifactId>jenkins-war</artifactId>
-      <version>2.60.3</version>
+      <version>2.110</version>
```
jglick: We prefer to use LTS baselines here.

Rather than doing this change, I would suggest just adding the method with

```java
/* TODO uncomment when core dep ≥ 2.110:
@Override
*/
```

since the JVM will still consider it an override when used against sufficiently new cores.

aHenryJard: it's done.

oleg-nenashev:
Thanks for your contribution @aHenryJard !

PR #167, opened 2019-09-27
Introduce new `JenkinsRule#buildAndAssertStatus()` method

slonopotamus:
This is identical to JenkinsRule.buildAndAssertSuccess but with user-specified build status

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -1368,22 +1368,30 @@ public class JenkinsRule implements TestRule, MethodRule, RootAction {
         return assertBuildStatus(Result.SUCCESS, r);
     }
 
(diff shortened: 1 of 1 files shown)


slonopotamus:
Done. I've also adjusted generics a bit so it now works without unchecked cast. This *might* break compilation of some tests that were giving something that is a `Job` but is not a `ParameterizedJobMixIn.ParameterizedJob` but such cases would fail with assert anyway.

oleg-nenashev:
Thanks!

daniel-beck:
This breaks with any newer core dependency. If you
```
$ git diff
diff --git a/pom.xml b/pom.xml
index c63b60d..586a3dd 100644
--- a/pom.xml
+++ b/pom.xml
@@ -85,7 +85,7 @@ THE SOFTWARE.
     <dependency>
-      <version>2.60.3</version>
+      <version>2.73.3</version>
       <type>executable-war</type>
       <exclusions>
         <exclusion>
```

then you get

> `JenkinsRule.java:[1377,64] type argument J is not within bounds of type-variable JobT`

PR #187, opened 2019-12-10
[JENKINS-55381] Add test for TestCrumbIssuer compatibility with JCasC

darxriggs:
Now that Jenkins 2.60.3 is used, this test can be added. It was originally provided already as #120 to test #119 but could not be merged at that time due to the dependency on a newer Jenkins version.

### DIFF
--- a/src/test/java/org/jvnet/hudson/test/TestCrumbIssuerConfigurationAsCodeTest.java
+++ b/src/test/java/org/jvnet/hudson/test/TestCrumbIssuerConfigurationAsCodeTest.java
@@ -1,0 +1,38 @@
+package org.jvnet.hudson.test;
+
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.not;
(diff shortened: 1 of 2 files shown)


src/test/java/org/jvnet/hudson/test/TestCrumbIssuerConfigurationAsCodeTest.java
```diff
@@ -0,0 +1,38 @@
...
+public class TestCrumbIssuerConfigurationAsCodeTest {
+
+    private final JenkinsRule j = new JenkinsRule();
+    private final LoggerRule logRule = new LoggerRule().record("io.jenkins.plugins.casc", Level.INFO).capture(1000);
+
+    @Rule
+    public RuleChain chain = RuleChain.outerRule(j).around(logRule);
+
+    @Test
+    public void testCrumpIssuerShouldBeSupportedWhenExportingConfiguration() throws Exception {
```
timja: Crump :D

PR #191, opened 2020-01-06
Add the GitHub repo URLs to the Maven metadata

casz:
I doubt dependabot can resolve variables in the URL as it reading XML directly instead of parsing variables.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -37,12 +37,13 @@ THE SOFTWARE.
   <version>${revision}${changelist}</version>
 
   <name>Test harness for Jenkins and plugins</name>
+  <url>https://github.com/jenkinsci/jenkins-test-harness</url>
   <description>Harness used to run functional tests of Jenkins core and plugins.</description>
 
   <scm>
     <connection>scm:git:git://github.com/jenkinsci/${project.artifactId}.git</connection>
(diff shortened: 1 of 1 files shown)


jglick:
> I doubt dependabot can resolve variables in the URL

I am not following. Why would Dependabot care about this field?

casz:
See
https://github.com/dependabot/dependabot-core/blob/01e6563639a665ca650fb93c6244ccac0df39fa2/maven/lib/dependabot/maven/metadata_finder.rb#L52

pom.xml
```diff
@@ -37,12 +37,13 @@ THE SOFTWARE.
...
 
   <name>Test harness for Jenkins and plugins</name>
+  <url>https://github.com/jenkinsci/jenkins-test-harness</url>
   <description>Harness used to run functional tests of Jenkins core and plugins.</description>
 
   <scm>
     <connection>scm:git:git://github.com/jenkinsci/${project.artifactId}.git</connection>
     <developerConnection>scm:git:ssh://git@github.com/jenkinsci/${project.artifactId}.git</developerConnection>
-    <url>https://github.com/jenkinsci/${project.artifactId}</url>
+    <url>https://github.com/jenkinsci/jenkins-test-harness</url>
```
jglick: Would rather just fix the bug in Dependabot.

casz: There are tests that should cover the use case but it's clearly not happening. 🤔

PR #193, opened 2020-01-16
[JENKINS-60694] Do not use defaults for acceptors & selectors

jglick:
I think this fixes [JENKINS-60694](https://issues.jenkins-ci.org/browse/JENKINS-60694). Just copies the fix from #69 to `HudsonTestCase` which I should have done to begin with.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
+++ b/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
@@ -522,7 +522,7 @@ public abstract class HudsonTestCase extends TestCase implements RootAction {
         context.setMimeTypes(MIME_TYPES);
         context.getSecurityHandler().setLoginService(configureUserRealm());
 
-        ServerConnector connector = new ServerConnector(server);
+        ServerConnector connector = new ServerConnector(server, 1, 1);
 
         HttpConfiguration config = connector.getConnectionFactory(HttpConnectionFactory.class).getHttpConfiguration();
         // use a bigger buffer as Stapler traces can get pretty large on deeply nested URL
--- a/src/main/java/org/jvnet/hudson/test/JavaNetReverseProxy.java
+++ b/src/main/java/org/jvnet/hudson/test/JavaNetReverseProxy.java
@@ -43,7 +43,7 @@ public class JavaNetReverseProxy extends HttpServlet {
         ServletContextHandler root = new ServletContextHandler(contexts, "/", ServletContextHandler.SESSIONS);
         root.addServlet(new ServletHolder(this), "/");
 
-        ServerConnector connector = new ServerConnector(server);
+        ServerConnector connector = new ServerConnector(server, 1, 1);
         server.addConnector(connector);
         server.start();
 


MarkEWaite:
As far as I can tell (one sample run before the change that failed, then two sample runs after that passed), this also resolves [JENKINS-60574](https://issues.jenkins-ci.org/browse/JENKINS-60754) - parallel tests freeze on machine with many cores.

I was able to run to completion on that many-core machine with `mvn clean -Dconcurrency=1C verify` and with the default setting of `mvn clean verify`.

jglick:
> how is this not causing lots of radnom issues for the Jetty project?

I was wondering the same thing.

joakime:
> determinism is good - and I have tested that with this patch my reproduceable case now passes. but WUT how is this not causing lots of radnom issues for the Jetty project??

It doesn't.

The default selectors and acceptors are just fine for large core count machines - 64 cores+ (verified up to 128 cores)
We build / test / load-test regularly on these kinds of machines. (all with default Jetty configurations)
We even build / test with Jenkins 2.209 on these kinds of machines. (both as master and as an agent to an identical machine)

Going from default to your settings of 1 acceptor and 1 selector is fine and wont cause any issues.
It just limits your ability to handle high new incoming connection loads.

jglick:
> It just limits your ability to handle high new incoming connection loads…

Please bear in mind here that this PR is about the test harness. So the “core count” of the machine could be huge for a nonvirtualized CI server, or a few for a laptop, or just one if containers are in use and being throttled; and the expected number of concurrent connections is zero or one depending on the test.

joakime:
> So the “core count” of the machine could be huge for a nonvirtualized CI server.

You should know that not all virtualized CI servers have proper isolation of core count too.
There's a long history for this, so don't assume a virtualized CI server that is declared to have limited core counts will actually report said limit to that container's linux or java.
We experienced theses kind of false core count issues with our docker and google cloud layers, there are patches to docker, linux, and java to address these bugs (have to have all three patched for it to report the configured limits).
Jetty has a workaround for these classes of bugs in virtualized container as well.  
Use the system environment variable (or System property) `JETTY_AVAILABLE_PROCESSORS` to override the number of processors that are reported from `Runtime.getRuntime().availableProcessors()`.

>  or a few for a laptop

I have a consumer grade dell laptop with 16 cores in it (purchased 3 months ago), and a workstation with 64-cores (purchased a year ago).
High core count machines are getting rather common.

For breakdown on how defaults are calculated for acceptors and selectors, see https://github.com/eclipse/jetty.project/issues/4492#issuecomment-575628375

jtnord:
regardless this was reproduced on a windows machine outside of continaers and I am not aware of a JVM issue that can not report the cores correctly here.  Sure its virtualized and it may have some shared cores etc, but the core count should be accurate.

Regardless even if the core count was wrong I would not expect Jetty to **not service requests at all** (a simple get request was not responded to in 3+ minutes), rather I would expect performance lower (as threads contended).

res0nance:
I managed to reproduce this issue on a physical windows machine with
https://github.com/res0nance/injected-hang-plugin
Its an empty plugin with google's re2j library as a dependency on POM 3.55 injected hangs on POM 3.56 it does not hang.

So I doubt it has anything to do with virtualization.

Some machine specs:
Xeon E5-2650 v3 2 Socket (10 cores per socket)
```
java -version
openjdk version "1.8.0_232"
OpenJDK Runtime Environment (Zulu 8.42.0.14-SA-win64) (build 1.8.0_232-b18)
OpenJDK 64-Bit Server VM (Zulu 8.42.0.14-SA-win64) (build 25.232-b18, mixed mode)
```
`Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)`

PR #166, opened 2019-09-23
Stop swallowing exceptions that happen in `TestEnvironment#dispose()`

slonopotamus:
This commit is excepted to reveal bogus tests that do not properly shutdown,
especially on Windows where directory cannot be removed if there are any running processes that use it as a working directory.

PR #198, opened 2020-01-22
Simplified TemporaryDirectoryAllocator.dispose

jglick:


src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
```diff
@@ -85,25 +93,12 @@ public synchronized File allocate() throws IOException {
      * Deletes all allocated temporary directories.
      */
     public synchronized void dispose() throws IOException, InterruptedException {
-        // TODO when we bump the Jenkins dependency to 2.157
-        IOException x = null;
+        System.gc();
```
jtnord: requireing gc that is indicitive of a bug in the code. (something has not been closed yet has gone out of scope so close is called in the finalize.  would suggest not putting this here.

jglick: Well `DisablePluginCommandTest` reliably fails without this.

slonopotamus: I'm against adding a call to gc here. The whole point of #166 was to detect when files are not properly closed. Explicit call to gc would hide bugs.

jglick: …and I am not sure how to track down what got released in finalization. [`file-leak-detector`](https://file-leak-detector.kohsuke.org/) is close but not exactly what we want.

slonopotamus: jenkinsci/git-plugin#798 <-- these bugs would not be detected if there was GC call.

jglick: Fine; any suggestions on how to diagnose though? The errors are of various plugin JARs having a mandatory file lock. I know that is generally due to a `JarFile` or a `URLClassLoader` not being `close`d, but there are a lot of JAR accesses in the Jenkins codebase.

slonopotamus:
#166 does not guarantee it will detect a bug. But if it detects something, you may be sure that there is a bug in code. I want to say that "able to detect bugs unreliably" is better than "unable to detect bugs". The point of #166 is not to delete directory. Nobody cares if it is deleted or not, actually! The point is to catch buggy code that forgot to close file descriptors after finishing.

jenkinsci/git-plugin#798, jenkinsci/subversion-plugin#240 and jenkinsci/workflow-support-plugin#99 were all actual bugs in production code of plugins, discovered thanks to #166.

slonopotamus:
If you need some help figuring out what exact code is guilty in failed tests in jenkinsci/jenkins#4447 - I can definitely help with that.

src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
jtnord: I would check where the files are opened first - I am guessing that one of them may not be expected JarURLCLassLoader?

https://github.com/jenkinsci/jenkins-test-harness/blob/93e9b64a89d3048896f77d62fba84fff715b932b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java#L363-L373

I can not reproduce locally issues with `DisablePluginCommandTest` (so either yay for timing or a bug in a specific JDK)

jglick:
I see no reason to hold off on this PR just because https://github.com/jenkinsci/jenkins/pull/4447 is stalled.

src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
```diff
@@ -115,13 +110,34 @@ public synchronized void disposeAsync() {
...
+                        LOGGER.log(Level.WARNING, null, e);
                     }
+                }
             }
         }.start();
     }
+
+    private void delete(Path p) throws IOException {
+        LOGGER.fine(() -> "deleting " + p);
+        if (Files.isDirectory(p, LinkOption.NOFOLLOW_LINKS)) {
```
jtnord, line 127: probably also a junction which I think may be used in some test case and `FIlePath` respects.. (`FIles.readAttributes` and `isOther` - as a general well it is not a normal directory)

jglick: I do not know much about junctions. Would they ever really be used in `$JENKINS_HOME`?

jtnord: I'm sure something uses them as we have specific code in Jenkins...
https://github.com/jenkinsci/jenkins/blob/715cdb15b3e855f8b6115aa02200860ed6caf49d/core/src/main/java/hudson/Util.java#L323-L343
I have not come across a real life example, but I'm guessing that it is for a reason?

Maybe it is just an easy way in Jenkins to put some subset of `Job`s in a different drive / pool etc and would not actually be done *inside* a workspace, or a build directory by jenkins (I do not know if zipping on windows copies the Junction or not, so if a workspace was coppied from an agent where the build created a junction if that would be replcicated on a master).

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -510,13 +510,6 @@ public void after() throws Exception {
             try {
                 env.dispose();
             } finally {
-                // Hudson creates ClassLoaders for plugins that hold on to file descriptors of its jar files,
-                // but because there's no explicit dispose method on ClassLoader, they won't get GC-ed until
-                // at some later point, leading to possible file descriptor overflow. So encourage GC now.
-                // see http://bugs.sun.com/view_bug.do?bug_id=4950148
-                // TODO use URLClassLoader.close() in Java 7
```
slonopotamus, line 517: Does Jenkins still work on Java < 7? Maybe it's time to use `URLClassLoader.close()`?

jglick: Cannot recall offhand if we do already in core. Anyway this is just a super old comment which I am _deleting_.

src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
jglick: There might be some obscure use case in production servers, but I doubt this would affect functional tests. If there is some plugin which does create a junction during a test for some reason, then it can explicitly remove it before generic cleanup, right?

src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
```diff
@@ -116,13 +110,34 @@ public synchronized void disposeAsync() {
...
+            try (DirectoryStream<Path> children = Files.newDirectoryStream(p)) {
+                for (Path child : children) {
+                    delete(child);
+                }
+            }
+        }
+        try {
+            Files.deleteIfExists(p);
```
jglick, line 135:
```
[ERROR] hudson.LauncherTest.remotable  Time elapsed: 4.281 s  <<< ERROR!
java.nio.file.FileSystemException: C:\Jenkins\workspace\Core_jenkins_PR-8006\test\target\j h12573638483276859145\log: The process cannot access the file because it is being used by another process
	at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:92)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
	at java.base/sun.nio.fs.WindowsFileSystemProvider.implDelete(WindowsFileSystemProvider.java:275)
	at java.base/sun.nio.fs.AbstractFileSystemProvider.deleteIfExists(AbstractFileSystemProvider.java:110)
	at java.base/java.nio.file.Files.deleteIfExists(Files.java:1191)
	at org.jvnet.hudson.test.TemporaryDirectoryAllocator.delete(TemporaryDirectoryAllocator.java:135)
	at org.jvnet.hudson.test.TemporaryDirectoryAllocator.delete(TemporaryDirectoryAllocator.java:130)
	at org.jvnet.hudson.test.TemporaryDirectoryAllocator.dispose(TemporaryDirectoryAllocator.java:98)
	at org.jvnet.hudson.test.TestEnvironment.dispose(TestEnvironment.java:84)
	at …
```

jglick:
Tested successfully in https://github.com/jenkinsci/jenkins/pull/8006 which I think is good enough.

src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
```diff
@@ -116,13 +110,34 @@ public synchronized void disposeAsync() {
...
+        } catch (DirectoryNotEmptyException x) {
+            try (DirectoryStream<Path> children = Files.newDirectoryStream(p)) {
+                throw new IOException(Iterables.toString(children), x);
+            }
```
jtnord, line 139: NIT:  could we replace guava by calling Files.list(p) (which is a `java.util.Stream` and collecting the results?

jtnord (approved):

basil:
Caused #591.

basil:
Caused [JENKINS-71406](https://issues.jenkins.io/browse/JENKINS-71406).

slonopotamus:
Well, if this change exposes bugs in code, it's not this change fault that other code is buggy.

MarkEWaite:
> Well, if this change exposes bugs in code, it's not this change fault that other code is buggy.

I agree that it is not the fault of this change that other code has bugs.  However, it seems important to the overall progress of the project that when general behavior (tests passing reliably) regresses, it is better to either fix the problems that have been exposed by the change or revert the change that exposed the problems.  Remaining for an extended period in a state where tests are flaky (the problem is exposed but not resolved) will reduce our ability to trust the tests and our ability to add new capabilities.

jglick:
Commented in Jira. Please keep discussion there.

basil:
Caused #612.

basil:
@jglick This PR is causing problems in https://github.com/jenkinsci/copyartifact-plugin

jglick:
https://github.com/jenkinsci/copyartifact-plugin/pull/205 :eyes:

basil:
@jglick This PR is causing problems in https://github.com/jenkinsci/workflow-cps-global-lib-http-plugin/pull/167

PR #200, opened 2020-02-10
[JENKINS-61031] Fixed download of SVG images

fqueiruga:
[JENKINS-61031](https://issues.jenkins-ci.org/browse/JENKINS-61031)

- JenkinsRule#assertAllImageLoadSuccessfully no longer breaks when download images on the format ``
- Refactored and added a JenkinsRule#assertImageLoadsSuccessfully to implement the same test but for single images

#### Desired reviewers
@timja 
@fcojfernandez 

### Explanation

I have change the way of testing whether images were download to just check for the status code. That way it's format-agnostic.

I appreciate suggestions for how to add unit tests for this one.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
+++ b/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
@@ -174,6 +174,8 @@ import java.net.HttpURLConnection;
 
 import jenkins.model.JenkinsLocationConfiguration;
 
+import static org.junit.Assert.assertEquals;
+
 /**
  * Base class for all Jenkins test cases.
  *
@@ -950,17 +952,23 @@ public abstract class HudsonTestCase extends TestCase implements RootAction {
         assertTrue("needle found in haystack", found); 
     }
 
+    public void assertImageLoadsSuccessfully(HtmlImage img) {
+        try {
+            assertEquals("Failed to load " + img.getSrcAttribute(),
+                    200,
+                    img.getWebResponse(true).getStatusCode());
+        } catch (IOException e) {
+            throw new Error("Failed to load " + img.getSrcAttribute());
+        }
+    }
+
     /**
      * Makes sure that all the images in the page loads successfully.
      * (By default, HtmlUnit doesn't load images.)
      */
     public void assertAllImageLoadSuccessfully(HtmlPage p) {
         for (HtmlImage img : DomNodeUtil.<HtmlImage>selectNodes(p, "//IMG")) {
-            try {
-                img.getHeight();
-            } catch (IOException e) {
-                throw new Error("Failed to load "+img.getSrcAttribute(),e);
-            }
+            assertImageLoadsSuccessfully(img);
         }
     }
 
--- a/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
+++ b/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
@@ -952,15 +952,6 @@ public abstract class HudsonTestCase extends TestCase implements RootAction {
         assertTrue("needle found in haystack", found); 
     }
 
-    public void assertImageLoadsSuccessfully(HtmlImage img) {
-        try {
-            assertEquals("Failed to load " + img.getSrcAttribute(),
-                    200,
-                    img.getWebResponse(true).getStatusCode());
-        } catch (IOException e) {
-            throw new AssertionError("Failed to load " + img.getSrcAttribute());
-        }
-    }
 
     /**
      * Makes sure that all the images in the page loads successfully.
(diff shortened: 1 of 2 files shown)


src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -1506,17 +1506,23 @@ public void assertXPathResultsContainText(DomNode page, String xpath, String nee
         assertThat("needle found in haystack", found, is(true));
     }
 
+
+    public void assertImageLoadsSuccessfully(HtmlImage img) {
+        try {
+            assertEquals(img.getWebResponse(true).getStatusCode(), 200);
+        } catch (IOException e) {
+            throw new Error("Failed to load "+img.getSrcAttribute());
```
timja:
```suggestion
            throw new Error("Failed to load " + img.getSrcAttribute());
```

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -1506,17 +1506,25 @@ public void assertXPathResultsContainText(DomNode page, String xpath, String nee
         assertThat("needle found in haystack", found, is(true));
     }
 
+
+    public void assertImageLoadsSuccessfully(HtmlImage img) {
+        try {
+            assertEquals("Failed to load " + img.getSrcAttribute(),
```
varyvol: This should not happen in a helper method. This should say wether the image is loaded or not and the test will decide if that is an assertion error, which kind, etc.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -1506,17 +1506,25 @@ public void assertXPathResultsContainText(DomNode page, String xpath, String nee
...
+                    200,
+                    img.getWebResponse(true).getStatusCode());
+        } catch (IOException e) {
+            throw new Error("Failed to load " + img.getSrcAttribute());
```
timja:
```suggestion
            throw new AssertionError("Failed to load " + img.getSrcAttribute());
```

fqueiruga:
I just noticed that the HudsonTestCase class also implements this function, but I don't see it being used (https://github.com/jenkinsci/jenkins-test-harness/blob/3f7042162b31fe3a26ab883788d5ea77544a2942/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java#L957).

I assume I should go ahead and change that one right?

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
fqueiruga: I extracted it not as a helper method but as a custom assertion on it's own for downloading single images, not all within a page.
If that's not OK I'd rather just inline the changes on the other function, as it was previously

timja:
> I just noticed that the HudsonTestCase class also implements this function, but I don't see it being used (
> 
> https://github.com/jenkinsci/jenkins-test-harness/blob/3f7042162b31fe3a26ab883788d5ea77544a2942/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java#L957
> 
> ).
> I assume I should go ahead and change that one right?

yeah should do just in case

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
varyvol: I'm not sure. In my opinion the test should be the one deciding what happens if the behaviour is not the expected one. For example something as simple as changing the logged message when the assertion fails.

fcojfernandez: if you see the rest of the class, it's the way to proceed. There are plenty of methods `assertXYZ` with `assertTrue/That/Equals` within.

src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
```diff
@@ -950,17 +952,23 @@ public void assertXPathResultsContainText(DomNode page, String xpath, String nee
         assertTrue("needle found in haystack", found); 
     }
 
+    public void assertImageLoadsSuccessfully(HtmlImage img) {
+        try {
+            assertEquals("Failed to load " + img.getSrcAttribute(),
+                    200,
+                    img.getWebResponse(true).getStatusCode());
+        } catch (IOException e) {
+            throw new Error("Failed to load " + img.getSrcAttribute());
```
timja:
```suggestion
            throw new AssertionError("Failed to load " + img.getSrcAttribute());
```

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
timja: unless you have a need for a new API I would probably inline it or make this method private,

the assert part of the method means that an assertion / other exception is expected to be thrown if the condition is not met

fqueiruga: I have inlined the assertion code into the old methods

fqueiruga: Resolving this one. Please open if you would like further changes

timja:
@jenkinsci/core could someone take a look at this please? it'll unblock https://github.com/jenkinsci/jenkins/pull/4463

MarkEWaite:
Unless there are objections, I will merge this tomorrow.

fqueiruga:
Thank you Mark. Any chance for a new release with this change?

oleg-nenashev:
I will try to cut a release tonight if nobody else does it before me

PR #201, opened 2020-02-12
[JENKINS-60410] Disable stack trace suppression for tests.

jeffret-b:
If this property is set and if the changes in jenkinscis/jenkins#4389 are active, then this changes the default behavior to show full stack traces. (It enables stack traces.) If the property isn't set with 4389 then stack traces are suppressed and some tests that check error results fail. If the property is set without 4389 then it's a no-op.

This matches other existing behavior. See the nearby line for allowing Dispatcher tracing, `Dispatcher.TRACE`.

It makes testing and troubleshooting a little easier.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -2682,6 +2682,7 @@ public class JenkinsRule implements TestRule, MethodRule, RootAction {
         {// enable debug assistance, since tests are often run from IDE
             Dispatcher.TRACE = true;
             MetaClass.NO_CACHE=true;
+            System.setProperty("jenkins.model.Jenkins.SHOW_STACK_TRACE", "true");
             // load resources from the source dir.
             File dir = new File("src/main/resources");
             if(dir.exists() && MetaClassLoader.debugLoader==null)


jeffret-b:
Hmm ... looks like maybe incrementals publishing is broken again.

jeffret-b:
Let's see if incrementals is working now.

timja:
> Let's see if incrementals is working now.

I've merged master as it was skipped in the last build due to not being up to date

jeffret-b:
Thanks for that @timja. I forgot about checking if master was up-to-date.

jeffret-b:
Putting this on-hold for some discussion on the disabling mechanism happening over in the companion core PR. Once I get that addressed, I'll update this one.

jeffret-b:
Removing the on-hold label as the concerns, especially relating to the other PR in core, have been accepted. This should now be ready for a release so the core PR can proceed.

oleg-nenashev:
https://github.com/jenkinsci/jenkins/pull/4389 needs to be merged/released first

jeffret-b:
@oleg-nenashev , since this just adds a property, this one isn't dependent on the change in core. It just doesn't do anything until this release is integrated with that change. This one should be good to go now.

I'm double-checking whether this change in jenkins-test-harness is required for getting the tests to all pass for the PR in core.

jeffret-b:
We will keep this on hold, until I verify the state with the other PR, though.

jeffret-b:
This one is ready to go and the on-hold can be removed. As explained, this is not dependent on the core change. Instead, it enables the tests to pass for the core change.

It would be great to get this merged and released so the core PR can move forward.

PR #206, opened 2020-04-02
JenkinsRule.waitOnline should stream Computer.logText

jglick:
Helpful for diagnosing while a launch is hanging, and just nice to know even when it succeeds.


oleg-nenashev:
Merge conflict

PR #207, opened 2020-04-04
Make this compile again with a Jenkins 2.61+ dependency

daniel-beck:
Looks like https://github.com/jenkinsci/jenkins/pull/2864 and https://github.com/jenkinsci/jenkins-test-harness/pull/167 don't work well together.

Untested.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -85,7 +85,7 @@ THE SOFTWARE.
     <dependency>
       <groupId>org.jenkins-ci.main</groupId>
       <artifactId>jenkins-war</artifactId>
-      <version>2.60.3</version>
+      <version>2.73.3</version>
       <type>executable-war</type>
       <exclusions>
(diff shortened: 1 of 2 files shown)


slonopotamus:
I came up with exactly the same changes locally. `ParameterizedJobMixIn.ParameterizedJob` got generics in https://github.com/jenkinsci/jenkins/commit/37dfa992eb1a6aba5e25b791703001f7516d5b63 so we need to account for that.

daniel-beck:
@oleg-nenashev It's a compile error, I doubt it's a problem at run time or we'd have noticed.

daniel-beck:
Changing misleading title to a better one.

This might not make sense standalone, but might be useful to keep around if https://github.com/jenkinsci/jenkins-test-harness/pull/133 doesn't get merged soon.

PR #209, opened 2020-04-09
Add WithTimeout support for JUnit4

Wadeck:
Useful during debugging to avoid writing:

```
{j.timeout = 0;}
```

Nothing depends on this improvement :)

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/recipes/WithTimeout.java
+++ b/src/main/java/org/jvnet/hudson/test/recipes/WithTimeout.java
@@ -24,6 +24,8 @@
 package org.jvnet.hudson.test.recipes;
 
 import org.jvnet.hudson.test.HudsonTestCase;
+import org.jvnet.hudson.test.JenkinsRecipe;
+import org.jvnet.hudson.test.JenkinsRule;
 
 import java.lang.annotation.Documented;
 import static java.lang.annotation.ElementType.METHOD;
@@ -41,6 +43,7 @@ import java.lang.annotation.Target;
  */
 @Documented
 @Recipe(WithTimeout.RunnerImpl.class)
+@JenkinsRecipe(WithTimeout.RunnerImpl4.class)
 @Target(METHOD)
 @Retention(RUNTIME)
 public @interface WithTimeout {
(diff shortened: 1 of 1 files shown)


MarkEWaite:
Could the same thing be achieved with the [DisableOnDebug rule](https://junit.org/junit4/javadoc/4.13/org/junit/rules/DisableOnDebug.html)?

Wadeck:
> Could the same thing be achieved with the DisableOnDebug rule?

Never used it :) Dunno if there is a link between the timeout from JenkinsRule and the "real" timeout of JUnit.

UltimateGeek:
@Wadeck I tried using the 2.64 released version of jenkins-test-harness with this change, but a test I'm adjusting the timeout settings for is still is timing out after 180 seconds. Can you tell me what I'm doing wrong?

> 180.006 [id=11]    WARNING o.j.hudson.test.JenkinsRule$2#evaluate: Test timed out (after 180 seconds).

test annotations are set to 30min for JenkinsTest and 1 minute for junit test (so I can confirm quickly the junit4 timeout setting is getting applied):
```
    @Test(timeout = 1800000L)
    @WithTimeout(1)
```

relevant dependencies in build.gradle:
```
plugins {
    ....
    id 'com.mkobit.jenkins.pipelines.shared-library' version '0.10.1'
}
...
dependencies {
    ...
    testImplementation 'junit:junit:4.12'
    testImplementation 'org.jenkins-ci.main:jenkins-test-harness:2.64'
}
```

Wadeck:
@MarkEWaite I looked at the code and tested it manually, the DisableOnDebug is already used in [JenkinsRule.java#L326|https://github.com/jenkinsci/jenkins-test-harness/blob/e944e0fca5cb16b28d2e2a3aee64aee30c924db4/src/main/java/org/jvnet/hudson/test/JenkinsRule.java#L326] and so, if you are debugging a test, it's automatically used. _(I imagine that I was not aware of this feature because of work on very old line of Jenkins without that integration ^^)_

@UltimateGeek Actually this PR does not bring any value as I discovered the WithTimeout was already supported, but directly from JenkinsRule. I suspect that your test is not using `JenkinsRule` otherwise it should work. If you can copy-paste a larger snippet, it could be easier.
(Also note, that GitHub is not the place to ask question, better looking at IRC, https://www.jenkins.io/chat/)

Revert of this PR, proposed in #220

UltimateGeek:
Thank you @Wadeck. Now I understand by what you meant by `{j.timeout = 0;}`. I think the documentation about the public field [JenkinsRule.timeout](https://github.com/jenkinsci/jenkins-test-harness/blob/master/src/main/java/org/jvnet/hudson/test/JenkinsRule.java#L324) could have its documentation improved to make a point that that is a global timeout, and one has to change it if using `@WithTimeout` and wanting to have it be longer than the default.

PR #214, opened 2020-04-27
add the ability to disable a plugin in RestartableJenkinsRule

jtnord:
it was not easy to write a PluginManager that loaded dependencies
correctly for Jars for use in RestartableJenkinsRule that supported
plugin disabling.

With this change UnitTestSupportingPluginManager works just like a
normal PluginManager *and* support the logic needed for maven test scope
dependencies (but without the run tests as fast as possible used by
TestPluginManager)

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/TestPluginManager.java
+++ b/src/main/java/org/jvnet/hudson/test/TestPluginManager.java
@@ -1,192 +1,28 @@
-/*
- * The MIT License
- *
- * Copyright (c) 2010, Kohsuke Kawaguchi
- *
- * Permission is hereby granted, free of charge, to any person obtaining a copy
- * of this software and associated documentation files (the "Software"), to deal
- * in the Software without restriction, including without limitation the rights
- * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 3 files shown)


src/main/java/org/jvnet/hudson/test/TestPluginManager.java
```diff
@@ -1,175 +1,56 @@
...
 
-    public TestPluginManager() throws IOException {
-        // TestPluginManager outlives a Jetty server, so can't pass in ServletContext.
-        super(null, Util.createTempDir());
-    }
+    public static final PluginManager INSTANCE;
 
-    /** @see LocalPluginManager#loadBundledPlugins */
-    @Override
-    protected Collection<String> loadBundledPlugins() throws Exception {
```
jglick, line 72: Unfortunately Git provides no way to rename a file and then create a distinct file with the original name in such a way as to make for a reviewable diff. So I have no idea what this PR is doing, and unfortunately we cannot rename `TestPluginManager` since there are a few external references to it.

jtnord: see #215 which has a separate commit for the rename and code changes.

PR #215, opened 2020-04-27
Add support for unit tests that want to disable plugins

jtnord:
it was not easy to write a PluginManager that loaded dependencies
correctly for Jars for use in RestartableJenkinsRule that supported
plugin disabling.

With this change UnitTestSupportingPluginManager works just like a
normal PluginManager and support the logic needed for maven test scope
dependencies (but without the run tests as fast as possible used by
TestPluginManager)

#214 with the rename as a separate commit


src/test/java/org/jvnet/hudson/test/RestartableJenkinsRuleTest.java
```diff
@@ -52,4 +74,35 @@ public void testPortReuse() throws Exception {
...
+        noPortReuse.then(jr -> {
+            Jenkins j = jr.jenkins;
+            PluginManager pm = j.getPluginManager();
+            PluginWrapper plugin = pm.getPlugin(pluginId);
+            assertNotNull("plugin should not be null", plugin);
+            plugin.disable();
+        });
+
+        noPortReuse.then(jr -> {
+            assertFalse("workflow-job is not enabled",
```
jglick, line 102:
```suggestion
            assertFalse(pluginId + " is not enabled",
```

src/test/java/org/jvnet/hudson/test/RestartableJenkinsRuleTest.java
```diff
@@ -52,4 +74,35 @@ public void testPortReuse() throws Exception {
...
+                        jr.jenkins.getPluginManager().getPlugin(pluginId).isEnabled());
+            assertFalse("workflow-job should not be active",
```
jglick, line 104:
```suggestion
            assertFalse(pluginId + " should not be active",
```

src/test/java/org/jvnet/hudson/test/RestartableJenkinsRuleTest.java
```diff
@@ -52,4 +74,35 @@ public void testPortReuse() throws Exception {
                     assertEquals(port.get(), portReuse.j.getURL().getPort());
                 });
     }
+
+    @Test
+    @WithPluginManager(UnitTestSupportingPluginManager.class)
```
jglick, line 79: Maybe Javadoc of `RestartableJenkinsRule` should note this idiom?

src/main/java/org/jvnet/hudson/test/TestPluginManager.java
```diff
@@ -0,0 +1,70 @@
...
+    // Releasing plugins result fail to access files in webapp directory in following tests.
+    @Override
+    public void stop() {
+        for (PluginWrapper p : activePlugins)
+            p.stop();
+    }
+
+    /**
+     * As we don't actually shut down classloaders, we instead provide this method that does
+     * what {@link #stop()} normally does.
```
jglick:
```suggestion
     * what {@link PluginManager#stop()} normally does.
```

since it is not obvious to readers what `super` is here

src/test/java/org/jvnet/hudson/test/RestartableJenkinsRuleTest.java
jtnord: happy to do that as a follow-up if it means I can get a release sooner

src/test/java/org/jvnet/hudson/test/RestartableJenkinsRuleTest.java
```diff
@@ -52,4 +74,35 @@ public void testPortReuse() throws Exception {
...
+        final String pluginId = "display-url-api";
+        noPortReuse.then(jr -> {
+            System.out.println(WarExploder.getExplodedDir());
+            Path srcLdap = new File(WarExploder.getExplodedDir(), "WEB-INF/detached-plugins/"+pluginId+".hpi").toPath();
+            Path dstLdap = new File(jr.jenkins.pluginManager.rootDir, pluginId+".jpi").toPath();
+            Files.createDirectories(dstLdap.getParent());
+            Files.copy(srcLdap, dstLdap);
+            //jr.getPluginManager().doCheckUpdatesServer();
+            //jr.getPluginManager().install(Collections.singletonList("cvs"),true);
+            System.out.println("hello");
```
jtnord, line 90:
```suggestion
```

PR #220, opened 2020-05-25
Revert the change with WithTimeout

Wadeck:
It was already implemented directly in JenkinsRule, private method: getTestTimeoutOverride
I updated a test case that was not asserting something useful.

Revert #209 + update test.

@jglick @MarkEWaite as reviewers of the first one

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/recipes/WithTimeout.java
+++ b/src/main/java/org/jvnet/hudson/test/recipes/WithTimeout.java
@@ -37,14 +37,13 @@ import java.lang.annotation.Target;
  *
  * All the tests time out after some number of seconds by default, but this recipe
  * allows you to override that value.
- * 
- * No need for @JenkinsRecipe in JUnit 4 as it's implemented directly in {@link JenkinsRule} 
- * by the private method: getTestTimeoutOverride.
  *
  * @author Kohsuke Kawaguchi
  */
 @Documented
 @Recipe(WithTimeout.RunnerImpl.class)
+// No need for @JenkinsRecipe in JUnit 4 as it's implemented directly in JenkinsRule 
+// by the private method: getTestTimeoutOverride.
 @Target(METHOD)
 @Retention(RUNTIME)
 public @interface WithTimeout {
--- a/src/main/java/org/jvnet/hudson/test/recipes/WithTimeout.java
+++ b/src/main/java/org/jvnet/hudson/test/recipes/WithTimeout.java
@@ -24,7 +24,6 @@
 package org.jvnet.hudson.test.recipes;
 
 import org.jvnet.hudson.test.HudsonTestCase;
-import org.jvnet.hudson.test.JenkinsRecipe;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jvnet/hudson/test/recipes/WithTimeout.java
```diff
@@ -38,12 +37,14 @@
  *
  * All the tests time out after some number of seconds by default, but this recipe
  * allows you to override that value.
+ * 
+ * No need for @JenkinsRecipe in JUnit 4 as it's implemented directly in {@link JenkinsRule} 
+ * by the private method: getTestTimeoutOverride.
```
UltimateGeek, line 42: Consider referencing the public `JenkinsRule.timeout method()` as what a user can call, instead of referencing a private method here in public documentation.

Wadeck: Good point. Actually the comment was not meant to be read by "user" but more "developer", to prevent future myself to re-do the same useless change :) So I move the comment outside of the JavaDoc.

src/main/java/org/jvnet/hudson/test/recipes/WithTimeout.java
```diff
@@ -62,17 +62,8 @@
     class RunnerImpl extends Recipe.Runner<WithTimeout> {
         @Override
         public void setup(HudsonTestCase testCase, WithTimeout recipe) throws Exception {
+            System.out.println("RunnerImpl call");
```
jglick, line 65:
```suggestion
```

src/test/java/org/jvnet/hudson/main/JenkinsRuleTimeoutTest.java
```diff
@@ -156,9 +156,10 @@ public void run() {
         }
     }
 
-    @Test @WithTimeout(20)
+    @Test 
+    @WithTimeout(15)
     public void withTimeoutPropagation() throws Exception {
-        Thread.sleep(1000 * 30);
+        Thread.sleep(1000 * 15);
```
jglick, line 162: Not following how this could be anything but flaky. You are setting a 15s timeout on the test and then sleeping for 15s, so the test will sometimes get the timeout (and pass, due to `thrown.expect(TestTimedOutException.class)`?) and sometimes not get the timeout and so fail with the assertion?

jglick:
Also docs change suggested in https://github.com/jenkinsci/jenkins-test-harness/pull/209#issuecomment-633649673

src/test/java/org/jvnet/hudson/main/JenkinsRuleTimeoutTest.java
Wadeck: My concern with this test is that if you remove the annotation it still passes and so, does not test what it's expected to test.

I was too optimistic about the guarantee from sleep, moving to 20s.

PR #226, opened 2020-07-07
Added JenkinsSessionRule

jglick:
This has been on my to-do list for years. There is a bunch of complex stuff in `RestartableJenkinsRule` that is rarely needed, like support for dirty shutdowns (used only for testing some edge cases in Pipeline) and custom port selection (used by the `kubernetes` plugin IIRC), but the immediate motivation was to run `then` blocks synchronously. The batching behavior in `RestartableJenkinsRule` is rather counterintuitive, seems unnecessary, and mystifyingly breaks basic JUnit stuff like `@After` methods.

Also this variant preserves the port across runs. All in all, should be a drop-in replacement for most uses.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsSessionRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsSessionRule.java
@@ -1,0 +1,115 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2020 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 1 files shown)


jglick:
> Actually this is used by Swarm in `PipelineJobRestartTest`

Only via `withReusedPort`, apparently. But reusing the port should just be automatic—you should not have to request it.

I may have been thinking of https://github.com/jenkinsci/kubernetes-plugin/blob/d9cf6a810fce807fe35cd56ebe17ffb343c702f4/src/test/java/org/jvnet/hudson/test/RestartableJenkinsNonLocalhostRule.java#L42 + https://github.com/jenkinsci/kubernetes-plugin/blob/d9cf6a810fce807fe35cd56ebe17ffb343c702f4/src/test/java/org/csanchez/jenkins/plugins/kubernetes/pipeline/RestartPipelineTest.java#L79 which is nasty and maybe it did not need to hard-code a port. More broadly, https://github.com/jenkinsci/kubernetes-plugin/blob/d9cf6a810fce807fe35cd56ebe17ffb343c702f4/src/test/java/org/jvnet/hudson/test/JenkinsRuleNonLocalhost.java#L52 is there because of some hacks https://github.com/jenkinsci/kubernetes-plugin/blob/d9cf6a810fce807fe35cd56ebe17ffb343c702f4/pom.xml#L296-L316 which are not necessary if you https://github.com/jenkinsci/kubernetes-plugin/blob/d9cf6a810fce807fe35cd56ebe17ffb343c702f4/test-in-k8s.sh#L22-L32 but are handy for using an IDE locally.

basil:
> But reusing the port should just be automatic—you should not have to request it.

I don't have a strong opinion either way. When I implemented #132 I wrote:

>The new functionality is opt-in to preserve existing behavior for tests which don't need it and which could be negatively impacted by races for open ports.

But you have more experience with the Jenkins plugin ecosystem, so you are in a better position to judge whether the risk is acceptable.

I think Swarm is the only plugin that uses `withReusedPort`, so once this is PR is released and I convert Swarm to using `JenkinsSessionRule`, we could just remove the (by then) dead `withReusedPort` functionality from `RestartableJenkinsRule`.

jglick:
There _is_ a theoretical race condition for an open server port in between Jenkins (and thus Jetty) restarts. Still, the chances of some other process grabbing the port (out of ~32k) in that short time window seems slim. And most CI test runs these days are done either within a VM or a container with an isolated port namespace. A tolerable risk given the convenience of having a stable `rootUrl` across sessions.

PR #242, opened 2020-09-21
Added FlagRule

jglick:
A lot of Jenkins tests turn on some hidden flag defined by static variable, so it would be convenient to not need `@Before` & `@After` methods for this common idiom.

### DIFF
--- a/src/test/java/org/jvnet/hudson/test/FlagRuleNoReplacementTest.java
+++ b/src/test/java/org/jvnet/hudson/test/FlagRuleNoReplacementTest.java
@@ -1,0 +1,57 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2020 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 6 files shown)


oleg-nenashev (approved):
LGTM. It might make sense to also add a convenience `SystemPropertyRule`, but it can be done later

jglick:
Yeah I was thinking of adding a factory method for system properties specifically. OTOH _most_ places in Jenkins use a `static` field initialized from the system property so I am not sure it would get used very much.

I need to find a good demonstration usage of this API in something I own, unlike https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/405.

src/test/java/org/jvnet/hudson/test/FlagRuleNoReplacementTest.java
```diff
@@ -0,0 +1,57 @@
...
+        assertFalse(FLAG);
+    }
+
+    @AfterClass
+    public static void end() {
+        assertFalse(FLAG);
+    }
+
+    @Rule
+    public TestRule flagRule = new FlagRule<Boolean>(() -> FLAG, x -> FLAG = x);
```
jglick, line 49: As per [this](https://stackoverflow.com/a/27472025/12916) there is no slick syntax for representing field accessors. There _could_ be a factory like

```java
@Rule
public TestRule flagRule = FlagRule.field(ThisTest.class, "FLAG");
```

which would look up the field reflectively, but then this is opaque to IDEs looking to find usages etc., which does not seem to justify the slightly shorter syntax.

oleg-nenashev (approved):
Great stuff!

PR #246, opened 2020-10-10
[Hacktoberfest] Code cleanup

basil:
A series of code cleanup changes for this repository. Most of these changes remove usages of deprecated functionality or otherwise fix IDE warnings. I kept each commit separate. Each commit message should be self-explanatory, but if anything is unclear I am happy to provide further explanation. Also, feel free to revert any specific cleanup commits that are unwanted.

### DIFF
--- a/src/main/java/com/gargoylesoftware/htmlunit/WebResponseListener.java
+++ b/src/main/java/com/gargoylesoftware/htmlunit/WebResponseListener.java
@@ -16,7 +16,7 @@ public interface WebResponseListener {
     final class StatusListener implements WebResponseListener {
 
         private final int statusCode;
-        private final List<WebResponse> responses = new CopyOnWriteArrayList<WebResponse>();
+        private final List<WebResponse> responses = new CopyOnWriteArrayList<>();
 
         public StatusListener(final int statusCode) {
             this.statusCode = statusCode;
--- a/src/main/java/com/gargoylesoftware/htmlunit/WebResponseListener.java
+++ b/src/main/java/com/gargoylesoftware/htmlunit/WebResponseListener.java
@@ -30,7 +30,7 @@ public interface WebResponseListener {
         }
 
         public void assertHasResponses() {
-            Assert.assertTrue(!responses.isEmpty());
+            Assert.assertFalse(responses.isEmpty());
         }
 
         public List<WebResponse> getResponses() {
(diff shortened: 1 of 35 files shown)


src/main/java/com/gargoylesoftware/htmlunit/WebResponseListener.java
```diff
@@ -30,7 +30,7 @@ public void onLoadWebResponse(WebRequest webRequest, WebResponse webResponse) th
         }
 
         public void assertHasResponses() {
-            Assert.assertTrue(!responses.isEmpty());
+            Assert.assertFalse(responses.isEmpty());
```
jglick, line 33: Or a more helpful `Matcher`.

src/main/java/org/jvnet/hudson/test/DefaultConstructorChecker.java
```diff
@@ -19,9 +19,7 @@ public DefaultConstructorChecker(Class clazz) {
     protected void runTest() throws Throwable {
         try {
             clazz.getConstructor();
-        } catch (NoSuchMethodException e) {
-            throw new Error(clazz+" must have the default constructor",e);
-        } catch (SecurityException e) {
+        } catch (NoSuchMethodException | SecurityException e) {
```
jglick, line 22:
```suggestion
        } catch (NoSuchMethodException e) {
```

Not sure what that was there for, should never be thrown.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -340,7 +340,7 @@
 
     private boolean origDefaultUseCache = true;
 
-    private static final Charset UTF8 = Charset.forName("UTF-8");
+    private static final Charset UTF8 = StandardCharsets.UTF_8;
```
jglick, line 343: just inline

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -2439,7 +2431,7 @@ public WebClient login(String username) throws Exception {
          */
         public <V> V executeOnServer(final Callable<V> c) throws Exception {
             final Exception[] t = new Exception[1];
-            final List<V> r = new ArrayList<V>(1);  // size 1 list
+            final List<V> r = new ArrayList<>(1);  // size 1 list
```
jglick, line 2434: `AtomicReference` is clearer BTW.

jglick (approved):

jglick:
Merging as is though @basil I left a few suggestions for further cleanups if you have the inclination.

PR #247, opened 2020-10-12
Cleanup 'Hudson' references in jenkins-test-harness

darkphoenix17:
Replacement of **Hudson** by **Jenkins**
Assigned by @oleg-nenashev 
let me know for further changes to be made or if there is an error in the changes.


MarkEWaite:
I'm not understanding the intent of this pull request.  The HudsonTestCase class is [deprecated](https://javadoc.jenkins.io/component/jenkins-test-harness/org/jvnet/hudson/test/HudsonTestCase.html) but still exists so that code tests relying on HudsonTestCase continue to operate.  This seems to rename the deprecated class from HudsonTestCase to JenkinsTestCase but leaves it deprecated.

That seems like it will break old code while it won't encourage developers using the old code to do anything more than text replacement of HudsonTestCase with JenkinsTestCase.  It seems like the desired end state is that tests using HudsonTestCase should switch to JenkinsRule, not just a rename.

Can you help me understand what I've missed in the intent of this change?

darkphoenix17:
I had made a similar kind of patch where i cleaned up the **Hudson** reference  for this [issue](https://github.com/jenkinsci/jenkinsfile-runner/pull/378). I thought a similar patch has to be made here as it was mentioned in the comment section there.  
I suppose i made a misunderstood it. Could you guide me throught the desired state which you are talking about. 
Also i am new to open-source a steady and stepful guidance would be helpful.

MarkEWaite:
> I had made a similar kind of patch where i cleaned up the **Hudson** reference for this [issue](https://github.com/jenkinsci/jenkinsfile-runner/pull/378). I thought a similar patch has to be made here as it was mentioned in the comment section there.
> I suppose i misunderstood it. Could you guide me through the desired state which you are talking about.
> Also i am new to open-source so a steady and stepful guidance would be helpful.

Thanks for the clarification.  That helped me understand the context.  Thanks also for your interest in open source and your willingness to contribute.  We're thrilled to have your help!

Oleg said in [his comment](https://github.com/jenkinsci/jenkinsfile-runner/pull/378#issuecomment-703454607) that binary compatibility needs to be retained.  In order to retain binary compatibility, we can't rename classes.  When a class is renamed, the resulting class is no longer binary compatible.  Code compiled with the old class will fail with a class not found exception when running with the new class.

We can replace text in comments and retain binary compatibility.  Thus, comments that refer to 'Hudson' can usually be changed to refer to 'Jenkins' with no loss of clarity and with binary compatibility.

We can add new classes and retain binary compatibility.  I'm not sure that will be needed in this case, but possibly.

.factorypath
```diff
@@ -0,0 +1,163 @@
+<factorypath>
```
MarkEWaite, line 1: I'm not accustomed to seeing [Eclipse specific configuration files](https://stackoverflow.com/questions/42949198/automatically-generate-factorypath-on-project-import-when-using-maven-project-i) added to the Jenkins repositories.  I may be mistaken on that, but I don't recall that being common.

Unless others guide differently, I would not include files that are specific to Eclipse IDE in the pull request.

src/main/java/com/gargoylesoftware/htmlunit/README.md
```diff
@@ -1,5 +1,5 @@
 # Overview
-Jenkins used to maintain a forked version of HtmlUnit and use it via `JenkinsRule` (and `HudsonTestCase`). Moving
+Jenkins used to maintain a forked version of HtmlUnit and use it via `JenkinsRule` (and `JenkinsTestCase`). Moving
```
MarkEWaite, line 2: Should not rename the HudsonTestCase class, so should not change this documentation in this way
```suggestion
Jenkins used to maintain a forked version of HtmlUnit and use it via `JenkinsRule` (and `HudsonTestCase`). Moving
```

src/main/java/org/jvnet/hudson/test/ClosureExecuterAction.java
```diff
@@ -35,7 +35,7 @@
 import java.util.UUID;
 
 /**
- * Server-side logic that implements {@link HudsonTestCase#executeOnServer(java.util.concurrent.Callable)}.
+ * Server-side logic that implements {@link JenkinsTestCase#executeOnServer(java.util.concurrent.Callable)}.
```
MarkEWaite, line 38: Since HudsonTestCase is deprecated but needed for binary compatibility, we should not rename the class.
```suggestion
 * Server-side logic that implements {@link HudsonTestCase#executeOnServer(java.util.concurrent.Callable)}.
```

src/main/java/org/jvnet/hudson/test/JenkinsHomeLoader.java
```diff
@@ -36,11 +36,11 @@
 import javax.annotation.CheckForNull;
 
 /**
- * Controls how a {@link HudsonTestCase} initializes <tt>JENKINS_HOME</tt>.
+ * Controls how a {@link JenkinsTestCase} initializes <tt>JENKINS_HOME</tt>.
```
MarkEWaite, line 39: Since HudsonTestCase is deprecated, it should not be renamed.

```suggestion
 * Controls how a {@link HudsonTestCase} initializes <tt>JENKINS_HOME</tt>.
```

darkphoenix17:
> Thanks for the clarification. That helped me understand the context. Thanks also for your interest in open source and your willingness to contribute. We're thrilled to have your help!
> 
> Oleg said in [his comment](https://github.com/jenkinsci/jenkinsfile-runner/pull/378#issuecomment-703454607) that binary compatibility needs to be retained. In order to retain binary compatibility, we can't rename classes. When a class is renamed, the resulting class is no longer binary compatible. Code compiled with the old class will fail with a class not found exception when running with the new class.
> 
> We can replace text in comments and retain binary compatibility. Thus, comments that refer to 'Hudson' can usually be changed to refer to 'Jenkins' with no loss of clarity and with binary compatibility.
> 
> We can add new classes and retain binary compatibility. I'm not sure that will be needed in this case, but possibly.

Thanks for making the binary compatitibilty concept clear. 
I got it that i shouldn't have renamed the class. 
So i am not sure what are the patches to be made in the codebase heres specifically?
Can you put down some fixes in bullet points so that i get a starting point from where i need to work upon?

MarkEWaite (changes requested):
Need to remove the method renames since they break binary compatibility

MarkEWaite:
> Can you put down some fixes in bullet points so that i get a starting point from where i need to work upon?

* Rename Hudson to Jenkins in comments where that rename is correct
* Rename method argument names that use `Hudson` as part of the argument name to use Jenkins instead (so long as references inside the method are also renamed, renaming the argument should not break compatibility.)

darkphoenix17:
Thanks Mark... I am on it!!!

src/main/java/jenkins/benchmark/jmh/JmhBenchmarkState.java
```diff
@@ -1,6 +1,6 @@
 package jenkins.benchmark.jmh;
 
-import hudson.model.Hudson;
+import hudson.model.*;       //syntax warning fix
```
jglick, line 3: Please do not use wildcard imports.

If you wish to suppress a deprecation warning, best to actually fix the usage, but if that is impossible, use a FQN rather than an `import` statement, and use

```java
@SuppressWarnings("deprecation")
```

in the smallest possible enclosing scope.

src/main/java/org/jvnet/hudson/test/JenkinsHomeLoader.java
```diff
@@ -36,11 +36,11 @@
...
  *
  * @author Kohsuke Kawaguchi
  */
-public interface HudsonHomeLoader {
+public interface JenkinsHomeLoader {
```
jglick, line 43: No, this is incompatible: https://github.com/search?q=user%3Ajenkinsci+HudsonHomeLoader&type=Code

jglick (changes requested):
I think this PR should be closed. It will break things and there is no reason to do it.

jglick:
https://github.com/jenkinsci/jenkins-test-harness/pull/247#issuecomment-706836357 would be fine, but that would basically be a fresh PR, nothing in common with this one.

darkphoenix17:
> 
> 
> > Can you put down some fixes in bullet points so that i get a starting point from where i need to work upon?
> 
>     * Rename Hudson to Jenkins in comments where that rename is correct
> 
>     * Rename method argument names that use `Hudson` as part of the argument name to use Jenkins instead (so long as references inside the method are also renamed, renaming the argument should not break compatibility.)

Thanks!! Now i can start a new PR and start afresh and make changes specified above.

PR #252, opened 2020-10-12
Code cleanup

basil:
Addressing review feedback from jenkinsci/jenkins-test-harness#246 plus more minor cleanup along the same lines.


MarkEWaite (approved):
Changes look reasonable to me.

PR #261, opened 2020-11-05
Auto-creation of API Token when using withBasicApiToken

Wadeck:
Proposal after discussion with Jesse.

Increase the core requirement to have at least the first LTS with the new API token system.

(Manually tested with 2.260 to ensure compatibility of the "else")

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -85,7 +85,7 @@ THE SOFTWARE.
     <dependency>
       <groupId>org.jenkins-ci.main</groupId>
       <artifactId>jenkins-war</artifactId>
-      <version>2.73.3</version>
+      <version>2.138.4</version>
       <type>executable-war</type>
       <exclusions>
         <exclusion>
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -111,6 +111,7 @@ import hudson.tools.ToolProperty;
 import hudson.util.PersistedList;
 import hudson.util.ReflectionUtils;
 import hudson.util.StreamTaskListener;
+import hudson.util.VersionNumber;
 import hudson.util.jna.GNUCLibrary;
 import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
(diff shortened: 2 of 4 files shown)


jglick:
Seems like it would be a lot easier just to turn on legacy mode, like tests in core are already doing?

jglick:
Or: on 2.260+, call `addFixedNewToken` (for now via reflection); otherwise turn on legacy mode and use the implied token. Would do the right thing (or a reasonable thing) going forward, while making this code a lot simpler (no need for the mock web response stuff).

Wadeck:
> Seems like it would be a lot easier just to turn on legacy mode, like tests in core are already doing?

The problem is the legacy mode will be enabled only when you are requested the withBasicApiToken, but the legacy behavior was to generate the API Token on user creation, which occurs in general at the beginning of the test.

jglick:
> the legacy behavior was to generate the API Token on user creation

So if you create a `User`, then turn on legacy behavior, then ask for the `UserProperty`, it is not there? Or the token is not honored?

Wadeck:
The ApiTokenProperty will be created but with a null apiToken (field). Leading to an error message returned from getApiToken() instead of the legacy API token.

If the legacy behavior is enabled, the field is generated and thus, contains the legacy API token.

basil:
FYI I had to deal with [a related issue](https://github.com/jenkinsci/swarm-plugin/blob/f4e4fdd4f8d7f33f47ba3510f3c01dddfbf0f186/plugin/src/test/java/hudson/plugins/swarm/test/GlobalSecurityConfigurationBuilder.java#L98-L107) when writing [`hudson.plugins.swarm.AuthorizationStrategyTest`](https://github.com/jenkinsci/swarm-plugin/blob/f4e4fdd4f8d7f33f47ba3510f3c01dddfbf0f186/plugin/src/test/java/hudson/plugins/swarm/AuthorizationStrategyTest.java).

jglick (approved):

timja:
Causes test breakages in Jenkins core:
https://github.com/jenkinsci/jenkins/pull/5227#issuecomment-770958751

PR #263, opened 2020-11-13
LoggerRule improvements

Vlatombe:
* Don't duplicate records with level `INFO` or above
* `quiet()` allows to record log messages but not print them to console. Useful if there are multiple LoggerRule instances.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/LoggerRule.java
+++ b/src/main/java/org/jvnet/hudson/test/LoggerRule.java
@@ -58,11 +58,12 @@ import org.junit.rules.RuleChain;
  */
 public class LoggerRule extends ExternalResource {
 
-    private final Handler consoleHandler = new ConsoleHandler();
+    private final Handler consoleHandler = new ConsoleHandlerWithMaxLevel();
     private final Map<Logger,Level> loggers = new HashMap<>();
-    // initialized iff capture is called:
+    // initialized if capture is called:
     private RingBufferLogHandler ringHandler;
     private List<String> messages;
+    private boolean verbose = true;
 
     /**
      * Initializes the rule, by default not recording anything.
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/LoggerRule.java
```diff
@@ -284,4 +297,22 @@ public void describeTo(org.hamcrest.Description description) {
         }
     }
 
+    /**
+     * Delegates to the given Handler but filter out records higher or equal to its initial level
```
jglick, line 301: Did you see #145 and specifically verify this behavior?

src/main/java/org/jvnet/hudson/test/LoggerRule.java
```diff
@@ -58,11 +58,12 @@
  */
 public class LoggerRule extends ExternalResource {
 
-    private final Handler consoleHandler = new ConsoleHandler();
+    private final Handler consoleHandler = new ConsoleHandlerWithMaxLevel();
     private final Map<Logger,Level> loggers = new HashMap<>();
-    // initialized iff capture is called:
+    // initialized if capture is called:
```
jglick, line 63: Math abbrev.

```suggestion
    // initialized if and only if capture is called:
```

Vlatombe: Ah, I didn't get it.

src/main/java/org/jvnet/hudson/test/LoggerRule.java
Vlatombe: I didn't see it before, but afaics all the FINE logs were ok when I verified it.

jglick: Great, just verifying that you tested more carefully than I did in #138. :disappointed:

jglick (approved):

jglick:
https://github.com/jenkinsci/jenkins-test-harness/releases/tag/jenkins-test-harness-2.71

PR #264, opened 2020-11-23
Ensure JenkinsRule#after() is called even when timeout is reached

Vlatombe:
I see from time to time some tests from `kubernetes-plugin` timing out. When that happens, subsequent tests fail because they try to allocate the same port. I've noticed that the timeout exception isn't handled the same way as a generic failure, in particular it doesn't run `after()`.


jglick (commented):
Maybe right? But could also perhaps regress something. I wonder if `JenkinsSessionRule` behaves more nicely?

jglick:
At least merge with `master` to rerun CI.

jglick:
> it doesn't run `after()`.

But that may be a good thing, since shutdown invokes user code and thus could hang, right?

PR #280, opened 2021-03-03
[JENKINS-41827] RealJenkinsRule

jglick:
A basic version of [JENKINS-41827](https://issues.jenkins.io/browse/JENKINS-41827): run functional tests in such a way that Jenkins actually gets its own JVM, started normally via Winstone, with plugins in their normal places (well we use one `*.jpl` but that is generally innocuous), and all the test code is dropped into some garbage class loader on top.

There is a bunch of stuff there which is not implemented, most notably the ability to use a Docker fixture for the controller, but this seems like a good start. Needs a lot more proofing.

It is not exactly a drop-in replacement for `JenkinsRule` but I tried to keep it as close as practical; see https://github.com/jenkinsci/artifact-manager-s3-plugin/pull/187 where I am actually reusing a test utility defined in `workflow-api` which expects `JenkinsRule`.


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -0,0 +1,308 @@
...
+    // · runRemotely: run a test thunk remotely, given a running session (whitebox testing)
+    // Thus rr.then(s) would be shorthand for rr.withSession(() -> rr.runRemotely(s))
+
+    /**
+     * Run one Jenkins session, send a test thunk, and shut down.
+     */
+    public void then(Step s) throws Throwable {
+        Body.writeSer(new File(home, "step.ser"), s);
+        ProcessBuilder pb = new ProcessBuilder(
+                /* TODO take from current JRE */"java",
```
timja, line 186: do you plan to address in this PR?

jglick: I could, especially if you think it might affect you, or some CI system.

timja (approved):
nice!

oleg-nenashev (commented):
If it works, awesome

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -0,0 +1,348 @@
...
+    // Thus rr.then(s) would be shorthand for rr.withSession(() -> rr.runRemotely(s))
+
+    /**
+     * Run one Jenkins session, send a test thunk, and shut down.
+     */
+    public void then(Step s) throws Throwable {
+        Body.writeSer(new File(home, "step.ser"), s);
+        String cp = System.getProperty("java.class.path");
+        ProcessBuilder pb = new ProcessBuilder(
+                new File(System.getProperty("java.home"), "bin/java").getAbsolutePath(),
```
jtnord, line 225: I wonder if this should pickup the JVM memory args from surefire so that it doesn't just allocate random amount of memory?

Or should we make it an option in the actual Rule?

jglick: https://github.com/jenkinsci/jenkins-test-harness/blob/9f4ca88414b78e8420e0bc55b71b8c4e3ca767b8/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java#L236 Plenty of room for improvement, consider this an MVP.

PR #287, opened 2021-03-25
RealJenkinsRule improvements

jglick:
- Call `stopJenkins` automatically
- Ability to disable plugins
- Ability to add Java options
- Apply an overall timeout, and print diagnostics if Jenkins cannot be contacted or the test hangs

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -228,6 +228,9 @@ public final class RealJenkinsRule implements TestRule {
                     }
                     base.evaluate();
                 } finally {
+                    if (proc != null) {
+                        stopJenkins();
+                    }
                     try {
                         tmp.dispose();
                     } catch (Exception x) {
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -364,8 +364,6 @@ public final class RealJenkinsRule implements TestRule {
                     throw new AssertionError("Jenkins did not start after 3m");
                 } else if (tries % /* 1m */ 600 == 0) {
                     x.printStackTrace();
-                } else if (tries % /* 5s */ 50 == 0) {
-                    System.err.println("Jenkins is not yet ready: " + x);
                 }
             }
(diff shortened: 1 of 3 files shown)


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -222,12 +245,17 @@
                                 } else {
                                     FileUtils.copyURLToFile(new URL(index, line + ".hpi"), new File(plugins, line + ".jpi"));
                                 }
-                                // TODO add method to disable a plugin (e.g. to test optional dependencies)
                             }
                         }
                     }
+                    for (String p : disabledPlugins) {
```
jtnord, line 251: I am assuming that we just do not skip copying the plugin as we re-use the plugin directory for all the test that may use a `RealJenkinsRule`?  If not why not just skip them rather than disabling them (and change disable plugins to `withoutPlugins` as that is a better use case at least for operations center things?)

jtnord (approved):

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jglick: > why not just skip [copying] them rather than disabling them

Good point, that would probably work. Will try it…

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -414,6 +461,12 @@ public static void register() throws Exception {
             Jenkins.get().setNoUsageStatistics(true);
             DownloadService.neverUpdate = true;
             UpdateSite.neverUpdate = true;
+            System.err.println("RealJenkinsRule ready");
+            Timer.get().schedule(JenkinsRule::dumpThreads, 2, TimeUnit.MINUTES);
+            Timer.get().schedule(() -> {
+                JenkinsRule.dumpThreads();
+                System.exit(1);
+            }, 5, TimeUnit.MINUTES);
```
timja, line 469: err can't this be configurable in some way, this test takes about 5 minutes to run.

https://github.com/jenkinsci/azure-container-agents-plugin/blob/dev/src/test/java/com/microsoft/jenkins/containeragents/AciCloudIT.java#L51

jtnord (approved):

PR #289, opened 2021-03-28
Add remote host binding for JenkinsRule

fr123k:
The reason for this change is that the JenkinRule based tests were all running
inside a docker container and i needed a way to bind it a non localhost address
in order to access it from outside the running container.

The host binding for the JenkinsRule is configurable with the `host`
property. The default value is `localhost` this keeps the backward
compatibility.

The following code snippet shows how to bind the JenkinsRule instance
```
@Rule public JenkinsRule j = new JenkinsRule() {{
    host = null;
  }
};
```
to all network interfaces. This is similar to set the `host` value to `0.0.0.0`.

The test case `JenkinsRuleHostIPTest` validates that the assignment to a specific
host ip address is working correctly.

The test case `JenkinsRuleAllHostInterfaceTest` validates that the assignment to all
network interfaces is working correctly.

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!

### DIFF
--- a/src/main/java/jenkins/benchmark/jmh/JmhBenchmarkState.java
+++ b/src/main/java/jenkins/benchmark/jmh/JmhBenchmarkState.java
@@ -6,6 +6,7 @@ import hudson.security.ACL;
 import jenkins.model.Jenkins;
 import jenkins.model.JenkinsLocationConfiguration;
 import org.apache.commons.lang3.mutable.MutableInt;
+import org.apache.commons.lang3.mutable.MutableObject;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.eclipse.jetty.server.Server;
 import org.jvnet.hudson.test.JenkinsRule;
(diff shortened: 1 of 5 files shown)


oleg-nenashev (approved):
Looks good to me 👍

jglick:
This is basically #84 I think, and is potentially problematic for the same reason. See https://github.com/jenkinsci/kubernetes-plugin/blob/55e01eaca472ddaf7f4ae15ea9569571db6672ab/src/test/java/org/jvnet/hudson/test/JenkinsRuleNonLocalhost.java

fr123k:
That's true if you bind JenkinsRule instance to 0.0.0.0. Then its possible to access it from remote depends on the network setup.

I could add a log warning too.

The default is still localhost and has to be conciously overwritten with a different IP or hostname.

jglick:
https://github.com/jenkinsci/kubernetes-plugin/pull/1593 worked without this. We already have https://github.com/jenkinsci/jenkins-test-harness/pull/541. I do not think this is advisable or necessary.

PR #292, opened 2021-03-29
Using a plugin to implement RealJenkinsRuleInit rather than a Groovy init script

jglick:
Amends #281. This allows initialization, especially of `Main.isUnitTest = true` and `DownloadService` / `UpdateSite.neverUpdate`, to happen much earlier during startup, thus solving some issues I have seen. For example, even after https://github.com/jenkinsci/artifact-manager-s3-plugin/pull/196 I have seen some flaky `OutOfMemoryError`s on Java 11 which seem to be due to parsing update center metadata, something we are attempting to suppress.


RealJenkinsRuleInit/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+                        </goals>
+                        <configuration>
+                            <artifactItems>
+                                <artifactItem>
+                                    <groupId>${project.groupId}</groupId>
+                                    <artifactId>${project.artifactId}</artifactId>
+                                    <version>${project.version}</version>
+                                    <type>${project.packaging}</type>
+                                    <outputDirectory>${project.basedir}/../src/main/resources/org/jvnet/hudson/test</outputDirectory>
+                                    <destFileName>RealJenkinsRuleInit.jpi</destFileName>
```
jglick, line 40: For now, building this artifact manually, since it looks awkward to include it in the main module’s build process (I do not particular want to switch to a reactor build).

RealJenkinsRuleInit/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+        <artifactId>plugin</artifactId>
+        <version>4.17</version>
+        <relativePath />
+    </parent>
+    <groupId>io.jenkins.plugins</groupId>
+    <artifactId>RealJenkinsRuleInit</artifactId>
+    <version>0-SNAPSHOT</version>
+    <packaging>hpi</packaging>
+    <properties>
+        <jenkins.version>2.204</jenkins.version>
```
jglick, line 15: https://github.com/jenkinsci/plugin-pom/pull/269#discussion_r358529278

jtnord (approved):

PR #303, opened 2021-04-30
[JENKINS-48466] JenkinsRule support for JUnit 5

offa:
Provides JUnit 5 support for `JenkinsRule` as discussed in [JENKINS-48466](https://issues.jenkins.io/browse/JENKINS-48466).

Tests written with JUnit 5 don't support rules, thus require code like this to work with the test harness.

#### Usage example:

```java
// JUnit 5
@ExtendWith(JenkinsRuleResolver.class)
class ExampleJUnit5Test {

    @Test
    public void example(JenkinsRuleExtension r) {
        // use 'r' as usual
        // ...
    }

}
```

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!
- [x] ~~Link to relevant pull requests, esp. upstream and downstream changes~~

/cc @uhafner

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,6 +53,7 @@ THE SOFTWARE.
     <jetty.version>9.4.40.v20210413</jetty.version>
     <hamcrest.version>2.2</hamcrest.version>
     <jmh.version>1.30</jmh.version>
+    <junit.jupiter.version>5.7.1</junit.jupiter.version>
     <java.level>8</java.level>
     <concurrency>1</concurrency> <!-- may use e.g. 2C for 2 × (number of cores) -->
     <!--TODO: fix FindBugs-->
@@ -147,6 +148,23 @@ THE SOFTWARE.
           <artifactId>commons-io</artifactId>
         </exclusion>
       </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <version>${junit.jupiter.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
+      <version>${junit.jupiter.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.vintage</groupId>
+      <artifactId>junit-vintage-engine</artifactId>
+      <version>${junit.jupiter.version}</version>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.jvnet.hudson</groupId>
--- a/src/main/java/org/jvnet/hudson/test/junit/jupiter/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/junit/jupiter/JenkinsRule.java
@@ -17,38 +17,34 @@ import java.lang.annotation.Target;
  * <p>
  * Annotating a <i>class</i> provides access for all of it's tests. Unrelated test cases can omit the parameter.
  *
- * <pre>
- * {@code
- * @JenkinsRule
+ * <blockquote><pre>
+ * &#64;JenkinsRule
  * class ExampleJUnit5Test {
  *
- *     @Test
+ *     &#64;Test
  *     public void example(JenkinsRuleExtension r) {
  *         // use 'r' ...
  *     }
  *
- *     @Test
+ *     &#64;Test
  *     public void exampleNotUsingRule() {
  *         // ...
  *     }
  * }
- * }
- * </pre>
+ * </pre></blockquote>
  * <p>
  * Annotating a <i>method</i> limits access to the method.
  *
- * <pre>
- * {@code
+ * <blockquote><pre>
  * class ExampleJUnit5Test {
  *
- *     @JenkinsRule
- *     @Test
+ *     &#64;JenkinsRule
+ *     &#64;Test
  *     public void example(JenkinsRuleExtension r) {
  *         // use 'r' ...
  *     }
  * }
- * }
- * </pre>
+ * </pre></blockquote>
  *
  * @see org.jvnet.hudson.test.junit.jupiter.JenkinsRuleResolver
  * @see org.junit.jupiter.api.extension.ExtendWith
--- a/src/main/java/org/jvnet/hudson/test/junit/jupiter/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/junit/jupiter/JenkinsRule.java
@@ -9,7 +9,49 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * JUnit 5 meta annotation providing {@link org.jvnet.hudson.test.JenkinsRule} integration.
+ * JUnit 5 meta annotation providing {@link org.jvnet.hudson.test.JenkinsRule JenkinsRule} integration.
+ *
+ * <p>
+ * Test methods using the rule extension need to accept it by {@link JenkinsRuleExtension} parameter; each test case
+ * gets a new rule object.
+ * <p>
+ * Annotating a <i>class</i> provides access for all of it's tests. Unrelated test cases can omit the parameter.
+ *
+ * <pre>
+ * {@code
+ * @JenkinsRule
+ * class ExampleJUnit5Test {
+ *
+ *     @Test
+ *     public void example(JenkinsRuleExtension r) {
+ *         // use 'r' ...
+ *     }
+ *
+ *     @Test
+ *     public void exampleNotUsingRule() {
+ *         // ...
+ *     }
+ * }
+ * }
+ * </pre>
+ * <p>
+ * Annotating a <i>method</i> limits access to the method.
+ *
+ * <pre>
+ * {@code
+ * class ExampleJUnit5Test {
+ *
+ *     @JenkinsRule
+ *     @Test
+ *     public void example(JenkinsRuleExtension r) {
+ *         // use 'r' ...
+ *     }
+ * }
+ * }
+ * </pre>
+ *
+ * @see org.jvnet.hudson.test.junit.jupiter.JenkinsRuleResolver
+ * @see org.junit.jupiter.api.extension.ExtendWith
  */
 @Target({ElementType.TYPE, ElementType.METHOD})
 @Retention(RetentionPolicy.RUNTIME)
--- a/src/main/java/org/jvnet/hudson/test/junit/jupiter/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/junit/jupiter/JenkinsRule.java
@@ -1,0 +1,19 @@
+package org.jvnet.hudson.test.junit.jupiter;
+
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * JUnit 5 meta annotation providing {@link org.jvnet.hudson.test.JenkinsRule} integration.
+ */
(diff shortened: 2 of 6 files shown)


offa:
Unfortunately I can't reproduce both CI test failures locally.

offa:
I have executed the tests on a few different systems now: Linux and Windows with Java 11 and 8. While Java 11 is green on all systems consistently, I managed to get the `JenkinsRuleTimeoutTest.handInterruptiblyInShutdown` to fail _occasionally_ on Linux Java 8.

However, the test failure is caused by different `IllegalStateException` reasons - depending on how the test are run: _"Jenkins.instance is missing. Read the documentation of Jenkins.getInstanceOrNull to see what you are doing wrong."_ starting `mvn clean test` through IDE; _"Expected 1 instance of hudson.model.UserIdMapper but got 0"_ running `mvn clean test` from shell. Test execution from IDE is always green.

offa:
I have added a `@JenkinsRule` meta annotation as a shorthand for `@ExtendWith(...)`. The example above can be written like this now:

```java
// JUnit 5
@JenkinsRule
class ExampleJUnit5Test {
    // ...
}
```

offa:
I get the test failure on the master branch occasionally too (Java 8 only). Can anyone more involved in this project take look at it?

offa:
ping @jglick @oleg-nenashev

pom.xml
```diff
@@ -148,6 +149,23 @@ THE SOFTWARE.
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <version>${junit.jupiter.version}</version>
+    </dependency>
```
jglick, line 156: Hope this alone does not cause problems like [SUREFIRE-1911](https://issues.apache.org/jira/browse/SUREFIRE-1911).

jglick (commented):
Merge conflict somewhere.

Seems fine though there are a bunch of other rules which will not work in 5 mode anyway, so I do not see any reason for people to expend effort adopting 5. (IMO it is overcomplicated—aggravates the abuse of annotations introduced in 4—and has no obviously compelling advantage.)

src/test/java/org/jvnet/hudson/test/junit/jupiter/JenkinsRuleResolverTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.collection.IsCollectionWithSize.hasSize;
+import static org.hamcrest.collection.IsEmptyCollection.empty;
+
+@JenkinsRule
+class JenkinsRuleResolverTest {
+
+    @Test
+    public void jenkinsRuleIsAccessible(JenkinsRuleExtension r) throws IOException {
```
uhafner, line 15:
```suggestion
   void jenkinsRuleIsAccessible(JenkinsRuleExtension r) throws IOException {
```

uhafner (commented):
Is the build failure related to this change? Other PRs seem to build fine without failures. It would be helpful to have a successful build so I can try the new rule in one of my plugins (by using the version of the incrementals repository).

jglick:
> Is the build failure related to this change?

Probably not, except perhaps by affecting timing. See #308.

src/main/java/org/jvnet/hudson/test/junit/jupiter/JenkinsRuleResolver.java
```diff
@@ -0,0 +1,43 @@
...
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.ParameterContext;
+import org.junit.jupiter.api.extension.ParameterResolutionException;
+import org.junit.jupiter.api.extension.ParameterResolver;
+import org.jvnet.hudson.test.JenkinsRule;
+
+/**
+ * JUnit 5 extension providing {@link JenkinsRule} integration.
+ */
+public class JenkinsRuleResolver implements ParameterResolver, AfterEachCallback {
```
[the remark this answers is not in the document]
offa: Both rule annotations are superseded by [`@ExtendWith` / `@RegisterExtension`](https://junit.org/junit5/docs/current/user-guide/#migrating-from-junit4-tips).

According to the [docs](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/extension/ExtendWith.html), `@ExtendWith` can be used on test class and method level.

offa:
> Seems fine though there are a bunch of other rules which will not work in 5 mode anyway, so I do not see any reason for people to expend effort adopting 5. (IMO it is overcomplicated—aggravates the abuse of annotations introduced in 4—and has no obviously compelling advantage.)

Indeed, there's still work to do for full JUnit 5 support (out of scope of this PR though). Additional support shouldn't force anyone to migrate of course. On the other hand there are users of Junit 5 already, those are more or less on their own at the moment. So at least we should get things started.

> > Is the build failure related to this change?
> 
> Probably not, except perhaps by affecting timing. See #308.

Nice catch! Thank you!

src/main/java/org/jvnet/hudson/test/junit/jupiter/JenkinsRuleResolver.java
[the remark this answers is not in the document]
jglick: Best to test that explicitly.

jglick:
> have a successful build so I can try the new rule in one of my plugins

@uhafner try `1557.v54a38bbede3b`

src/main/java/org/jvnet/hudson/test/junit/jupiter/JenkinsRuleResolver.java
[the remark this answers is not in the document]
offa: I have done a quick test: `@JenkinsRule` (= `@ExentdsWith`) can either be used on a class or specific method. If placed on the former, all tests within can access it; if on the latter, only the annotated test method can.

Or do you mean additional unit tests to verify this behaviour?

pom.xml
offa: Oh dear! That would be unfortunate :-(

src/main/java/org/jvnet/hudson/test/junit/jupiter/JenkinsRuleResolver.java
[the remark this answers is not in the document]
jglick: > can either be used on a class or specific method

That is something very different. (See the occasionally used `@WithoutJenkins` in 4 tests, though `JenkinsSessionRule` is the cleaner solution since you can explicitly control whether and when Jenkins starts and shuts down.)

I (and perhaps @uhafner also) was referring to the difference between

```java
public class XTest {
  @Rule public JenkinsRule r = new JenkinsRule();
  @Test public one() {
    WorkflowJob p = r.createProject(WorkflowJob.class, "p");
    // …
  }
  @Test public two() {
    WorkflowJob p = r.createProject(WorkflowJob.class, "p");
    // …
  }
}
```

and

```java
public class XTest {
  @ClassRule public static JenkinsRule r = new JenkinsRule();
  @Test public one() {
    WorkflowJob p = r.createProject(WorkflowJob.class, "p1");
    // …
  }
  @Test public two() {
    WorkflowJob p = r.createProject(WorkflowJob.class, "p2");
    // …
  }
}
```

`@ClassRule` is a bit faster since Jenkins only starts up once for the whole suite. (The effect is not as big as you might think, since most of the overhead is class loading, which applies only to the first test run—unlike `RealJenkinsRule`, Jenkins core & plugin classes are loaded in the app class loader.) But you must take care that tests do not interfere with one another. `@Rule` is safer since every test gets its own copy of Jenkins (incl. `$JENKINS_HOME`).

> do you mean additional unit tests to verify this behaviour

I did mean an automated test yes.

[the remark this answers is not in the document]
offa: Thanks for your detailed explanation. Now, only the behaviour of `@Rule` is supported – each test gets it's own `JenkinsRule` object.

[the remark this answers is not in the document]
jglick: Fine for now but should be documented explicitly.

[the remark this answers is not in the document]
uhafner: Yes, I meant the same thing. Thanks for providing the example! My integration tests are lasting quite some time so I tried to optimize the runtime. And using a `ClassRule` helps to reduce the runtime by around 20%. 

But anyway, I think this is a good first iteration, where the most important use case is working now. Thanks for adding it! I think this is a good basis and can be extended later with additional PRs.

src/main/java/org/jvnet/hudson/test/junit/jupiter/JenkinsRule.java
```diff
@@ -0,0 +1,19 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * JUnit 5 meta annotation providing {@link org.jvnet.hudson.test.JenkinsRule} integration.
```
uhafner, line 12: I think it would be helpful if you can add an example snippet of a test case in the JavaDoc here (or in `JenkinsRuleExtension`). So one sees how to correctly create such a test method.

src/main/java/org/jvnet/hudson/test/junit/jupiter/JenkinsRuleExtension.java
```diff
@@ -0,0 +1,36 @@
...
+import org.jvnet.hudson.test.JenkinsRecipe;
+import org.jvnet.hudson.test.JenkinsRule;
+
+import javax.annotation.Nonnull;
+import java.lang.reflect.Method;
+import java.util.Optional;
+
+/**
+ * Provides JUnit 5 compatibility for {@link JenkinsRule}.
+ */
```
uhafner, line 15: See Jesse's comment below: make clear that the rule does create a new Jenkins instance per test.

uhafner (changes requested):
Migrating my warnings tests to the new rule is possible but requires some work so I did not really test this PR in detail. Basically every of the test methods needs a new parameter that contains the Jenkins rule now. So I think this new rule helps essentially new developers to implement new tests.

The code looks good to me, just some additional documentation (see review comments) would be nice.

src/main/java/org/jvnet/hudson/test/junit/jupiter/JenkinsRuleExtension.java
offa: Added to `@JenkinsRule`.

src/main/java/org/jvnet/hudson/test/junit/jupiter/JenkinsRule.java
offa: Done, examples and some additional documentation added to `@JenkinsRule` – typically this should be used mainly.

src/main/java/org/jvnet/hudson/test/junit/jupiter/JenkinsRuleResolver.java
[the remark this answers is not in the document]
offa: Documentation updated (as suggested by @uhafner below) and there's a test for method level annotation too now.

uhafner (approved):
Thanks!

pom.xml
jglick: I am afraid that acc. to `git bisect` on https://github.com/jenkinsci/plugin-pom/pull/416 it does: with this patch, _no_ JUnit 4 tests are run.

jglick: ⇒ #317

reda-alaoui: According to https://github.com/jenkins-infra/pipeline-library/pull/272, `junit-vintage-engine` shoudl have been added with scope `compile` instead of `test`.

PR #307, opened 2021-05-17
`JenkinsRuleTimeoutTest` was flaky

jglick:
See https://github.com/jenkinsci/jenkins-test-harness/pull/303#issuecomment-831811098. By running against an instrumented version of core, the root problem seems to be that `FailOnTimeout` does its interruption in a separate thread, so it is unpredictable when `Jenkins.cleanUp` will be called—sometimes _after_ another test has started `Jenkins.<init>`. `cleanUp` does supposedly verify that it is cleaning up `this`, and `JenkinsRule` has an additional defense, but it seems there are lots of race conditions here. Safest to run each of these tests in its own JVM. We do not really care about a test that failed due to timeout also “poisoning” other tests in the same JVM since we expect the test suite to fail at that point anyway, so this probably only affects `JenkinsRuleTimeoutTest` itself.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -241,7 +241,7 @@ THE SOFTWARE.
               <hudson.maven.debug>${mavenDebug}</hudson.maven.debug>
               <buildDirectory>${project.build.directory}</buildDirectory>
           </systemPropertyVariables>
-          <reuseForks>true</reuseForks>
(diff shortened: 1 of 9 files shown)


jglick:
Wrong branch. Refiled as #308.

PR #314, opened 2021-05-28
`WithPlugin` equivalent for `RealJenkinsRule`

basil:
A straightforward port of `@WithPlugin` to `RealJenkinsRule`.

### DIFF
--- a/src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
+++ b/src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
@@ -33,13 +33,14 @@ import java.io.IOException;
 import java.util.concurrent.atomic.AtomicReference;
 import jenkins.model.Jenkins;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 import org.junit.Rule;
 import org.junit.Test;
 
 public class RealJenkinsRuleTest {
 
-    @Rule public RealJenkinsRule rr = new RealJenkinsRule();
+    @Rule public RealJenkinsRule rr = new RealJenkinsRule().addPlugins("structs.hpi");
 
     @Test public void smokes() throws Throwable {
         rr.then(RealJenkinsRuleTest::_smokes);
@@ -47,6 +48,7 @@ public class RealJenkinsRuleTest {
     private static void _smokes(JenkinsRule r) throws Throwable {
         System.err.println("running in: " + r.jenkins.getRootUrl());
         assertTrue(Main.isUnitTest);
+        assertNotNull(r.jenkins.getPlugin("structs"));
     }
 
     @Test public void error() {
--- a/pom.xml
+++ b/pom.xml
@@ -290,6 +290,33 @@ THE SOFTWARE.
           </dependency>
         </dependencies>
       </plugin>
+      <plugin>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>extra-plugins</id>
+            <phase>process-test-resources</phase>
+            <goals>
+              <goal>copy</goal>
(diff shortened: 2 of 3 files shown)


src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
```diff
@@ -33,20 +33,22 @@
...
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 import org.junit.Rule;
 import org.junit.Test;
 
 public class RealJenkinsRuleTest {
 
-    @Rule public RealJenkinsRule rr = new RealJenkinsRule();
+    @Rule public RealJenkinsRule rr = new RealJenkinsRule().addPlugins("structs.hpi");
```
jglick, line 43: To match Javadoc,

```suggestion
    @Rule public RealJenkinsRule rr = new RealJenkinsRule().addPlugins("structs.jpi");
```

and adjust the `dependency:copy` goal to rename accordingly.

jglick (changes requested):

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
basil: What's the reason for this request?

jglick: To start with, the Javadoc specifically requests `*.jpi`. But that is not arbitrary: Jenkins core expects plugins to be stored as `$JENKINS_HOME/plugins/${artifactId}.jpi` and can get confused when this is not the case.

basil: Interesting. I've never noticed a difference when using `.hpi` vs `.jpi` with core. But you've been involved much longer than me and I trust that you know where the bodies are buried…

jglick: Jenkins _will_ load any files with a set of extensions in that directory, but the plugin manager has some expectations about filename, and will itself only ever be _storing_ downloaded plugins as `${artifactId}.jpi`. Cannot recall exact behavior as of today but see https://github.com/jenkinsci/jenkins/pull/4000 & https://github.com/jenkinsci/jenkins/pull/4100.

I would actually recommend breaking tradition with `@WithPlugins` and accepting only plugin short names, constructing the filename from that.

basil: > and adjust the `dependency:copy` goal to rename accordingly.

I don't see any way of using `dependency:copy` to rename the artifact. Googling the problem turned up [this page](https://stackoverflow.com/questions/3998596/renaming-resources-in-maven) which shows at least 4 ways to do this using various Maven plugins, none of which seem particularly elegant. Do you have a preference as to which of those 4 solutions I should try to implement (or some other idea entirely)? I figured I'd ask before going down this rabbit hole to avoid wasting time if I'm missing something. As must be obvious by now, I am a Maven novice. I sure wish I was back in Gradle where I could just write Groovy to do anything I want…

jglick: The Maven Ant plugin is convenient, or you can use the GMaven plugin and just write Groovy to do anything you want. :grin:

jglick: Alternately: change the API here to specify an absolute or relative resource path (use `Class.getResource`) and make `RealJenkinsRule` a bit smarter. You can pretty easily pop open the `JarFile` and look for the plugin short name and save exactly the right `$JENKINS_HOME/plugins/*.jpi` file regardless of the input filename.

PR #315, opened 2021-05-28
Make `RealJenkinsRule` usable in core

basil:
Before this change, `RealJenkinsRule` isn't usable in core. After this change, it is.

Tested by copying `RealJenkinsRuleTest` to core and running it there.


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -336,6 +336,24 @@ private URL endpoint(String method) throws MalformedURLException {
         return new URL(getUrl(), "RealJenkinsRule/" + method + "?token=" + token);
     }
 
+    private static File findJenkinsWar() throws Exception {
+        // Adapted from WarExploder.explode
```
jglick, line 340: https://github.com/jenkinsci/jenkins-test-harness/blob/d3dacb2a75d1aeede2952f051ab6a9d1a8287526/src/main/java/org/jvnet/hudson/test/WarExploder.java#L105-L115

jglick (approved):

PR #317, opened 2021-06-02
Revert JUnit 5 integration

jglick:
Somehow the inclusion of JUnit 5 in the test classpath breaks Surefire. Until that is worked out, reverting #303.


jtnord:
I fixed something very similar in the Jenkins project a while ago. https://github.com/jenkinsci/jenkins/pull/5104/files
I hope we have not just lost a lot of coverage there :(

jglick:
Well this was using 5.7.2, so just updating to 5.7.0 as in https://github.com/jenkinsci/jenkins/pull/5104 is apparently not sufficient.

I have not yet tried to track down what exactly went wrong here. Looked similar to https://issues.apache.org/jira/browse/SUREFIRE-1911 but that seemed to be a regression in M5 whereas `plugin-pom` is still on M4. Did this reversion to unblock integration of unrelated JTH changes in `plugin-pom`, until someone can sort this out and reapply.

PR #323, opened 2021-06-15
`RealJenkinsRule` incompatible with servlet filters checking request parameters

joseblas:
When executing RealJenkinsRule post are send in binary mode. If any filter uses `request.getParameter` it throws a invalid UTF-8 code. Also, on the step method `invalid stream header: ED000573` was showing up.

In order to fix this, the step POST is sent using Base64 encoding for the step and token.


basil (changes requested):
Adapted from, not copied from. This version returns `jenkins.war`; the original returns `war/target/jenkins`. Deleting this code will regress #315.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -444,6 +444,7 @@ public void stopJenkins() throws Throwable {
 
     public void runRemotely(Step s) throws Throwable {
         HttpURLConnection conn = (HttpURLConnection) endpoint("step").openConnection();
+        conn.setRequestProperty("Content-Type", "text/plain; utf-8");
```
jglick, line 447: What? It is binary, not text.

RealJenkinsRuleInit/src/main/java/org/jvnet/hudson/test/RealJenkinsRuleInit.java
```diff
@@ -36,6 +44,30 @@ public RealJenkinsRuleInit() {}
...
+        /*
+        According to ServletRequest documentation:
+            If the parameter data was sent in the request body, such as occurs with an HTTP POST request,
+            then reading the body directly via getInputStream or getReader can interfere with the execution of this method.
+        Check RealJenkinsRule#runRemotely, it has
+         conn.setRequestProperty("Content-Type", "text/plain; utf-8");
+
+         to avoid this issue.
+        */
+        PluginServletFilter.addFilter(new Filter() {
```
jglick, line 56: I am not sure what this is doing here. Is this part of the patch fixing some bug, or demonstrating it?

jglick (commented):
> If any filter uses `request.getParameter` it throws a invalid UTF-8 code.

I have not seen such an error. Is there an example of such a filter? This needs a test case to demonstrate the problem being addressed.

> the step POST is sent using Base64 encoding

It is? I do not see any such change here.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
joseblas: It is binary but when a filter does a request.getParameter it invalidates the data hence an error when reading the data stream. 
If it's send with that Content Type, request.getParam goes well and the getInputStream works smoothly.

RealJenkinsRuleInit/src/main/java/org/jvnet/hudson/test/RealJenkinsRuleInit.java
joseblas: This is a filter I've added to make the tests fail. 
With the ContentType to text it works.

joseblas:
You're right, no text in base64 any longer. 

I noticed earlier that with the ContentType is enough to work, hence the reference. 
Apologies.

RealJenkinsRuleInit/src/main/java/org/jvnet/hudson/test/RealJenkinsRuleInit.java
jglick: Well revert this and do it properly in a single test case.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jglick: Weird. Definitely deserves to be noted in a comment here that the content type is totally incorrect.

joseblas: Indeed. After this comment, I tried out another Content-Type (what if missing the  content-type was the problem).
And if worked with application/octec-stream, which makes more sense that text.

RealJenkinsRuleInit/src/main/java/org/jvnet/hudson/test/RealJenkinsRuleInit.java
joseblas: I think adding complexity to the current plugin should be enough. By adding complexity to the current plugin used more cases are covered.
If I had a new test case for this particular case only, the options of having an issue when a filter is used are bigger.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jglick: So IIUC merely adding _any_ `Content-Type` header fixes the problem?

RealJenkinsRuleInit/src/main/java/org/jvnet/hudson/test/RealJenkinsRuleInit.java
jglick: Not following what you are saying. To repeat: if the bug is fixed by the one-line change to `RealJenkinsRule.java`, then please revert all changes to this file, and add a test case to `RealJenkinsRuleTest.java` which reproduces the error when you

```bash
git checkout master -- src/main
```

joseblas: `RealJenkinsRule.java` only has one line change. Regarding `RealJenkinsRuleTest` I think I needs no change.
In order to reproduce the bug I had to change `RealJenkinsRuleInit` to add a filter to the plugin. 
I do not know how to add a filter to a plugin from the test file.
Another option would be to add another plugin `RealJenkinsRuleInitWithFilter`.  Do you want that?

jglick: > I do not know how to add a filter to a plugin from the test file.

Just this same code, but from a `runRemotely` block. Presumably you will need to then run another no-op `runRemotely` block to reproduce the error after adding the filter.

jglick: Thanks. Should be able to do a final review Monday.

jglick (changes requested):
Please revert modifications to `src/main/resources/org/jvnet/hudson/test/RealJenkinsRuleInit.jpi`.

RealJenkinsRuleInit/src/main/java/org/jvnet/hudson/test/RealJenkinsRuleInit.java
jglick: > which reproduces the error

Checked locally. Does fix the error, though I do not get any colorful details, just

```
java.io.IOException: Server returned HTTP response code: 500 for URL: http://localhost:63078/jenkins/RealJenkinsRule/step?token=8b4e3d47-75e9-4cba-a236-51e6514d381c
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1900)
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1498)
	at org.jvnet.hudson.test.RealJenkinsRule.runRemotely(RealJenkinsRule.java:449)
	at org.jvnet.hudson.test.RealJenkinsRuleTest.testFilter(RealJenkinsRuleTest.java:62)
```

PR #337, opened 2021-10-17
Terminology Cleanup: Introduce `createAgent()` and `createOnlineAgent()` in `JenkinsRule`

oleg-nenashev:
Cleans up API for JenkinsRule. I did not touch `HudsonTestCase`, because we might want to drop JUnit 3 support alltogether


src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -903,39 +903,89 @@ public File createTmpDir() throws IOException {
         return env.temporaryDirectoryAllocator.allocate();
     }
 
+    // TODO replace return types once the Jenkins core is updated
```
jglick, line 906: Do you really foresee such a class rename in core? Seems like a lot of disruption for a cosmetic issue.

jglick (commented):
-0 I guess. Just adds complexity to the API surface. Would prefer to limit renames to user-facing strings.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
oleg-nenashev: I do not expect it anytime soon TBH. Depends on contributions though

jglick:
More inclined to just use the new terminology incrementally as changes are made, as in #428.

PR #346, opened 2021-11-14
`addPlugins` does not currently take effect when used inside test method

basil:
Moves the existing logic a little further down so that it takes effect in more scenarios.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -315,22 +315,6 @@ public final class RealJenkinsRule implements TestRule {
                             }
                         }
                     }
-                    for (String extraPlugin : extraPlugins) {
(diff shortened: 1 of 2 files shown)


jglick (commented):
I was considering this change earlier but rejected it because it is not exactly right, either. This copies specified files to `$JENKINS_HOME/plugins/` before every startup, unlike the plugins from `test-dependencies` which are just copied once, when preparing `$JENKINS_HOME` initially.

Arguably what a caller would expect is for a call to `addPlugins` before the test runs to collect its arguments, including them in the initial list in `apply`; but a call after `$JENKINS_HOME` already exists to immediately copy the plugin there (once). A bit more complicated but seems better.

jglick:
I suppose better would be for all the `$JENKINS_HOME` setup currently done eagerly inside `apply` should be deferred until `startJenkins` (or `getHome`) is first called.

basil:
I have not found the time to get back to this, so closing this for now so we do not waste build resources on it. Great task for a volunteer to pick up!

PR #347, opened 2021-11-18
[JENKINS-67086] Ensure a free port is used by RealJenkinsRule

Vlatombe:
[JENKINS-67086](https://issues.jenkins.io/browse/JENKINS-67086)

When running tests in parallel I sometimes see port conflicts

```
java.io.IOException: Failed to start Jetty
	at winstone.Launcher.<init>(Launcher.java:194)
	at winstone.Launcher.main(Launcher.java:369)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at Main._main(Main.java:375)
	at Main.main(Main.java:151)
Caused by: java.io.IOException: Failed to bind to /127.0.0.1:50786
	at org.eclipse.jetty.server.ServerConnector.openAcceptChannel(ServerConnector.java:349)
	at org.eclipse.jetty.server.ServerConnector.open(ServerConnector.java:310)
	at org.eclipse.jetty.server.AbstractNetworkConnector.doStart(AbstractNetworkConnector.java:80)
	at org.eclipse.jetty.server.ServerConnector.doStart(ServerConnector.java:234)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:73)
	at org.eclipse.jetty.server.Server.doStart(Server.java:401)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:73)
	at winstone.Launcher.<init>(Launcher.java:192)
	... 7 more
Caused by: java.net.BindException: Address already in use
	at sun.nio.ch.Net.bind0(Native Method)
	at sun.nio.ch.Net.bind(Net.java:461)
	at sun.nio.ch.Net.bind(Net.java:453)
	at sun.nio.ch.ServerSocketChannelImpl.bind(ServerSocketChannelImpl.java:222)
	at sun.nio.ch.ServerSocketAdaptor.bind(ServerSocketAdaptor.java:85)
	at org.eclipse.jetty.server.ServerConnector.openAcceptChannel(ServerConnector.java:344)
	... 14 more
```

The harness should ensure the port is free before trying to use it.

This includes a partial copy of IOUtil from acceptance-test-harness repo.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/IOUtil.java
+++ b/src/main/java/org/jvnet/hudson/test/IOUtil.java
@@ -1,0 +1,70 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2015 Red Hat, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.jvnet.hudson.test;
+
+import java.io.IOException;
+import java.net.ServerSocket;
+import java.util.logging.Logger;
+
+public class IOUtil {
+
+    private static final Logger LOGGER = Logger.getLogger(IOUtil.class.getName());
+
+    /**
+     * Gives random available TCP port in the given range.
+     *
+     * @param from if <=0 then default value 49152 is used
+     * @param to   if <=0 then default value 65535 is used
+     */
+    public static int randomTcpPort(int from, int to){
+        from = (from <=0) ? 49152 : from;
+        to = (to <= 0) ? 65535 : to;
+
+
+        while(true){
+            int candidate = (int) ((Math.random() * (to-from)) + from);
+            if(isTcpPortFree(candidate)){
+                return candidate;
+            }
+            LOGGER.info(String.format("Port %s is in use", candidate));
+        }
+    }
+
+    /**
+     * Gives random available TCP port.
+     */
+    public static int randomTcpPort(){
+        return randomTcpPort(-1,-1);
+    }
+
+    public static boolean isTcpPortFree(int port){
+        try {
+            ServerSocket ss = new ServerSocket(port);
+            ss.close();
+            return true;
+        } catch (IOException ex) {
+            return false;
+        }
+    }
+}
--- a/src/main/java/org/jvnet/hudson/test/IOUtil.java
+++ b/src/main/java/org/jvnet/hudson/test/IOUtil.java
@@ -27,6 +27,9 @@ import java.io.IOException;
 import java.net.ServerSocket;
 import java.util.logging.Logger;
 
+/**
+ * Copied from https://github.com/jenkinsci/acceptance-test-harness/blob/f749d27b229ce5a4737d1614fa1f31d2e6402f4f/src/main/java/org/jenkinsci/test/acceptance/utils/IOUtil.java#L80-L115
+ */
 public class IOUtil {
 
     private static final Logger LOGGER = Logger.getLogger(IOUtil.class.getName());
--- a/src/main/java/org/jvnet/hudson/test/IOUtil.java
+++ b/src/main/java/org/jvnet/hudson/test/IOUtil.java
@@ -34,8 +34,8 @@ public class IOUtil {
     /**
      * Gives random available TCP port in the given range.
      *
-     * @param from if <=0 then default value 49152 is used
-     * @param to   if <=0 then default value 65535 is used
+     * @param from if &lt;=0 then default value 49152 is used
+     * @param to   if &lt;=0 then default value 65535 is used
      */
     public static int randomTcpPort(int from, int to){
         from = (from <=0) ? 49152 : from;
--- a/src/main/java/org/jvnet/hudson/test/IOUtil.java
+++ b/src/main/java/org/jvnet/hudson/test/IOUtil.java
@@ -30,7 +30,7 @@ import java.util.logging.Logger;
 /**
  * Copied from https://github.com/jenkinsci/acceptance-test-harness/blob/f749d27b229ce5a4737d1614fa1f31d2e6402f4f/src/main/java/org/jenkinsci/test/acceptance/utils/IOUtil.java#L80-L115
  */
-public class IOUtil {
+class IOUtil {
 
     private static final Logger LOGGER = Logger.getLogger(IOUtil.class.getName());
 
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -58,7 +58,6 @@ import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.List;
 import java.util.Map;
-import java.util.Random;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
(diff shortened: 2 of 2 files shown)


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -246,7 +245,7 @@ public RealJenkinsRule withHost(String host) {
                 System.out.println("=== Starting " + description);
                 try {
                     home = tmp.allocate();
-                    port = new Random().nextInt(16384) + 49152; // https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Dynamic,_private_or_ephemeral_ports
+                    port = IOUtil.randomTcpPort(); // https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Dynamic,_private_or_ephemeral_ports
```
olamy, line 248: Jetty can be start using port 0 which will use any available port as defined by ServerSocket class (https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/net/ServerSocket.html#%3Cinit%3E(int)) then you need the port really used after Jetty has been started. 
see how it's done here https://github.com/jenkinsci/winstone/blob/08016690a4b5d70f97e16769838d7694f13774bc/src/test/java/winstone/HttpConnectorFactoryTest.java#L24
I don't have all the context here but there can be still problem here depending how long it take for the os to really free the port.

Vlatombe: `RealJenkinsRule` starts Jenkins in a separate jvm, would need a different method to retrieve the port.

olamy: oh ok. Sounds hackhish to read logs :) 
you should probably the ServerSocket port 0 feature rather than some random.
Look how it's done in Jetty testing https://github.com/eclipse/jetty.project/blob/ec3d7dbe92532cfdbb719cb7b736ad115aea40e5/tests/test-distribution/src/main/java/org/eclipse/jetty/tests/distribution/JettyHomeTester.java#L193

Vlatombe: Exploring some patch to winstone that would allow to pass a file name where to write the port to.

src/main/java/org/jvnet/hudson/test/IOUtil.java
```diff
@@ -0,0 +1,70 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.jvnet.hudson.test;
+
+import java.io.IOException;
+import java.net.ServerSocket;
+import java.util.logging.Logger;
+
+public class IOUtil {
```
jglick, line 30: Please leave a comment with the permalink of where you copied this from. Also
```suggestion
class IOUtil {
```

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jglick, line 248:
```suggestion
                    port = IOUtil.randomTcpPort();
```

jglick: No need for that. You can just have `Endpoint.register` write the port number to a file in `$JENKINS_HOME`. While you are at it, you could have the Jenkins JVM create the random token instead of reading `RealJenkinsRule.token` from a system property, and just write the whole endpoint URL incl. query string to that file: https://github.com/jenkinsci/jenkins-test-harness/blob/959ceb93f5a6cd6927e863de3a5e9b738dd8a051/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java#L388

**Edit**: no you cannot, because https://github.com/jenkinsci/jenkins-test-harness/blob/959ceb93f5a6cd6927e863de3a5e9b738dd8a051/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java#L381 would no longer be implementable (and some tests in fact use that method to determine the URL before Jenkins starts).

src/main/java/org/jvnet/hudson/test/IOUtil.java
```diff
@@ -0,0 +1,70 @@
...
+     */
+    public static int randomTcpPort(){
+        return randomTcpPort(-1,-1);
+    }
+
+    public static boolean isTcpPortFree(int port){
+        try {
+            ServerSocket ss = new ServerSocket(port);
+            ss.close();
+            return true;
```
jglick, line 65: I had actually thought about doing exactly this when first writing `RealJenkinsRule`. I avoided it because I thought that, besides the rare race condition, it would not actually work because the Linux kernel can decline to let you open a server socket for two minutes after closing one. Maybe that only applies to sockets that had actually accepted connections?

jglick (approved):

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
Vlatombe: > No need for that.

There is a similar problem for ath, so I'd rather have something that works in both cases.

jglick: Maybe enhance https://github.com/jenkinsci/jenkins-test-harness/blob/ed2b9ab8b19eaabeb2db699f2db28364c9e96cb0/src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java#L120-L122 to enforce test coverage of the `getUrl` contract.

Vlatombe: > and some tests in fact use that method to determine the URL before Jenkins starts

Do you have some links?

Vlatombe: I just wonder whether that contract still makes sense if it prevents foolproof concurrent tests.

Vlatombe: WIP in https://github.com/jenkinsci/winstone/pull/174 but I can't get incrementals to work...

src/main/java/org/jvnet/hudson/test/IOUtil.java
jtnord: the ATH does this and it still has random failures.  would guess this is trading one form of random failure for another.

Vlatombe: That's why i'm trying https://github.com/jenkinsci/winstone/pull/174 (+ downstreams pending)

jglick: https://docs.oracle.com/javase/8/docs/api/java/net/Socket.html#setReuseAddress-boolean- sounds fine. While there would still be a race condition, it should be much less likely than in the status quo.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
Vlatombe: Is there any actual usage of `rr.getUrl()` in tests? I did a quick lookup in `jenkinsci` and couldn't find any beyond the test you linked.

jglick: There are several usages in @cloudbees that could not easily be rewritten and which I think correspond to a general pattern.

Vlatombe: ok found one indeed there, seems easy enough for me to rewrite if needed.

jglick: IIRC I found multiple and did not see a way to rewrite. If you have done a full search, please file PRs removing the `getUrl` call there and CC me and link to this PR. In general I think `getUrl` is useful because you may need to set up WireMock or a Docker fixture of some other service with the Jenkins URL as an initialization parameter (something you cannot modify once the mock service is running).

jglick: I still think the current patch is safer, but if you want to go this route, I think it would be best to delete `getUrl` (perhaps add a similar method with a different name) as it could no longer be implemented according to its original contract and so any tests previously calling it would need to be adjusted.

Vlatombe: I guess it could still exist with a different semantic (available as soon as you started Jenkins at least once).

Vlatombe: > I still think the current patch is safer

I know for a fact that for ATHs this aproach is better but not foolproof.

Vlatombe:
I think this is already a pretty good improvement considering the other way I'm exploring would require winstone + jenkins + jth + ath patches, which will take much more time to deliver anyway.

amuniz (approved):

jglick:
> I think this is already a pretty good improvement

Yes, this ought to solve the problem 99% of the time at least. If the port 0 system works we can consider that as a follow-up, but it is not backward compatible so some test cases would need to be prepared for it.

src/main/java/org/jvnet/hudson/test/IOUtil.java
olamy: sounds a bit complicated to me compared to use 0 port to get a free port such this one https://github.com/eclipse/jetty.project/blob/cd1f146867f66f37df73184a086f0e17828d70c8/tests/test-distribution/src/main/java/org/eclipse/jetty/tests/distribution/JettyHomeTester.java#L193
here you will eventually iterate on few used port before get a free whereas server port 0 just tell the os to give a free port. 
but as it seems to be a temporary solution until winstone patch...

PR #348, opened 2021-11-18
Ensure free port winstone patch

Vlatombe:


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -639,7 +659,8 @@ public CustomJenkinsRule() throws Exception {
         }
 
         @Override public URL getURL() throws IOException {
-            return new URL(System.getProperty("RealJenkinsRule.url"));
+            int port = readPort(new File(System.getProperty("winstone.portFileName")));
```
jglick, line 662: Should be unnecessary to use a sysprop.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -426,7 +429,8 @@ public void startJenkins() throws Throwable {
         argv.addAll(javaOptions);
         argv.addAll(Arrays.asList(
                 "-jar", findJenkinsWar().getAbsolutePath(),
-                "--httpPort=" + port, "--httpListenAddress=127.0.0.1",
+                "--httpPort=" + port,
```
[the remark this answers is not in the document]
Vlatombe: Yes. the purpose is to retain the fixed port across restarts (assuming this is brief and there is little room for another process to steal the port in between)

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
Vlatombe: I want something that can be used for ath as well, https://github.com/jenkinsci/acceptance-test-harness/pull/711/

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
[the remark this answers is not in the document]
Vlatombe: Commented the field.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -142,6 +141,9 @@
 
     /**
      * TCP/IP port that the server is listening on.
+     *
+     * Before the first start, it will be 0. Once started, it is set to the actual port Jenkins is listening to.
+     *
```
jglick, line 146:
```suggestion
     * <p>
     * Before the first start, it will be 0. Once started, it is set to the actual port Jenkins is listening to.
     * <p>
```

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jtnord: doesn't this need to be relative to `JENKINS_HOME`, otherwise you would not be able to start 2 Jenkins with `RealJenkinsRule`?

Vlatombe: It is initialized at https://github.com/jenkinsci/jenkins-test-harness/pull/348/files#diff-ade203da1dc447eff46cf045f9a21d8ac019ccffdae26359a6838a54a1ac48fbR428, which is within `JENKINS_HOME`

pom.xml
```diff
@@ -88,7 +88,7 @@ THE SOFTWARE.
     <dependency>
       <groupId>org.jenkins-ci.main</groupId>
       <artifactId>jenkins-war</artifactId>
-      <version>2.204</version>
+      <version>2.338-rc32148.f12eee9c85e2</version><!-- https://github.com/jenkinsci/jenkins/pull/5928 and https://github.com/jenkinsci/winstone/pull/174 -->
```
jglick, line 91: Note that this means `jenkins-test-harness` and thus `plugin-pom` will require new `jenkins.version`s. Not necessarily a problem though (along with https://github.com/jenkinsci/plugin-pom/pull/478) it makes me reconsider my decision years ago to split these repos out of `jenkins`. It does mean that CloudBees CI development cannot pick up this fix for months. (CC @aneveux!)

The alternative is to have RJR detect the version of Winstone and fall back to the current “random & pray” algorithm for older versions, with a `TODO` comment to simplify in a year or so. Would make the logic harder to read and test, though.

jglick:
Do we even need https://github.com/jenkinsci/winstone/pull/174? Just pass `--httpPort=0` and have `RealJenkinsRule.Endpoint.register` write the port number to a well-known file in `$JENKINS_HOME`. (Pass the selected port number on subsequent starts since some tests expect the Jenkins URL to be stable across sessions. Reintroduces the race condition for tests which do use multiple sessions, but these are in the minority and I think we have no alternative since you cannot just reserve a port without binding it.)

Mentioned previously in https://github.com/jenkinsci/jenkins-test-harness/pull/347#discussion_r752198494. That discussion diverged into whether `RealJenkinsRule.getUrl` needed to be supported prior to first startup for some tests. But this PR already breaks that semantics, and has no advantage that I can see over `register`, but the clear disadvantage of only working with new Jenkins versions.

There was also a discussion of `acceptance-test-harness`, and perhaps ATH could benefit from the new Winstone feature (I have not studied that case), but this does not mean `RealJenkinsRule` needs to rely on it. Am I missing something?

Vlatombe:
Correct, for RJR we could probably take the route you propose.

jglick:
I tried to prototype that but hit a wall—there is no apparent way from inside Jenkins code to tell which port Jetty started on, without having a `HttpServletRequest`. The JUnit JVM could parse the stdio of the process looking for messages like

```
o.e.j.server.AbstractConnector#doStart: Started ServerConnector@4201c465{HTTP/1.1,[http/1.1]}{127.0.0.1:32955}
```

but that does not seem reliable.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -381,7 +383,7 @@ public void then(Step s) throws Throwable {
     }
 
     /**
-     * Like {@link JenkinsRule#getURL} but does not require Jenkins to have been started yet.
+     * Like {@link JenkinsRule#getURL} but can be called even does not require Jenkins to have been started yet.
```
Vlatombe, line 386:
```suggestion
     * Like {@link JenkinsRule#getURL} but does not require Jenkins to have been started yet.
```

jglick:
https://github.com/jenkinsci/jenkins-test-harness/pull/348#discussion_r820697503 again. Is it feasible to make this work when on newer Jenkins versions but fall back to the previous behavior? Otherwise rollout will be a headache since Dependabot POM updates will break nearly every plugin.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -533,6 +556,27 @@ public void startJenkins() throws Throwable {
         }
     }
 
+    private String getWarAbsolutePath() throws Exception {
+        return war == null ? findJenkinsWar().getAbsolutePath() : war.getAbsolutePath();
+    }
+
+    private static boolean supportsPortFileName(String war) throws IOException {
+        try (JarFile warFile = new JarFile(war)) {
```
jglick, line 564: Is this called multiple times per test? If so, can we cache t? Might be slow on Windows.

jglick (approved):
Looks right.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
Vlatombe: I can probably optimize it.

PR #350, opened 2021-11-28
Remove usages of JNR

basil:
Matching https://github.com/jenkinsci/jenkins-test-harness/blob/32abb9aca149ad502edcd1e2669a938ee25321a4/src/main/java/org/jvnet/hudson/test/JenkinsRule.java#L2857-L2863

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
+++ b/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
@@ -90,7 +90,6 @@ import hudson.model.TaskListener;
 import hudson.model.UpdateSite;
 import hudson.model.User;
 import hudson.model.View;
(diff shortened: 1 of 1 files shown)


MarkEWaite (approved):
I was worried that systems that don't always have glibc (like OpenBSD, FreeBSD, ...) might lose functionality from this change.  However, when I run `println(Functions.isGlibcSupported())` on my OpenBSD and FreeBSD agents, they report `true`.  Based on that result and based on this being test harness code in the deprecated HudsonTestCase, I approve.

basil:
Indeed. More importantly, the code was already like this in `JenkinsRule`, so this PR just makes the copypasta in `JenkinsRule` and `HudsonTestCase` identical.

PR #352, opened 2021-11-29
Allowing `RealJenkinsRule` to use `@LocalData`

jglick:
Keeping in draft until I have a working plugin test demonstrating effectiveness.

PR #353, opened 2021-12-02
Add new JenkinsSessionRule.getHome method

dwnusbaum:
Analogous to `RealJenkinsRule.getHome`, which was added in #352. Useful in cases where you want to modify something in `JENKINS_HOME` between Jenkins restarts.

Once the CI build completes I will update https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/185 to consume the new method.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsSessionRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsSessionRule.java
@@ -31,6 +31,7 @@ import org.junit.After;
 import org.junit.rules.TestRule;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
+import org.jvnet.hudson.test.recipes.LocalData;
 
 /**
  * Simpler alternative to {@link RestartableJenkinsRule}.
--- a/src/main/java/org/jvnet/hudson/test/JenkinsSessionRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsSessionRule.java
@@ -60,7 +60,9 @@ public class JenkinsSessionRule implements TestRule {
      * Normally it will suffice to use {@link LocalData} to populate files.
      */
     public File getHome() {
-        assert home != null : "JENKINS_HOME has not been allocated yet";
+        if (home == null) {
+            throw new IllegalStateException("JENKINS_HOME has not been allocated yet");
+        }
         return home;
     }
 
--- a/src/main/java/org/jvnet/hudson/test/JenkinsSessionRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsSessionRule.java
@@ -55,6 +55,15 @@ public class JenkinsSessionRule implements TestRule {
      */
     private int port;
 
+    /**
+     * Get the Jenkins home directory, which is consistent across restarts.
+     * Normally it will suffice to use {@link LocalData} to populate files.
+     */
+    public File getHome() {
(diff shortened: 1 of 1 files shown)


basil (approved):

src/main/java/org/jvnet/hudson/test/JenkinsSessionRule.java
```diff
@@ -55,6 +55,15 @@
      */
     private int port;
 
+    /**
+     * Get the Jenkins home directory, which is consistent across restarts.
+     * Normally it will suffice to use {@link LocalData} to populate files.
```
jglick, line 60: I am not actually sure if `@LocalData` works with `JenkinsSessionRule`.

src/main/java/org/jvnet/hudson/test/JenkinsSessionRule.java
```diff
@@ -55,6 +55,15 @@
...
+     */
+    public File getHome() {
+        assert home != null : "JENKINS_HOME has not been allocated yet";
```
jglick, line 63: Probably better made into an `IllegalStateException`, though assertions ought to be enabled in tests IIRC.

jglick (approved):

jglick:
> Once the CI build completes

As soon as you are successfully using it in a test, just let me know and I can merge this which will release it too.

src/main/java/org/jvnet/hudson/test/JenkinsSessionRule.java
dwnusbaum: Yeah, I had it that way at first, but from some quick testing was not able to trigger the condition in normal usage so I just used an assertion for brevity. I can switch back.

src/main/java/org/jvnet/hudson/test/JenkinsSessionRule.java
dwnusbaum: Ok I will check.

dwnusbaum: From a quick test `@LocalData` does seem to work with `JenkinsSessionRule`.

src/main/java/org/jvnet/hudson/test/JenkinsSessionRule.java
dwnusbaum: 8f5d840

src/main/java/org/jvnet/hudson/test/JenkinsSessionRule.java
dwnusbaum: Ok looking a bit deeper, the behavior is not really what you would want, so I will remove the recommendation. `@LocalData` does function, but each time Jenkins restarts, the `@LocalData` replaces the contents of `JENKINS_HOME`, and using it at all causes a new `JENKINS_HOME` to be allocated by `HudsonHomeLoader`, leaving `JenkinsSessionRule.home` unused (and distinct from `Jenkins.root`).

dwnusbaum: See 71560d6.

dwnusbaum:
@jglick https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/185 has been updated to use the new method.

jglick:
Looks good! If all goes well you should see a new release of this component in half an hour or so.

PR #357, opened 2021-12-08
GC before trying to delete the Jenkins home directory

basil:
We run `System.gc()` after shutting down Jenkins and deleting the Jenkins home directory, but by then it is too late on Windows: deleting the Jenkins home directory may have already failed due to open files held by class loaders. This PR solves the problem by moving the GC to before we delete the Jenkins home directory.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
+++ b/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
@@ -433,17 +433,17 @@ public abstract class HudsonTestCase extends TestCase implements RootAction {
             ExtensionList.clearLegacyInstances();
             DescriptorExtensionList.clearLegacyInstances();
 
-            try {
-                env.dispose();
-            } catch (Exception x) {
-                x.printStackTrace();
-            }
-
             // Jenkins creates ClassLoaders for plugins that hold on to file descriptors of its jar files,
             // but because there's no explicit dispose method on ClassLoader, they won't get GC-ed until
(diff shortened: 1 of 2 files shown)


basil:
In draft state until I open a PR demonstrating that with this change the test suite passes on Windows.

basil:
This PR was tested successfully in [a passing Windows CI run](https://ci.jenkins.io/job/Core/job/jenkins/job/PR-6024/25/) in jenkinsci/jenkins#6024.

src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
```diff
@@ -433,17 +433,17 @@ protected void tearDown() throws Exception {
...
-            try {
-                env.dispose();
-            } catch (Exception x) {
-                x.printStackTrace();
-            }
-
             // Jenkins creates ClassLoaders for plugins that hold on to file descriptors of its jar files,
             // but because there's no explicit dispose method on ClassLoader, they won't get GC-ed until
             // at some later point, leading to possible file descriptor overflow. So encourage GC now.
             // see http://bugs.sun.com/view_bug.do?bug_id=4950148
```
timja, line 439: fixes a broken link
```suggestion
            // see https://bugs.java.com/bugdatabase/view_bug.do?bug_id=4950148
```

timja (approved):
Interestingly it seems that the bug has been fixed for `URLClassLoader` as of Java 7, once switching to `URLClassLoader` we may not need this work around

PR #378, opened 2022-01-07
Adding JenkinsRule json methods for GET, POST and PUT that allows authentication using configured WebClient.

aHenryJard:
For testing plugin json API I needed:
- allow both getJSON with and without authentication
- allow to assert error code (to test that error management is correct)
- send PUT with json body too.
- adding a postJSON that can be used with authentication (token)

There is still the existing one postJSON, it's still working if you want to use the crumb instead (faster for testing, because this does not required to setup the security part).

+ tests added in JenkinsRuleTest


pom.xml
```diff
@@ -200,6 +200,25 @@ THE SOFTWARE.
       <version>1.0.2</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.kohsuke.stapler</groupId>
+      <artifactId>json-lib</artifactId>
```
jglick, line 205: Why? This ought to have been picked up from Jenkins core already.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -1270,6 +1270,32 @@ public Page search(String q) throws Exception {
...
     public JSONWebResponse getJSON(@NonNull String path) throws IOException {
+
+        JenkinsRule.WebClient webClient = createWebClient();
+        return getJSON(path, webClient);
+    }
+
+    public JSONWebResponse getJSON(@NonNull String path, @NonNull JenkinsRule.WebClient webClient) throws IOException {
+        assert !path.startsWith("/");
+
+        Page runsPage = null;
```
jglick, line 1281:
```suggestion
        Page runsPage;
```

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -1270,6 +1270,32 @@ public Page search(String q) throws Exception {
...
+
+        return new JSONWebResponse(webResponse);
+    }
+
+    /**
+     * Put JSON from A Jenkins endpoint.
+     * @param path The endpoint URL.
+     * @return The JSON.
+     */
+    public JSONWebResponse putJSON(@NonNull String path) throws IOException {
```
jglick, line 1298: How is this “put”? Still looks like a `GET` to me.

aHenryJard: I am not done coding at all, I did a draft to find why I cannot test it first with `simpleGetShouldWork`

pom.xml
aHenryJard: I am trying to fix my `//FIXME response is empty here :(` below, I was thinking that json parser is in this lib so I tried a new version, but it's not working anyway

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -1265,24 +1267,57 @@ public Page search(String q) throws Exception {
...
+    /**
+     * Send JSON content to a Jenkins relative endpoint.
+     * You can preconfigure webClient for example to set a token for authentication, or accept error HTTP status.
+     * @param path relative path, should not start with '/'
+     * @param json the json payload to send
+     * @param webClient a pre-configured web client.
+     * @return The JSON response from server.
+     * @throws IOException
+     */
+    public JSONWebResponse putJSON(@NonNull String path, @NonNull Object json,@NonNull JenkinsRule.WebClient webClient) throws IOException {
```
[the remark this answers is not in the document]
aHenryJard: I have imitate postJSON that has an object as json parameter, but I think it's better as you said to use JSON. Let me change it. Maybe postJSON should be updated later (to take authentication for example)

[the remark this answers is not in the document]
aHenryJard: done.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -1265,24 +1269,57 @@ public Page search(String q) throws Exception {
...
+     * @param path relative path, should not start with '/'
+     * @param json the json payload to send
+     * @param webClient a pre-configured web client.
+     * @return The JSON response from server.
+     * @throws IOException
+     */
+    public JSONWebResponse putJSON(@NonNull String path, @NonNull JSON json, @NonNull JenkinsRule.WebClient webClient) throws IOException {
+
+        URL URLtoCall = new URL(getURL(),path);
+        WebRequest putRequest = new WebRequest(URLtoCall, HttpMethod.PUT);
```
jglick, line 1316: Seems like a `POST` method would be more useful, as few Jenkins endpoints expect `PUT`?

jglick: https://github.com/jenkins-infra/jenkins.io/pull/4812#discussion_r787189376

aHenryJard: There is already a postJSON() method, and I wanted to add a version that allows to pass a webclient also but I couldn't make it work sadly. I can remove PUT from here if it's a blocker, so the get can be merge sooner if you want.

aHenryJard: update: I have found a way, I hope you will like it :)

src/test/java/org/jvnet/hudson/test/JenkinsRuleTest.java
```diff
@@ -127,6 +142,206 @@ public void testTokenHelperMethods() throws Exception {
...
+            JSONObject response = JSONObject.fromObject(body);
+            return new JsonHttpResponse(response, 200);
+        }
+
+        @POST
+        @WebMethod(name = "createFailure")
+        public JsonHttpResponse createFailure(@JsonBody MyJsonObject body) {
+            body.setMessage(body.getMessage()+" - NOT CREATED");
+            JSONObject response = JSONObject.fromObject(body);
+            return new JsonHttpResponse(response, 500);
```
kerogers-cloudbees, line 322: Just curious why this method returns the response and the other failure methods throw it.

aHenryJard: it's technically having the same behavior, but I prefer to throw when it's an exception, thanks for seeing it

aHenryJard: done.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
aHenryJard: done.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -1265,28 +1269,89 @@ public Page search(String q) throws Exception {
...
+     *
+     * @param path relative path, should not start with '/'
+     * @param json the json payload to send
+     * @param webClient a pre-configured web client either with crumb or token.
+     * @param user user to use for authentication
+     * @return The JSON response from server.
+     * @throws IOException
+     */
+    public JSONWebResponse postJSON(@NonNull String path, @NonNull JSON json, @NonNull WebClient webClient,
+                                    @NonNull User user) throws IOException {
```
Vlatombe, line 1338: Why not requesting a pre-configured webClient, and ask for the caller to do `webClient.withBasicApiToken(user)` ?

aHenryJard: because it will not work at all if the client is not pre-configured with withBasicApiToken() ; when for get it's working fine without token. So I found it misleading to just pass a webClient in this case, since having a User is mandatory.

aHenryJard: And I tried to fallback on crumb if there is no user, it's complicated to do so, because you need to check if a user token has been added and there is no existing method to tell "is the token set ?". Since the other postJSON works fine with crumb, I think this one should be used when authentication is needed, and the other method when you don't want to setup authentication on your test.

aHenryJard: and also if the crumb is set, then you cannot add a token anymore.

jglick: > having a User is mandatory

In a realistic installation perhaps, but plenty of functional tests do not bother setting up authentication & authorization, and just want to verify behavior of an endpoint unrelated to permission checks.

jglick: Now I see your point about a crumb being demanded in the case of an anonymous request, unless of course there is a `CrumbExclusion`. Still, clearer to have the test call one of the `withBasicApiToken` overloads.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -1265,28 +1269,89 @@ public Page search(String q) throws Exception {
...
-        }
-        WebResponse webResponse = runsPage.getWebResponse();
+        return getJSON(path, webClient);
+    }
+
+    /**
+     * Get JSON from a Jenkins relative endpoint.
+     * You can preconfigure webClient for example to set a token for authentication, or accept error HTTP status.
+     * @param path relative path, should not start with '/'
+     * @param webClient a pre-configured web client.
```
jglick, line 1288: Indicates that these should be instance methods on `WebClient`.

aHenryJard: I don't understand what you are asking here, is it a link to java classes ?

jglick: In general if you have a method

```java
class A {
  ReturnType someMethod(B b, String otherArg) {
    // does not use `this`
  }
}
class B {
}
```

that is a sign you should consider switching the receiver

```java
class A {
}
class B {
  ReturnType someMethod(String otherArg) {
    // uses `this`
  }
}
```

aHenryJard: it's not going to be a breaking change if I move them all ? Including the existing one ?

jglick: I was referring to new methods introduced in this PR, not existing methods in `JenkinsRule` with signatures

```java
public JSONWebResponse getJSON(String path);
public JSONWebResponse postJSON(String path, Object json);
```

src/test/java/org/jvnet/hudson/test/JenkinsRuleTest.java
```diff
@@ -127,6 +142,206 @@ public void testTokenHelperMethods() throws Exception {
...
+
+        // Testing a simple GET with parameter that should answer 200 OK and a json
+        response = j.getJSON("testing-cli/getWithParameters?paramValue=whatelse");
+        assertTrue(response.getContentAsString().contains("I am JenkinsRule whatelse"));
+        assertEquals(response.getStatusCode(), 200);
+
+        //Testing with a GET that the test expect to raise an server error: we want to be able to assert the status
+        JenkinsRule.WebClient webClientAcceptException = j.createWebClient();
+        webClientAcceptException.setThrowExceptionOnFailingStatusCode(false);
+        response = j.getJSON("testing-cli/getError500", webClientAcceptException);
```
jglick, line 161: For example https://github.com/jenkinsci/jenkins-test-harness/pull/378#discussion_r789919555 would mean that this could be simplified to

```suggestion
        response = j.createWebClient().withThrowExceptionOnFailingStatusCode(false).getJSON("testing-cli/getError500");
```

src/test/java/org/jvnet/hudson/test/JenkinsRuleTest.java
```diff
@@ -127,6 +142,206 @@ public void testTokenHelperMethods() throws Exception {
...
+
+        j.jenkins.setSecurityRealm(j.createDummySecurityRealm());
+        j.jenkins.setAuthorizationStrategy(auth);
+
+        // - simple call without authentication should be forbidden
+        response = j.getJSON("testing-cli/getMyJsonObject", webClientAcceptException);
+        assertEquals(response.getStatusCode(), 403);
+
+        // - same call but authenticated should be fine
+        response = j.getJSON("testing-cli/getMyJsonObject", webClientAcceptException.withBasicApiToken(admin));
```
jglick, line 177: Or this (which we do _not_ expect to throw an exception so why are we using `webClientAcceptException`?) would better be expressed as
```suggestion
        response = j.createWebClient().withBasicApiToken(admin).getJSON("testing-cli/getMyJsonObject");
```

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -179,6 +180,7 @@
 import java.util.logging.Level;
 import java.util.logging.LogRecord;
 import java.util.logging.Logger;
+import javax.annotation.Nonnull;
```
jglick, line 183: Deprecated, use the SpotBugs Annotations version (`@NonNull`).

src/test/java/org/jvnet/hudson/test/JenkinsRuleTest.java
```diff
@@ -127,6 +142,206 @@ public void testTokenHelperMethods() throws Exception {
...
+        assertEquals(response.getStatusCode(), 200);
+
+        //Testing with a GET that the test expect to raise an server error: we want to be able to assert the status
+        JenkinsRule.WebClient webClientAcceptException = j.createWebClient();
+        webClientAcceptException.setThrowExceptionOnFailingStatusCode(false);
+        response = j.getJSON("testing-cli/getError500", webClientAcceptException);
+        assertEquals(response.getStatusCode(), 500);
+
+        //Testing a GET that requires the user to be authenticated
+        User admin = User.getById("admin", true);
```
jglick, line 165: You do not actually need to do this; you can just use method overloads taking `String`. (Fine either way, just to make tests a bit more concise.)

src/test/java/org/jvnet/hudson/test/JenkinsRuleTest.java
```diff
@@ -127,6 +142,206 @@ public void testTokenHelperMethods() throws Exception {
...
+
+        //Testing a PUT that requires the user to be authenticated
+        User admin = User.getById("admin", true);
+        MockAuthorizationStrategy auth = new MockAuthorizationStrategy()
+                .grant(Jenkins.ADMINISTER).everywhere().to(admin);
+
+        j.jenkins.setSecurityRealm(j.createDummySecurityRealm());
+        j.jenkins.setAuthorizationStrategy(auth);
+
+        // - simple call without authentication should be forbidden
```
jglick, line 236: (Note distinction between `RootAction` and `UnprotectedRootAction`.)

jglick (changes requested):
In summary, as a matter of API design I would recommend

* moving new methods to use `WebClient` as the receiver rather than an argument
* drop `User` parameter from `postJSON`

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
aHenryJard: > In summary, as a matter of API design I would recommend
> 
>     * moving new methods to use `WebClient` as the receiver rather than an argument
> 
>     * drop `User` parameter from `postJSON`

Ok I think I get it now, thanks for detailed explanations. I wanted to imitate the existing method but it was no the best idea. I am going to rework on that.

aHenryJard:
> In summary, as a matter of API design I would recommend
> 
>     * moving new methods to use `WebClient` as the receiver rather than an argument
> 
>     * drop `User` parameter from `postJSON`

Ok I think I get it now, thanks for detailed explanations. I wanted to imitate the existing methods but it was no the best design. I am going to rework on that.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
aHenryJard: code moved

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
aHenryJard: done.

src/test/java/org/jvnet/hudson/test/JenkinsRuleTest.java
aHenryJard: yes to me the default should be protected (so RootAction) so I prefer to do example with this one.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -2777,6 +2771,75 @@ public Dim interactiveJavaScriptDebugger() {
...
+            putRequest.setAdditionalHeader("Content-Type", "application/json");
+            putRequest.setAdditionalHeader("Accept", "application/json");
+            putRequest.setAdditionalHeader("Accept-Encoding", "*");
+
+            return new JSONWebResponse(loadWebResponse(putRequest));
+        }
+
+        /**
+         * POST JSON content to a Jenkins relative endpoint using the user token for authentication.
+         * You can preconfigure webClient for example to accept error HTTP status before calling this method.
```
jglick, line 2821:
```suggestion
         * POST JSON content to a Jenkins relative endpoint.
         * You can preconfigure the web client for example to set a token for authenticaion or pass a crumb before calling this method.
```
?

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -2777,6 +2771,75 @@ public Dim interactiveJavaScriptDebugger() {
...
+            getRequest.setAdditionalHeader("Content-Type", "application/json");
+            getRequest.setAdditionalHeader("Accept", "application/json");
+            getRequest.setAdditionalHeader("Accept-Encoding", "*");
+
+            return new JSONWebResponse(loadWebResponse(getRequest));
+        }
+
+        /**
+         * Send JSON content to a Jenkins relative endpoint.
+         * You can preconfigure webClient for example to set a token for authentication, or accept error HTTP status
```
jglick, line 2798:
```suggestion
         * You can preconfigure the web client for example to set a token for authentication, or accept error HTTP status
```

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -2777,6 +2771,75 @@ public Dim interactiveJavaScriptDebugger() {
 
             return dim;
         }
+
+        /**
+         * Get JSON from a Jenkins relative endpoint.
+         * You can preconfigure webClient for example to set a token for authentication, or accept error HTTP status
```
jglick, line 2777:
```suggestion
         * You can preconfigure the web client for example to set a token for authentication, or accept error HTTP status
```

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -1265,28 +1267,20 @@ public Page search(String q) throws Exception {
...
-        }
-        WebResponse webResponse = runsPage.getWebResponse();
-
-        return new JSONWebResponse(webResponse);
+        return webClient.getJSON(path);
     }
 
     /**
-     * POST a JSON payload to a URL on the underlying Jenkins instance.
+     * POST a JSON payload to a URL on the underlying Jenkins instance using the crumb.
```
jglick, line 1283: Probably want to link to `WebClient.postJSON` as a more straightforward option, noting that it does not automatically pass a crumb (if there is no `CrumbExclusion` you can `createCrumbedURL` or just pass an API token). I am not sure what the other complexity in this old method is about; seems unnecessary.

jglick (approved):
Looks good to me. Javadoc could use a final pass.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
aHenryJard: Yes the old method seems a bit complex, but it would requires investigations on usages to change it IMO. I don't want to touch it :D

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -2777,6 +2771,75 @@ public Dim interactiveJavaScriptDebugger() {
...
+            putRequest.setAdditionalHeader("Content-Type", "application/json");
+            putRequest.setAdditionalHeader("Accept", "application/json");
+            putRequest.setAdditionalHeader("Accept-Encoding", "*");
+
+            return new JSONWebResponse(loadWebResponse(putRequest));
+        }
+
+        /**
+         * POST JSON content to a Jenkins relative endpoint.
+         * You can preconfigure the web client for example to set a token for authenticaion or pass a crumb before calling this method.
```
aHenryJard, line 2821:
```suggestion
         * You can preconfigure the web client for example to set a token for authentication or pass a crumb before calling this method.
```

PR #380, opened 2022-01-19
constrain an agents memory

jtnord:
Noticed when looking at the memory usage of pipeline-model-defintion
that the agents created in the tests did not have a constrained memory
set.  This means that the tests would be platform dependent. or flaky
depending on what GC happens at various times.

the number 512 is relatively arbitrary, but k8s agents use 256Mi as a
request and so 512 is double that, and the agent process is supposed to
be lightweight.

Whilst this could reveal a few issues in plugins when they upgrade, it
seems to me that that could be a bug in the plugin, and so I have not
implemented a way to create an online slave with a specific amount of
memory.  if that is indeed required it can be added later as jth updates
are not forced on plugins.

here is a before from a pipeline-model-definition test (nearly 3gb claimed for the heap...) (and there are 4 agents in this test!):

image

and after :

image

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
+++ b/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
@@ -685,7 +685,7 @@ public abstract class HudsonTestCase extends TestCase implements RootAction {
     public ComputerLauncher createComputerLauncher(EnvVars env) throws URISyntaxException, IOException {
         int sz = jenkins.getNodes().size();
         return new SimpleCommandLauncher(
-                String.format("\"%s/bin/java\" %s %s -jar \"%s\"",
+                String.format("\"%s/bin/java\" %s %s Xmx512M -jar \"%s\"",
                         System.getProperty("java.home"),
                         SLAVE_DEBUG_PORT>0 ? " -Xdebug -Xrunjdwp:transport=dt_socket,server=y,address="+(SLAVE_DEBUG_PORT+sz): "",
                         "-Djava.awt.headless=true",
--- a/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
+++ b/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
@@ -685,7 +685,7 @@ public abstract class HudsonTestCase extends TestCase implements RootAction {
     public ComputerLauncher createComputerLauncher(EnvVars env) throws URISyntaxException, IOException {
         int sz = jenkins.getNodes().size();
         return new SimpleCommandLauncher(
-                String.format("\"%s/bin/java\" %s %s Xmx512M -jar \"%s\"",
+                String.format("\"%s/bin/java\" %s %s -Xmx512M -jar \"%s\"",
                         System.getProperty("java.home"),
                         SLAVE_DEBUG_PORT>0 ? " -Xdebug -Xrunjdwp:transport=dt_socket,server=y,address="+(SLAVE_DEBUG_PORT+sz): "",
                         "-Djava.awt.headless=true",
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -1053,13 +1053,13 @@ public class JenkinsRule implements TestRule, MethodRule, RootAction {
      * Creates a launcher for starting a local agent.
      *
      * @param env
-     *      Environment variables to add to the slave process. Can be null.
+     *      Environment variables to add to the slave process. Can be {@code null}.
      */
     @NonNull
     public ComputerLauncher createComputerLauncher(@CheckForNull EnvVars env) throws URISyntaxException, IOException {
         int sz = jenkins.getNodes().size();
         return new SimpleCommandLauncher(
-                String.format("\"%s/bin/java\" %s %s -jar \"%s\"",
+                String.format("\"%s/bin/java\" %s %s -Xmx512M -jar \"%s\"",
                         System.getProperty("java.home"),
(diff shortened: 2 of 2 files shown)


src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
```diff
@@ -685,7 +685,7 @@ public PretendSlave createPretendSlave(FakeLauncher faker) throws Exception {
     public ComputerLauncher createComputerLauncher(EnvVars env) throws URISyntaxException, IOException {
         int sz = jenkins.getNodes().size();
         return new SimpleCommandLauncher(
-                String.format("\"%s/bin/java\" %s %s -jar \"%s\"",
+                String.format("\"%s/bin/java\" %s %s Xmx512M -jar \"%s\"",
```
timja, line 688: doesn't this need a `-` ?
```suggestion
                String.format("\"%s/bin/java\" %s %s -Xmx512M -jar \"%s\"",
```

jtnord: yes it does - good catch :)

raul-arabaolaza (approved):

jglick (approved):

basil:
I think a realistic test of this PR would involve the following:

1. (Regression test) Running the core test suite with the incremental from this PR
2. (Regression test) Running a plugin's test suite with the incremental from this PR and a `Jenkinsfile`, assuming that:
   a) The plugin has test cases that create 1-2 agents
   b) The plugin is running on our standard `buildPlugin(useContainerAgent: true)` setup
   c) The plugin's tests currently pass
3. (Fix test) Find a plugin with a test case that had to be disabled on Windows ACI container agents due to the fact that it created too many agents and used up too much memory (`ExtendedEmailPublisherMatrixTest` is a good one), enable the test on Windows ACI container agents to observe that it still fails, and then confirm that with this fix the test case now uses less memory and passes on Windows ACI container agents

jglick:
Yes we could do it that way rather than just waiting to see if there are any regressions and reverting if so. Any volunteers?

jtnord:
> Yes we could do it that way rather than just waiting to see if there are any regressions and reverting if so. Any volunteers?

I am creating some PRs

timja (approved):

timja:
> > Yes we could do it that way rather than just waiting to see if there are any regressions and reverting if so. Any volunteers?
> 
> I am creating some PRs

good luck while https://github.com/jenkins-infra/helpdesk/issues/2733 is a thing

jtnord:
CI build failed due to a timeout (which I hope is not my change but the artifactory issue) so there is no incremental to attempt to use at the moment.

jtnord:
I was thinking this would most likely break the maven-plugin as that does horrible things with maven in the agent process but we are in luck - it copies the [JTH code](https://github.com/jenkinsci/maven-plugin/blob/2f1a2ddf4a849301d594d1aa99aeb1e0cf5ec6d3/src/test/java/hudson/maven/MavenJenkinsRule.java#L38-L53) 🥳 
/cc @aheritier incase he want to make similar changes in the job.

jtnord:
@basil


(Regression test) Running the core test suite with the incremental from this PR

https://github.com/jenkinsci/jenkins/pull/6192 exists and has 2 failures - both have been observed before.
Also given https://github.com/jenkinsci/jenkins/pull/6198#pullrequestreview-858942948 that the previous release has been reverted there it is unclear how to demonstrate this further in jenkins core

> (Regression test) Running a plugin's test suite with the incremental from this PR and a Jenkinsfile, assuming that:

[pipeline-model-definition](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/master/Jenkinsfile) does that and there is a PR that fixes a lot of tests.

also given the extended issue with antifactory it is not clear if I can demonstrate this any further before getting a release?

timja:
Shouldn't we identify the cause of the regression and revert the fix forward before adding more changes to this module?

jtnord:
> Shouldn't we identify the cause of the regression and revert the fix forward before adding more changes to this module?

I assumed given Basil reverted it that he is working on it somewhere.  there are so few things in the change the only thing that looked applicable was the junit5 stuff.

the Jenkins build is also no longer happy regardless of the JTH version and given this does not seem to impact everyone, merging this whilst that investigation goes on should not be a blocker should it?  (well it should - nothing should be merged with a failing build - but that does not seem to be the case for jenkins core at the moment)

basil:
> it is unclear how to demonstrate this further in jenkins core given the current state.

You could demonstrate this PR in core by rebasing it on top of the older version of `jenkins-test-harness` that core is currently using.

You have also not addressed my third point about Windows ACI tests like `ExtendedEmailPublisherMatrixTest`.

> Shouldn't we identify the cause of the regression and revert the fix forward before adding more changes to this module?

Agreed. The reason `jenkins-test-harness` is in a bad state is that previous PRs were merged without sufficient testing. The bad state should be corrected by reverting the problematic PR, and we should avoid getting into a bad state going forward by doing more thorough testing of PRs (like this one) prior to merge.

> I assumed given Basil reverted it that he is working on it somewhere.

Do not make assumptions about what I am working on. I am _not_ working on stabilizing `jenkins-test-harness`.

jtnord:
> You have also not addressed my third point about Windows ACI tests like ExtendedEmailPublisherMatrixTest.


sure see https://github.com/jenkinsci/jenkins-test-harness/pull/jenkinsci/pipeline-model-definition-plugin#480

jtnord:
> You could demonstrate this PR in core by rebasing it on top of the older version of jenkins-test-harness that core is currently using.

I could but I am not going to.

> Agreed. The reason jenkins-test-harness is in a bad state is that previous PRs were merged without sufficient testing. The bad state should be corrected by reverting the problematic PR, and we should avoid getting into a bad state going forward by doing more thorough testing of PRs (like this one) prior to merge.

people in glass houses should not throw stones.

olamy (approved):

PR #375, opened 2021-12-31
[JENKINS-48466] Provide JUnit 5 support for JenkinsRule

reda-alaoui:
This PR reapplies https://github.com/jenkinsci/jenkins-test-harness/pull/303 (reverted by https://github.com/jenkinsci/jenkins-test-harness/pull/317).

According to what we learned in https://github.com/jenkins-infra/pipeline-library/pull/272 , this PR fixes the original one by moving `junit-vintage-engine` from scope `test` to scope `compile` to make it available to all dependent projects.

----------------------------------------------------------------------------------------------------------------------------

Provides JUnit 5 support for JenkinsRule as discussed in JENKINS-48466.

Tests written with JUnit 5 don't support rules, thus require code like this to work with the test harness.

Usage example:

```java
// JUnit 5
@ExtendWith(JenkinsRuleResolver.class)
class ExampleJUnit5Test {

    @Test
    public void example(JenkinsRuleExtension r) {
        // use 'r' as usual
        // ...
    }

}
```

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [X] Link to relevant issues in GitHub or Jira
- [X] Link to relevant pull requests, esp. upstream and downstream changes
- [X] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #381, opened 2022-01-24
Revert "[JENKINS-48466] Provide JUnit 5 support for JenkinsRule"

jtnord:
Reverts jenkinsci/jenkins-test-harness#375

Cause of the instability in Jenkins unit tests reported by @basil in jenkinsci/jenkins#6198  and jenkinsci/jenkins#6195

Before this PR running `mvn test -Dtest=QueueTest#flyweightTasksWithoutMasterExecutors` would fail (generally due to a timeout) withi approximately 10 consequtive runs locally (windows)(instability was also reported for builds on ci.jenkins.io).

With this change the same command has been run 47 times before failing, and consistently above 40 iterations.
This is on a par with `1682.v4a_71b_f124ca_3` (actually better as 1682 would fail after 24 consecutive runs)

No further checking has been performed as to determine which part of the PR was causing the flakyness, nor do I intend to perform such analysis, the primary cause of this PR is to unblock other changes in the test suite as noted in #380


jtnord:
before this running `mvn test -Dtest=QueueTest#flyweightTasksWithoutMasterExecutors` would fail (generally due to a timeout) after approximately 10 runs locally (instability was also reported for builds on ci.jenkins.io).

with this change the same command has been run 16 times (and counting) in a row without failure

timja (approved):
Weird can’t see anything terribly suspicious in the PR that would cause it

jtnord:
> Weird can’t see anything terribly suspicious in the PR that would cause it

yup, the tests are still flaky (they are flaky in the current version in master which is before this change, just they are less flaky than with the junit5 work. 

Maybe this only a trigger due to memory and the fact that the test is creating a [`DumbSlave`](https://github.com/jenkinsci/jenkins-test-harness/blob/fcb959006c07922db0bb11b32452a074f22f9ec2/src/main/java/hudson/slaves/DummyCloudImpl.java#L123) and the few extra listeners are tipping the memory (however visualvm did not look like the surefire vm was close to being out of memory). 

possibly deserves some more digging into.
when the test fails, provisioning is not called.  why that would change with this PR is anyone's guess, it would seem very strange - to the point that I am thinking memory...
either that or all this is just pure fluke.

jtnord:
compare https://ci.jenkins.io/job/Core/job/jenkins/view/change-requests/job/PR-6198/  (dependabot update to the latest) which is a sea of instability and https://ci.jenkins.io/job/Core/job/jenkins/job/PR-6210/ (core with this PR) that had 2 successful runs in a row.

So even if I picked the wrong test ci is stable with this revert, but not with the release that contains the code we are reverting and the parent bump -> pointing strongly at the junit5 work that is reverted here.

I am not planning on doing more investigation into why the junit5 change made the tests more flaky, but I may look into why the core tests are still somewhat (less) flaky without it.

jglick:
> I am not planning on doing more investigation into why the junit5 change made the tests more flaky

Agreed, this makes sense as a hotfix. Not sure but perhaps inclusion of new libraries into the test classpath causes Surefire to automatically switch modes?

jtnord:
> Is there some way to provide the JUnit 5 support without getting into core, eg. an optional or opt-in component?

it can always be done in a separate repo that gets pulled in by plugins that want this functionality (or c&p) until someone works out junit5 caused the tests to be flaky in the `test` module in `jenkins`

offa:
I see, but nothing as part of this project ...

jtnord:
> I see, but nothing as part of this project ...

the ideal solution would be to fix the cause of the instability and then revert this.

jglick:
https://github.com/jenkinsci/jenkins-test-harness/releases/tag/1692.v0a_3884e172a_4

jglick:
> any consumer that wanted junit-5 would have to opt in and provide the dependencies themselves

Seems reasonable given that Jenkins core and plugins normally use JUnit 4 and there is no particular driver to migrate (that I am aware of).

PR #393, opened 2022-02-16
Remove unneeded injected test

timja:
As of https://github.com/jenkinsci/jenkins/pull/5923 `f:radio` uses `for=` to attach labels by generating a unique ID per input.

[Theme manager](https://github.com/jenkinsci/theme-manager-plugin/pull/75) is using a similar (completely safe) method.

Can we remove this test please given core is using this pattern?

One of the annoying parts of `InjectedTest` is you can't selectively disable it -.-


jglick:
I know nothing about this assertion or the change that allegedly made it incorrect.

uhafner (approved):
I think it is fine to remove this test when it does not make sense in all cases.

PR #399, opened 2022-03-08
Restore working version of `MemoryAssert#assertGC`

### LINKED ISSUE #360: Test coverage for `MemoryAssert.assertGC`
Does not look like this has any in-repo test coverage, which could be a problem if an `INSANE` update breaks usage: https://github.com/jenkinsci/jenkins-test-harness/pull/358#issuecomment-991173329

basil:
Reverts #397 and adds test coverage for `MemoryAssert#assertGC`. The new test fails without the revert of #397 and passes with the revert of #397.

Fixes #360

### DIFF
--- a/src/test/java/org/jvnet/hudson/test/MemoryAssertTest.java
+++ b/src/test/java/org/jvnet/hudson/test/MemoryAssertTest.java
@@ -25,8 +25,14 @@
 package org.jvnet.hudson.test;
 
 import java.lang.ref.WeakReference;
+import static org.junit.Assume.assumeTrue;
 import static org.jvnet.hudson.test.MemoryAssert.*;
 import static org.junit.Assert.*;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import hudson.util.VersionNumber;
 import org.junit.Test;
 
 public class MemoryAssertTest {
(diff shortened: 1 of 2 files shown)


timja (approved):

src/test/java/org/jvnet/hudson/test/MemoryAssertTest.java
```diff
@@ -46,4 +52,22 @@
         assertTrue(e.toString(), e.getMessage().contains("3000"));
     }
 
+    @Test
+    public void gc() {
+        // TODO JENKINS-67974 does not work on Java 9+
+        assumeTrue(new VersionNumber(System.getProperty("java.specification.version")).isOlderThan(new VersionNumber("9")));
```
jglick, line 58:
```suggestion
        assumeTrue("TODO JENKINS-67974 does not work on Java 9+", new VersionNumber(System.getProperty("java.specification.version")).isOlderThan(new VersionNumber("9")));
```

jglick (approved):
Thanks! What fails in `RELEASE130`? When I have a moment I can try to bisect and find the upstream culprit.

jglick:
Never mind, just saw https://github.com/jenkinsci/jenkins-test-harness/pull/397#issuecomment-1061339056.

jglick:
See https://github.com/apache/netbeans/pull/3206 and https://issues.jenkins.io/browse/JENKINS-67974. Unlikely to be able to update without a lot of hassle. On Java 8 we could patch INSANE to fall back to `AccessibleObject.setAccessible` when `MakeAccessible` is not in the classpath, but would be more complicated on Java 11.

INSANE has a number of limitations anyway. Would be possible to replace it with a new library that uses JMX to capture a `*.hprof` heap dump of the current JVM and then you can use e.g. the NetBeans Profiler library to analyze that. No idea whether heap dump capture is straightforward on Java 11 either.

PR #415, opened 2022-03-30
Allow `RealJenkinsRule` to use a specific WAR file

### LINKED ISSUE #414: Allow RealJenkinsRule to use a war file
### What feature do you want to see added?

RealJenkinsRule currently use the war in the classpath or `war/target/jenkins.war`. That war is reused by all rules in the test. To be able to use different wars per rule, a new `withWar` method has to be added.

Acceptance criteria:

* The war file can be configure as part of the RealJenkinsRule configuration.
* The dependencies from the pom.xml might be excluded easily.

### Upstream changes

_No response_

escoem:
Fixes #414

Instead of using the war on the path defined as part of the pom, or the one at war/target/jenkins.war, or using the system property affecting all tests, I can now create a RealJenkinsRule using a war (ie. a war built as part of the workflow). 

```java
@Rule
public RealJenkinsRule sc = new RealJenkinsRule()
                .javaOptions("-Xmx512M")
                .withHost("sc.127.0.0.1.nip.io")
                .withWar(new File("/path/to/my/own/jenkins.war"));
```

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -246,10 +246,7 @@ public final class RealJenkinsRule implements TestRule {
      * @param war The war file to be used.
      */
     public RealJenkinsRule withWar(File war) {
-        if (war != null && war.isFile()) {
-            this.war = war;
-        }
-
+        this.war = war;
         return this;
     }
 
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -144,6 +144,16 @@ public final class RealJenkinsRule implements TestRule {
      */
     private int port;
 
+    /**
+     * The war file to be used instead of the one in the path or war/target/jenkins.war if it is core build.
+     */
+    private File war;
+
+    /**
+     * If true, all plugins from pom will be skipped. Extra plugins are not ignored.
+     */
+    private boolean skipAllPlugins = false;
+
     private final String token = UUID.randomUUID().toString();
 
     private final Set<String> extraPlugins = new TreeSet<>();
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -144,6 +144,16 @@
...
 
+    /**
+     * The war file to be used instead of the one in the path or war/target/jenkins.war if it is core build.
+     */
+    private File war;
+
+    /**
+     * If true, all plugins from pom will be skipped. Extra plugins are not ignored.
+     */
+    private boolean skipAllPlugins = false;
```
fcojfernandez, line 155: isn't this changing the original default behaviour because of https://github.com/jenkinsci/jenkins-test-harness/pull/415/files#diff-ade203da1dc447eff46cf045f9a21d8ac019ccffdae26359a6838a54a1ac48fbR283 ? Right now the `RealJenkinsRule` is executing all that in this PR is within the `if (skipAllPlugins)`. If by default `skipAllPlugins` is false, then that code won't be executed unless the method `skipAllPlugins` is called with `true` as parameter

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -252,68 +279,71 @@ public RealJenkinsRule withHost(String host) {
...
-                            }
-                            if (Files.exists(snapshotManifest)) {
-                                String shortName;
-                                try (InputStream is = Files.newInputStream(snapshotManifest)) {
-                                    shortName = new Manifest(is).getMainAttributes().getValue("Short-Name");
-                                }
-                                if (shortName == null) {
-                                    throw new IOException("malformed " + snapshotManifest);
+
+                    if (skipAllPlugins) {
```
fcojfernandez, line 283: because my previous comment, shouldn't be the opposite?
```suggestion
                    if (!skipAllPlugins) {
```

fcojfernandez (approved):

escoem:
@jglick maybe can review it too

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -144,6 +144,16 @@
      */
     private int port;
 
+    /**
+     * The war file to be used instead of the one in the path or war/target/jenkins.war if it is core build.
+     */
```
jglick, line 149: redundant, at most `@see` the quasi-setter

jglick (approved):
Minor suggested Javadoc edits, but otherwise looks good to go.

jglick (approved):
Looks right. Before merging, can you confirm that this has been successfully prototyped in a non-plugin repo?

escoem:
> Looks right. Before merging, can you confirm that this has been successfully prototyped in a non-plugin repo?

yes :-)

PR #417, opened 2022-03-31
[JENKINS-67974] `assertGC` works on Java 8 and 17 but not 11

basil:
The following tests (when run against tip-of-trunk core, Maven HPI plugin, and test harness) fail when run with Java 11, but not when run with Java 8 or Java 17:

```
mvn clean verify -Dtest=org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.GroovyMemoryLeakTest
mvn clean verify -Dtest=org.jenkinsci.plugins.workflow.libs.LibraryMemoryTest
mvn clean verify -Dtest=org.jenkinsci.plugins.workflow.cps.CpsFlowExecutionMemoryTest
mvn clean verify -Dtest=org.jenkinsci.plugins.workflow.job.MemoryCleanupTest
```

I examined the test output from the Java 17 run and saw that it looked identical to the Java 8 run, so I think the Java 17 run really is doing the proper testing and not just passing accidentally. The Java 17 run is using the latest version of INSANE with the INSANE hooks.

I am not sure why the Java 11 run is failing, but I am not inclined to spend much time on it if things are working on Java 8 and Java 17. Rather I think we should just skip the test on Java 11.

Interestingly enough, the test coverage for `assertGC` in this repository does pass on Java 11, and the `MemoryAssert#assertGC`-based tests in core do pass on Java 11, so I think `assertGC` is actually somewhat working on Java 11, just not in all cases. Java 17 has worked in all cases that I tested.

So rather than skipping `assertGC` on Java 11 in all cases, let's just skip it on Java 11 in the cases where it fails. That is what this PR does.

To test this, I ran the 4 tests described above with Java 8, 11, and 17. Before, 8 and 17 passed and 11 failed. Now, 11 is skipped after printing "[WARNING] Corrupted STDOUT by directly writing to native stream in forked JVM 1"; 8 and 17 still pass.

The tests that call `assertGC` in this repository and in core passed in Java 8, 11, and 17 before this PR and continue to pass after this PR.

Unfortunately I am not sure how the test coverage in this repository could be improved.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/MemoryAssert.java
+++ b/src/main/java/org/jvnet/hudson/test/MemoryAssert.java
@@ -155,8 +155,12 @@ public class MemoryAssert {
      * @param allowSoft if true, pass even if {@link SoftReference}s apparently needed to be cleared by forcing an {@link OutOfMemoryError};
      *                  if false, fail in such a case (though the failure will be slow)
      */
-    @SuppressFBWarnings({"DLS_DEAD_LOCAL_STORE_OF_NULL", "DM_GC"})
+    @SuppressFBWarnings("DLS_DEAD_LOCAL_STORE_OF_NULL")
     public static void assertGC(WeakReference<?> reference, boolean allowSoft) {
+        VersionNumber javaVersion = new VersionNumber(System.getProperty("java.specification.version"));
+        assumeTrue(
+                "TODO JENKINS-67974 works on Java 8 and 17 but not 11",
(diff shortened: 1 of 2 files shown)


jglick (approved):
Fine until someone has the time to dig into it.

Should this really be `internal`? Seems like `bug` or `enhancement` to me, unless the last-released version does not support 11 at all?

basil:
I suppose not. If it's causing internal tests to fail let's just skip `assertGC` unconditionally on Java 11. This will be a good motivation to get Java 17 builds up and running.

PR #422, opened 2022-04-08
Run RealJenkinsRule steps on a separate thread to avoid unintended interaction with the request for doStep

dwnusbaum:
I recently ran into a `RealJenkinsRule` test that called code that used `Stapler.getCurrentRequest`, which ended up failing with a confusing error message because `Stapler.getCurrentRequest` returned the request for `RealJenkinsRule.doStep` instead of `null` as would typically be the case in a test. This PR does two things to improve this case:

1. It includes the response body of `RealJenkinsRule.doStep` in the error thrown by `RealJenkinsRule.runRemotely` if `doStep` returns a non-200 response to make it easier to diagnose errors.
2. It runs the steps passed to `RealJenkinsRule.runRemotely` on a separate thread so that they cannot accidentally interact with the HTTP request for `RealJenkinsRule.doStep` if they use `Stapler.getCurrentRequest`.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -29,6 +29,8 @@ import hudson.model.UnprotectedRootAction;
 import hudson.security.ACL;
 import hudson.security.ACLContext;
 import hudson.security.csrf.CrumbExclusion;
+import hudson.util.DaemonThreadFactory;
+import hudson.util.NamingThreadFactory;
 import hudson.util.StreamCopyThread;
 import java.io.BufferedReader;
 import java.io.File;
@@ -62,6 +64,9 @@ import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.UUID;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.jar.JarInputStream;
 import java.util.jar.Manifest;
@@ -657,16 +662,30 @@ public final class RealJenkinsRule implements TestRule {
             System.err.println("Checking status");
             checkToken(token);
         }
+        /**
+         * Used to run test methods on a separate thread so that code that uses {@link Stapler.getCurrentRequest}
+         * does not inadvertently interact with the request for {@link #doStep} itself.
+         */
+        private static final ExecutorService STEP_RUNNER = Executors.newSingleThreadExecutor(
+                new NamingThreadFactory(new DaemonThreadFactory(), RealJenkinsRule.class.getName() + ".STEP_RUNNER"));
         @POST
         public void doStep(StaplerRequest req, StaplerResponse rsp) throws Throwable {
             List<?> tokenAndStep = (List<?>) Init2.readSer(req.getInputStream(), Endpoint.class.getClassLoader());
             checkToken((String) tokenAndStep.get(0));
             Step s = (Step) tokenAndStep.get(1);
             Throwable err = null;
-            try (CustomJenkinsRule rule = new CustomJenkinsRule(); ACLContext ctx = ACL.as(ACL.SYSTEM)) {
-                s.run(rule);
-            } catch (Throwable t) {
-                err = t;
+            try {
+                STEP_RUNNER.submit(() -> {
+                    try (CustomJenkinsRule rule = new CustomJenkinsRule(); ACLContext ctx = ACL.as(ACL.SYSTEM)) {
+                        s.run(rule);
+                    } catch (Throwable t) {
+                        throw new RuntimeException(t);
+                    }
+                }).get();
+            } catch (ExecutionException e) {
+                err = e.getCause().getCause();
+            } catch (InterruptedException e) {
+                err = e;
             }
             // TODO use raw err if it seems safe enough
             Init2.writeSer(rsp.getOutputStream(), err != null ? new ProxyException(err) : null);
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -553,11 +553,13 @@ public final class RealJenkinsRule implements TestRule {
         try {
             error = (Throwable) Init2.readSer(conn.getInputStream(), null);
         } catch (IOException e) {
-            try {
-                String errorMessage = IOUtils.toString(conn.getErrorStream(), StandardCharsets.UTF_8);
-                e.addSuppressed(new IOException("Response body: " + errorMessage));
-            } catch (IOException e2) {
-                e.addSuppressed(e2);
+            if (conn.getErrorStream() != null) {
+                try {
+                    String errorMessage = IOUtils.toString(conn.getErrorStream(), StandardCharsets.UTF_8);
+                    e.addSuppressed(new IOException("Response body: " + errorMessage));
+                } catch (IOException e2) {
+                    e.addSuppressed(e2);
+                }
             }
             error = e;
         }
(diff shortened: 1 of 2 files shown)


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -544,7 +549,18 @@ public void runRemotely(Step s) throws Throwable {
         conn.setRequestProperty("Content-Type", "application/octet-stream");
         conn.setDoOutput(true);
         Init2.writeSer(conn.getOutputStream(), Arrays.asList(token, s));
-        Throwable error = (Throwable) Init2.readSer(conn.getInputStream(), null);
+        Throwable error;
+        try {
+            error = (Throwable) Init2.readSer(conn.getInputStream(), null);
+        } catch (IOException e) {
```
dwnusbaum, line 555: Note that if we keep the changes in `doStep` to run steps on a separate thread, I think this would really only happen if `doStep` itself had a problem. Still, it seems relatively harmless as a "just in case" kind of thing.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -646,16 +662,30 @@ public void doStatus(@QueryParameter String token) {
             System.err.println("Checking status");
             checkToken(token);
         }
+        /**
+         * Used to run test methods on a separate thread so that code that uses {@link Stapler.getCurrentRequest}
+         * does not inadvertently interact with the request for {@link #doStep} itself.
+         */
+        private static final ExecutorService STEP_RUNNER = Executors.newSingleThreadExecutor(
+                new NamingThreadFactory(new DaemonThreadFactory(), RealJenkinsRule.class.getName() + ".STEP_RUNNER"));
```
dwnusbaum, line 670: Not sure whether the thread factories matter here. Should we use `ClassLoaderSanityThreadFactory`? Do we even care if this is a daemon thread or not?

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -646,16 +662,30 @@ public void doStatus(@QueryParameter String token) {
             System.err.println("Checking status");
             checkToken(token);
         }
+        /**
+         * Used to run test methods on a separate thread so that code that uses {@link Stapler.getCurrentRequest}
+         * does not inadvertently interact with the request for {@link #doStep} itself.
```
dwnusbaum, line 667: I would be fine reverting this part of the PR if anyone has any concerns that are not easy to address.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -646,16 +662,30 @@ public void doStatus(@QueryParameter String token) {
...
-            } catch (Throwable t) {
-                err = t;
+            try {
+                STEP_RUNNER.submit(() -> {
+                    try (CustomJenkinsRule rule = new CustomJenkinsRule(); ACLContext ctx = ACL.as(ACL.SYSTEM)) {
+                        s.run(rule);
+                    } catch (Throwable t) {
+                        throw new RuntimeException(t);
+                    }
+                }).get();
```
dwnusbaum, line 684: TODO: I did not check whether the test timing out is handled appropriately here. `e.getCause().getCause()` might throw an NPE in that case.

dwnusbaum: From some testing, this process just get killed if the timeout is exceeded, so it doesn't matter whether we call `get(long, TimeUnit)`. There are a few superfluous "Connection already closed" error messages if you do hit the timeout, but they existed before my PR as well.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -544,7 +549,20 @@ public void runRemotely(Step s) throws Throwable {
...
+        } catch (IOException e) {
+            if (conn.getErrorStream() != null) {
+                try {
+                    String errorMessage = IOUtils.toString(conn.getErrorStream(), StandardCharsets.UTF_8);
+                    e.addSuppressed(new IOException("Response body: " + errorMessage));
+                } catch (IOException e2) {
+                    e.addSuppressed(e2);
+                }
+            }
+            error = e;
```
jglick, line 564: or just
```suggestion
            throw e;
```

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -646,16 +664,30 @@ public void doStatus(@QueryParameter String token) {
             System.err.println("Checking status");
             checkToken(token);
         }
+        /**
+         * Used to run test methods on a separate thread so that code that uses {@link Stapler.getCurrentRequest}
```
jglick, line 668: either
```suggestion
         * Used to run test methods on a separate thread so that code that uses {@code Stapler.getCurrentRequest}
```
or
```suggestion
         * Used to run test methods on a separate thread so that code that uses {@link Stapler#getCurrentRequest}
```
after `import`ing it.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jglick: > Do we even care if this is a daemon thread or not?

Likely not.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -646,16 +664,30 @@ public void doStatus(@QueryParameter String token) {
...
+            try {
+                STEP_RUNNER.submit(() -> {
+                    try (CustomJenkinsRule rule = new CustomJenkinsRule(); ACLContext ctx = ACL.as(ACL.SYSTEM)) {
+                        s.run(rule);
+                    } catch (Throwable t) {
+                        throw new RuntimeException(t);
+                    }
+                }).get();
+            } catch (ExecutionException e) {
+                err = e.getCause().getCause();
```
jglick, line 688:
```suggestion
                // Once for ExecutionException, once for RuntimeException:
                err = e.getCause().getCause();
```
May be more robust and clearer to use an `AtomicReference<Throwable>` though.

jglick (approved):
A couple nits but looks good.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
dwnusbaum: Good catch, I fixed this in [869ed42](https://github.com/jenkinsci/jenkins-test-harness/pull/422/commits/869ed42113bfc9b6f3d3a4f70e26fa1841e7fc3e).

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
dwnusbaum: Done in [869ed42](https://github.com/jenkinsci/jenkins-test-harness/pull/422/commits/869ed42113bfc9b6f3d3a4f70e26fa1841e7fc3e).

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
dwnusbaum: Switched to a regular thread pool in [869ed42](https://github.com/jenkinsci/jenkins-test-harness/pull/422/commits/869ed42113bfc9b6f3d3a4f70e26fa1841e7fc3e).

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
dwnusbaum: I added a comment in [869ed42](https://github.com/jenkinsci/jenkins-test-harness/pull/422/commits/869ed42113bfc9b6f3d3a4f70e26fa1841e7fc3e). I experimented with `AtomicReference<Throwable>` but it was less clear in my opinion.

jglick (approved):

PR #426, opened 2022-04-22
New `configRoundtrip` method for `Cloud` configuration

sparsick:
In Jenkins plugin [azure-container-plugin](https://github.com/jenkinsci/azure-container-agents-plugin/blob/1bf31f31ad77524b0699e4c6266cb709eef20855/src/test/java/com/microsoft/jenkins/containeragents/AciCloudConfigTest.java#L26), there is a use case to test the correctness of the cloud configuration. The JenkinsRule has no configRoundtrip method for the cloud configuration. 

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -105,6 +105,7 @@ import hudson.security.ACL;
 import hudson.security.AbstractPasswordBasedSecurityRealm;
 import hudson.security.GroupDetails;
 import hudson.security.csrf.CrumbIssuer;
+import hudson.slaves.Cloud;
 import hudson.slaves.ComputerConnector;
 import hudson.slaves.ComputerLauncher;
 import hudson.slaves.DumbSlave;
@@ -1424,6 +1425,13 @@ public class JenkinsRule implements TestRule, MethodRule, RootAction {
     public <V extends View> V configRoundtrip(V view) throws Exception {
         submit(createWebClient().getPage(view, "configure").getFormByName("viewConfig"));
         return view;
+    }
+
+    public <C extends Cloud> C configRoundtrip (C cloud) throws Exception {
+        jenkins.clouds.add(cloud);
+        jenkins.save();
+        submit(createWebClient().goTo("configureClouds/").getFormByName("config"));
+        return (C)jenkins.getCloud(cloud.name);
     }
 
 
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -1427,7 +1427,7 @@ public class JenkinsRule implements TestRule, MethodRule, RootAction {
         return view;
     }
 
-    public <C extends Cloud> C configRoundtrip (C cloud) throws Exception {
+    public <C extends Cloud> C configRoundtrip(C cloud) throws Exception {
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -1426,6 +1427,13 @@ public <V extends View> V configRoundtrip(V view) throws Exception {
         return view;
     }
 
+    public <C extends Cloud> C configRoundtrip (C cloud) throws Exception {
```
jglick, line 1430:
```suggestion
    public <C extends Cloud> C configRoundtrip(C cloud) throws Exception {
```

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -1426,6 +1427,13 @@ public <V extends View> V configRoundtrip(V view) throws Exception {
...
+        jenkins.clouds.add(cloud);
```
jglick, line 1431: Does this replace any existing `Cloud` of the same `name`?

jglick (approved):
Looks right. To prove that the API can be (and will be) used as written, you should create a draft PR in whatever plugin calling this method. (Include the URL of this PR in the **Description** to create a backlink.) Look up the incremental version number of this PR (see the **Checks** tab) and in your draft plugin PR add to `<properties>` in `pom.xml`

```xml
<jenkins-test-harness.version>1234.vwhatever</jenkins-test-harness.version> <!-- TODO https://github.com/jenkinsci/jenkins-test-harness/pull/426 -->
```

If your plugin has not yet been “incrementalified” then also

```bash
mkdir .mvn
echo -Pconsume-incrementals > .mvn/maven.config
git add .mvn
```

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
sparsick: Under the hood, the cloud object is stored in a `ArrayList` and it should not overwrite an already existing element in the list.

jglick: Then I would suggest one of
* checking for an existing `Cloud` of the same `name` and deleting it first, so you are actually replacing it
* removing this `add` & `save`, so the caller (test case) is responsible for registering the `Cloud` initially
* documenting the current behavior, since if I recall correctly it differs from that of other `configRoundtrip` overloads

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -1427,7 +1428,17 @@ public <V extends View> V configRoundtrip(V view) throws Exception {
         return view;
     }
 
-    public <C extends Cloud> C configRoundtrip (C cloud) throws Exception {
+    /**
+     * Performs a configuration round-trip testing for a cloud.
+     * The given cloud is added to the cloud list of Jenkins.
+     *
```
jglick, line 1434:
```suggestion
     * <p>
```
(When you are used to Markdown, it is hard to remember that Javadoc is just HTML!)

sparsick: Sorry :) and thank you

jglick:
(just awaiting downstream PR)

sparsick:
Downstream PR is created https://github.com/jenkinsci/azure-container-agents-plugin/pull/132

PR #425, opened 2022-04-19
Print flags specified on command line or set by ergonomics

basil:
Sample output below. This information can be essential when reasoning about memory usage by multiple JVMs within a single VM or container.

Java 8:

```
$ "/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java"  -Djava.awt.headless=true -Xmx512m -XX:+PrintCommandLineFlags -jar "/home/basil/src/jenkinsci/text-finder-plugin/target/jenkins-for-test/WEB-INF/lib/remoting-4.5.jar"
-XX:InitialHeapSize=523530944 -XX:MaxHeapSize=536870912 -XX:+PrintCommandLineFlags -XX:+UseCompressedClassPointers -XX:+UseCompressedOops -XX:+UseParallelGC 
<===[JENKINS REMOTING CAPACITY]===>channel started
Remoting version: 4.5
This is a Unix agent
Evacuated stdout
```

Java 11:

```
$ "/usr/lib/jvm/java-11-openjdk-amd64/bin/java"  -Djava.awt.headless=true -Xmx512m -XX:+PrintCommandLineFlags -jar "/home/basil/src/jenkinsci/text-finder-plugin/target/jenkins-for-test/WEB-INF/lib/remoting-4.5.jar"
-XX:G1ConcRefinementThreads=8 -XX:GCDrainStackTargetSize=64 -XX:InitialHeapSize=523530944 -XX:MaxHeapSize=536870912 -XX:+PrintCommandLineFlags -XX:ReservedCodeCacheSize=251658240 -XX:+SegmentedCodeCache -XX:+UseCompressedClassPointers -XX:+UseCompressedOops -XX:+UseG1GC 
<===[JENKINS REMOTING CAPACITY]===>channel started
Remoting version: 4.5
This is a Unix agent
```

Java 17:

```
$ "/usr/lib/jvm/java-17-openjdk-amd64/bin/java"  -Djava.awt.headless=true -Xmx512m -XX:+PrintCommandLineFlags -jar "/home/basil/src/jenkinsci/text-finder-plugin/target/jenkins-for-test/WEB-INF/lib/remoting-4.13.jar"
[...]

PR #428, opened 2022-05-02
[JENKINS-26398] More flexible versions of `createSlave`

jglick:
[JENKINS-26398](https://issues.jenkins.io/browse/JENKINS-26398)

Inspired by changes being made in https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/180. Cleans up some tech debt from https://github.com/jenkinsci/pipeline-plugin/pull/34 & https://github.com/jenkinsci/credentials-binding-plugin/pull/3.

* Fixes `createSlave` overloads to use a more stable agent workdir compatible with `JenkinsSessionRule`.
* Introduces `InboundAgentRule` which is better suited to simulating agent disconnections.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
+++ b/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
@@ -35,7 +35,10 @@ import hudson.slaves.JNLPLauncher;
 import hudson.slaves.RetentionStrategy;
 import hudson.util.StreamCopyThread;
 import java.io.File;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import org.apache.tools.ant.util.JavaEnvUtils;
 import org.junit.rules.ExternalResource;
@@ -74,7 +77,18 @@ public final class InboundAgentRule extends ExternalResource {
     @SuppressFBWarnings(value = "COMMAND_INJECTION", justification = "just for test code")
     public void start(@NonNull JenkinsRule r, String name) throws Exception {
         stop(name);
-        ProcessBuilder pb = new ProcessBuilder(JavaEnvUtils.getJreExecutable("java"), "-Djava.awt.headless=true", "-jar", Which.jarFile(Launcher.class).getAbsolutePath(), "-jnlpUrl", r.getURL() + "computer/" + name + "/slave-agent.jnlp");
+        List<String> cmd = new ArrayList<>(Arrays.asList(JavaEnvUtils.getJreExecutable("java"),
+            "-Xmx512m",
+            "-XX:+PrintCommandLineFlags",
+            "-Djava.awt.headless=true"));
+        if (JenkinsRule.SLAVE_DEBUG_PORT > 0) {
+            cmd.add("-Xdebug");
+            cmd.add("Xrunjdwp:transport=dt_socket,server=y,address=" + (JenkinsRule.SLAVE_DEBUG_PORT + r.jenkins.getNodes().size() - 1));
+        }
+        cmd.addAll(Arrays.asList(
+            "-jar", Which.jarFile(Launcher.class).getAbsolutePath(),
+            "-jnlpUrl", r.getURL() + "computer/" + name + "/slave-agent.jnlp"));
+        ProcessBuilder pb = new ProcessBuilder(cmd);
         pb.redirectErrorStream(true);
         System.err.println("Running: " + pb.command());
         Process proc = pb.start();
--- a/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
+++ b/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
@@ -1,0 +1,108 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2022 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.jvnet.hudson.test;
+
+import edu.umd.cs.findbugs.annotations.CheckForNull;
+import edu.umd.cs.findbugs.annotations.NonNull;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
```diff
@@ -0,0 +1,108 @@
...
+     * Creates, attaches, and starts a new inbound agent.
+     * @param name an optional {@link Slave#getNodeName}
+     */
+    public Slave createAgent(@NonNull JenkinsRule r, @CheckForNull String name) throws Exception {
+        if (name == null) {
+            name = "agent" + r.jenkins.getNodes().size();
+        }
+        // TODO 2.216+ support WebSocket option
+        DumbSlave s = new DumbSlave(name, new File(r.jenkins.getRootDir(), "agent-work-dirs/" + name).getAbsolutePath(), new JNLPLauncher(true));
+        s.setNumExecutors(1); // TODO pending 2.234+
```
timja, line 62: seems reasonable to bump it?

jglick: Possibly. Seems the `plugin-pom` currently expects 2.249+ but it is unclear to me whether this is a hard requirement:

> `jenkins.version>2.249.1` is expected to be used for most plugins.

basil: We don't have a formal policy, but I think a de facto policy of waiting until the given core disappears from the Update Center (currently 500 days) is a good enough policy as any.

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
```diff
@@ -0,0 +1,111 @@
...
+        return s;
+    }
+
+    /**
+     * (Re-)starts an existing inbound agent.
+     */
+    @SuppressFBWarnings(value = "COMMAND_INJECTION", justification = "just for test code")
+    public void start(@NonNull JenkinsRule r, String name) throws Exception {
+        stop(name);
+        ProcessBuilder pb = new ProcessBuilder(JavaEnvUtils.getJreExecutable("java"), "-Djava.awt.headless=true", "-jar", Which.jarFile(Launcher.class).getAbsolutePath(), "-jnlpUrl", r.getURL() + "computer/" + name + "/slave-agent.jnlp");
```
basil, line 77: This uses JVM ergonomics to set the maximum heap size and provides no observability. In the past, these deficiencies have led to the development of #380 and #425. If we are introducing a new API and expecting consumers to use it, I think it should not come with these deficiencies which have been already identified and solved elsewhere.

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
```diff
@@ -0,0 +1,111 @@
...
+     * @param name an optional {@link Slave#getNodeName}
+     */
+    @SuppressFBWarnings(value = "PATH_TRAVERSAL_IN", justification = "just for test code")
+    public Slave createAgent(@NonNull JenkinsRule r, @CheckForNull String name) throws Exception {
+        if (name == null) {
+            name = "agent" + r.jenkins.getNodes().size();
+        }
+        // TODO 2.216+ support WebSocket option
+        DumbSlave s = new DumbSlave(name, new File(r.jenkins.getRootDir(), "agent-work-dirs/" + name).getAbsolutePath(), new JNLPLauncher(true));
+        s.setNumExecutors(1); // TODO pending 2.234+
```
basil, line 64: A common use case I have seen in plugins is the desire to create an agent with more than one executor for complex testing scenarios, for example in `throttle-concurrent-builds`. While I recognize that this is not a use case you are intending to address in this PR, I think it would be worth thinking about how to accommodate such a use case in a future change and designing the API to support (or at least not preclude) such a future use case.

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
jglick: Seems reasonable. I do not see any particular reason to complicate this PR with such a bump, since it works fine as it is and the edits to clean up with a new core are very simple.

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
jglick: I will take a look at those and try to make matching changes. Note that this code is taken verbatim from `workflow-durable-task-step-plugin` where it has been running without apparent issues for 7½y (https://github.com/jenkinsci/pipeline-plugin/commit/cd1572249248608ef9ccb904c99c707ee8e0a655).

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
basil: I suppose they might be simple, but they were not obvious to me. In the past you have written comments like "TODO 2.xxx+ [newer API]" but I notice that in this PR the future work is left as an exercise for the reader.

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
basil: But if I recall correctly that might be too late, since we would have already launched the agent with the wrong number of executors, and `setNumExecutors` doesn't relaunch it. And it is awkward to launch with the wrong number and then relaunch with the correct number. So I think to accommodate such a future use case the API that launches the agent would need to support customizing the number of executors prior to launch. If the API you are proposing here can be extended to support that use case, great.

basil:
Causes `org.jenkinsci.plugins.workflow.steps.scm.SCMStepTest#checkoutsRestored` to start failing with

```
java.lang.AssertionError: 

Expected: a string containing "Cloning the remote Git repository"
     but: was "Started by an SCM change
[Pipeline] Start of Pipeline
[Pipeline] node
Running on slave0 in /home/jenkins/workspace/Tools_bom_PR-1255/pct-work/workflow-scm-step/target/tmp/j h17894114103649561928/agent-work-dirs/slave0/workspace/p
[Pipeline] {
[Pipeline] ws
Running in /home/jenkins/workspace/Tools_bom_PR-1255/pct-work/workflow-scm-step/target/tmp/j h17894114103649561928/agent-work-dirs/slave0/workspace/p@2
[Pipeline] {
[Pipeline] git
The recommended git tool is: NONE
No credentials specified
Fetching changes from the remote Git repository
Checking out Revision 01ae2e7fa78cc0696363f238c012a7aff4240801 (refs/remotes/origin/master)
 > git rev-parse --resolve-git-dir /home/jenkins/workspace/Tools_bom_PR-1255/pct-work/workflow-scm-step/target/tmp/j h17894114103649561928/agent-work-dirs/slave0/workspace/p@2/.git # timeout=10
 > git config remote.origin.url /tmp/junit13354845484223236222/junit3675871739211897507 # timeout=10
Fetching upstream changes from /tmp/junit13354845484223236222/junit3675871739211897507
 > git --version # timeout=10
 > git --version # 'git version 2.34.1'
 > git fetch --tags --force --progress -- /tmp/junit13354845484223236222/junit3675871739211897507 +refs/heads/*:refs/remotes/origin/* # timeout=10
 > git rev-parse refs/remotes/origin/master^{commit} # timeout=10
 > git config core.sparsecheckout # timeout=10
 > git checkout -f 01ae2e7fa78cc0696363f238c012a7aff4240801 # timeout=10
 > git branch -a -v --no-abbrev # timeout=10
 > git branch -D master # timeout=10
 > git checkout -b master 01ae2e7fa78cc0696363f238c012a7aff4240801 # timeout=10
Commit message: "next"
 > git rev-list --no-walk c41f60902fcb334dd411f0032c66e949505c05dc # timeout=10
[Pipeline] }
[Pipeline] // ws
[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline
Finished: SUCCESS
"
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
	at org.jvnet.hudson.test.JenkinsRule.assertLogContains(JenkinsRule.java:1547)
	at org.jenkinsci.plugins.workflow.steps.scm.SCMStepTest.lambda$checkoutsRestored$2(SCMStepTest.java:125)
	at org.jvnet.hudson.test.RestartableJenkinsRule$3.evaluate(RestartableJenkinsRule.java:245)
	at org.jvnet.hudson.test.RestartableJenkinsRule$6.evaluate(RestartableJenkinsRule.java:293)
	at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:616)
```

PR #450, opened 2022-07-01
`RealJenkinsRule`: Add new flavors of `then` / `runRemotely` allowing to return `Serializable` values

Vlatombe:

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -418,13 +418,28 @@ public final class RealJenkinsRule implements TestRule {
         void run(JenkinsRule r) throws Throwable;
     }
 
+    @FunctionalInterface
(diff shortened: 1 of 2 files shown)


jglick (approved):

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -590,14 +602,21 @@ public void stopJenkins() throws Throwable {
         proc = null;
     }
 
-    public void runRemotely(Step s) throws Throwable {
+    public Serializable runRemotely(Step s) throws Throwable {
+        return runRemotely(new StepToStep2(s));
```
jglick, line 606:
```suggestion
    public void runRemotely(Step s) throws Throwable {
        runRemotely(new StepToStep2(s));
```
?

jglick (commented):
Looks like javac does not get confused about overloads when presented with a method handle.

PR #451, opened 2022-07-04
`RealJenkinsRule`: add ability to check if Jenkins fails to boot.

joseblas:
When executing some tests that are expecting Jenkins not to start the timeout is never executed. Changing to before the status check makes that possible, also makes more sense as the time starts to count when the server starts, not when is up and running.

No new tests required as the functionality remains the same.


jtnord:
This seems incorrect to me - if Jenkins is never going to start up then the process should have exited or there should be a [boot failure](https://github.com/jenkinsci/jenkins/blob/a351a55418e69a01bf6ffa47fa8130171f6012da/core/src/main/java/hudson/util/BootFailure.java).

If Jenkins has failed to start then the `boot-failure.txt` file should exist and this should probably be used to exit early rather than a timeout which will always lead to the test taking longer than expected?  if the java process has failed to start the proc should have exited so checking that it has not in fact terminated would also be expected.

joseblas:
Indeed, didn't remember that file. Thanks @jtnord

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -525,6 +525,8 @@ public void startJenkins() throws Throwable {
         // TODO prefix streams with per-test timestamps & port
         new StreamCopyThread(description.toString(), proc.getInputStream(), System.out).start();
         new StreamCopyThread(description.toString(), proc.getErrorStream(), System.err).start();
+
+        addTimeout();
```
jglick, line 529: (or could just be inlined, though the current diff is minimal which is nice)

jglick:
> No new tests required as the functionality remains the same.

That is not true, right? The whole point is that this patch (possibly augmented by a more eager check for `boot-failure.txt`) would enable you to write a test that expects Jenkins not to start which would fail before and pass now. Or am I missing something? Hard to evaluate without a concrete example of what you are trying to do (preferably in the form of a new test case).

joseblas:
@jglick no tests needed if the change were only to move up the timeout. With current changes it needs them for sure.

jglick:
> no tests needed if the change were only to move up the timeout

I am not sure that is true. I am still unclear on what problem you are solving exactly—the original motivation. Unit tests are fine (though I would prefer simple Java code rather than mocks) but there is no test of RJR showing what was broken before which is now fixed. Specifically what does

> tests that are expecting Jenkins not to start

mean?

joseblas:
When RJR cannot start (for whatever reason) it hangs up in the while(true) loop.
First attempt was to put timeout up before the loop, which is flaky as @jtnord mentioned.
So, I added this change, if the status is providing a 500 (instead of classic not ready - 503) then there is an issue starting RJR.
This corner case if what this PR wants to fix.

With the UT I demonstrate that when a 500 is coming the loop is broken and an exception is thrown.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -534,21 +536,22 @@ public void startJenkins() throws Throwable {
...
-                        throw new IOException("Response code " + code + " for " + status + ": " + err + " " + conn.getHeaderFields());
+                    }else {
+                        throw new IOException("Response code " + conn.getResponseCode() + " for " + status + ": " + checkResult.get() +
+                                                      " " + conn.getHeaderFields());
                     }
+
+                } catch (JenkinsStartupException jse) {
+                    // Jenkins has completed startup but failed
+                    // do not make any further attempts and kill the process
+                    System.err.println("Jenkins failed to start");
```
jtnord, line 551: debug code - should not be needed as this should be obvious from the exception

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -560,6 +563,33 @@ public void startJenkins() throws Throwable {
             }
             Thread.sleep(100);
         }
+        addTimeout();
+    }
+
+    public static Optional<String> checkResult(HttpURLConnection conn) throws IOException {
+
+        int code = conn.getResponseCode();
+        System.out.println("Response Code "+ code);
```
jtnord, line 572: debug code - should be removed or use logging at a lower level?

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -560,6 +563,33 @@ public void startJenkins() throws Throwable {
...
+            String err = "?";
+            try (InputStream is = conn.getErrorStream()) {
+                if (is != null) {
+                    err = IOUtils.toString(is, StandardCharsets.UTF_8);
+                }
+            } catch (Exception x) {
+                x.printStackTrace();
+            }
+            if (code == 500) {
+                throw new JenkinsStartupException("Jenkins failed to start");
```
jtnord, line 586: keep the `err` message so that you can check the error if needed (and it is not lost after the fact)

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -823,4 +853,11 @@ public Serializable run(JenkinsRule r) throws Throwable {
             return null;
         }
     }
+
+    public static class JenkinsStartupException extends IOException {
+
```
jtnord, line 858: keep the error messsage returned from the server so it can be inspected?

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
```diff
@@ -179,6 +189,42 @@ private static void _stepsDoNotRunOnHttpWorkerThread(JenkinsRule r) throws Throw
         assertNull(Stapler.getCurrentRequest());
     }
 
+
+    @Test
+    public void test500Errors() throws IOException {
```
jtnord, line 194: an IT with a plugin that throws an exception in an initializer that would block jenkins startup would be good here, as that is what you are making these changes for?

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -560,6 +562,32 @@ public void startJenkins() throws Throwable {
             }
             Thread.sleep(100);
         }
+        addTimeout();
+    }
+
+    public static Optional<String> checkResult(HttpURLConnection conn) throws IOException {
```
jglick, line 568: BTW I think a simple `@CheckForNull String` would suffice here. `Optional` is useful for certain advanced idioms (like `Stream`s) but just seems like overkill for direct use. Not sure if there is any evolving consensus on this stylistic point.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -560,6 +562,32 @@ public void startJenkins() throws Throwable {
...
+            String err = "?";
+            try (InputStream is = conn.getErrorStream()) {
+                if (is != null) {
+                    err = IOUtils.toString(is, StandardCharsets.UTF_8);
+                }
+            } catch (Exception x) {
+                x.printStackTrace();
+            }
+            if (code == 500) {
+                throw new JenkinsStartupException("Jenkins failed to start", err);
```
jglick, line 584: Why not just make `err` be the `message`? The message passed here is redundant (implied by the exception class).

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
```diff
@@ -179,6 +190,57 @@ private static void _stepsDoNotRunOnHttpWorkerThread(JenkinsRule r) throws Throw
...
+     * plugins/failure.hpi
+     *  Plugin that has this:
+     *
+     *  @Initializer(after=JOB_LOADED)
+     *     public static void init() throws IOException {
+     *         throw new IOException("oops");
+     *     }
+     *
+     */
+    @Test(expected = RealJenkinsRule.JenkinsStartupException.class)
```
jglick, line 239: Is there an expected `err` in this case? (can use `assertThrows` for more control)

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
```diff
@@ -179,6 +190,57 @@ private static void _stepsDoNotRunOnHttpWorkerThread(JenkinsRule r) throws Throw
...
 
+
+    @Test
+    public void test500Errors() throws IOException {
+        HttpURLConnection conn = mock(HttpURLConnection.class);
+        when(conn.getResponseCode()).thenReturn(500);
+        try {
+            RealJenkinsRule.checkResult(conn);
+            fail("expected exception JenkinsStartupException");
+        }catch (RealJenkinsRule.JenkinsStartupException ae){
```
jglick, line 201: `assertThrows`?

jglick (commented):
Looks close. Some minor suggestions.

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
```diff
@@ -195,22 +198,19 @@ private static void _stepsDoNotRunOnHttpWorkerThread(JenkinsRule r) throws Throw
...
         when(conn.getResponseCode()).thenReturn(500);
-        try {
-            RealJenkinsRule.checkResult(conn);
-            fail("expected exception JenkinsStartupException");
-        }catch (RealJenkinsRule.JenkinsStartupException ae){
-
-        }
+        RealJenkinsRule.JenkinsStartupException jse = assertThrows(RealJenkinsRule.JenkinsStartupException.class,
+                                                                    () -> RealJenkinsRule.checkResult(conn));
+        assertThat(jse, isA(RealJenkinsRule.JenkinsStartupException.class));
```
jglick, line 203: I think this cannot fail?

jglick (approved):
Some trivia comments, but I think this is good to go; @jtnord do you care to review again?

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -79,6 +80,7 @@
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import javax.annotation.CheckForNull;
```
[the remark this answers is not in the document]
joseblas: I thought edu.umd.cs.findbugs.annotations was the deprecated?

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -560,6 +563,33 @@ public void startJenkins() throws Throwable {
...
+            conn.getInputStream().close();
+            return null;
+        } else {
+            String err = "?";
+            try (InputStream is = conn.getErrorStream()) {
+                if (is != null) {
+                    err = IOUtils.toString(is, StandardCharsets.UTF_8);
+                }
+            } catch (Exception x) {
+                x.printStackTrace();
```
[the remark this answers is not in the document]
joseblas: my 2cts, this exception is ok to be swallowed, the error is already there, not a extra fail is needed

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
jtnord: not sure if the stack suppression takes care of hiding that from us or not - but if not could be worth a check.

jtnord (approved):

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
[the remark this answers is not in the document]
jtnord: `javax.annotation.CheckForNull` is the JSR-305 variant which has questionable licensing.
the `edu.umd.cs.findbugs` ones are not deprecated (with a suitably recent version of spotbugs)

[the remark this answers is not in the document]
jglick: Actually in this repo `edu.umd.cs.findbugs.annotations.CheckForNull` etc. _are_ still deprecated, but only because it is still compatible with an old version of Jenkins. That will change. In the meantime we use the proper imports, e.g. https://github.com/jenkinsci/jenkins-test-harness/blob/711534844e8a24c857c45165600996391972b0af/src/main/java/org/jvnet/hudson/test/JenkinsRule.java#L61

jtnord (approved):

PR #452, opened 2022-07-04
Show log records when `RecordedMatcher` doesn't match

timja:
Previously the matcher showed the LoggerRule toString:

```
Expected: has LogRecord with level \"INFO\" with a message matching \"Entry 2\" 
    but: was <org.jvnet.hudson.test.LoggerRule@60dcc9fe>
```

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/LoggerRule.java
+++ b/src/main/java/org/jvnet/hudson/test/LoggerRule.java
@@ -81,6 +81,14 @@ public class LoggerRule extends ExternalResource {
         return this;
     }
 
+    @Override
+    public String toString() {
+        return getRecords()
+                .stream()
+                .map(logRecord -> logRecord.getLevel().toString() + "->" + logRecord.getMessage())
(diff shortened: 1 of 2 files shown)


basil (approved):
I think it is generally our preferred practice to review both producer and consumer in tandem if possible to validate the use case, but I trust you Tim.

timja:
> I think it is generally our preferred practice to review both producer and consumer in tandem if possible to validate the use case, but I trust you Tim.

The unit test demonstrates this, I provided the previous behaviour in the PR description, that should be enough. :)

basil:
> The unit test demonstrates this, I provided the previous behaviour in the PR description, that should be enough. :)

The unit test shows the what, but not the why (knowledge vs understanding). A PR of the expected consumers reveals the motivation for the functionality. Like I wrote earlier, I trust that you have a solid use case.

timja:
I was developing a unit test and it failed showing me a useless toString :).


but sure :)

PR #453, opened 2022-07-05
Jetty 10 based `JenkinsRule`

basil:
Do _not_ review. This PR makes this repository require Java 11, so it is not suitable for merge. But I am filing this PR to get an incremental build of the test harness with a `JenkinsRule` based on Jetty 10. When running core's `jenkins.agents.WebSocketAgentsTest` with the Jetty 10 PR and this version of the test harness, the test failed as expected, because core's WebSocket implementation currently doesn't work on Jetty 10. Once it does, I plan to run the core test suite and BOM/PCT tests with this test harness to make sure there are no other unexpected issues.


src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
```diff
@@ -543,6 +546,7 @@ protected ServletContext createWebServer() throws Exception {
         context.setConfigurations(new Configuration[]{new WebXmlConfiguration()});
         context.addBean(new NoListenerConfiguration(context));
         server.setHandler(context);
+        JettyWebSocketServletContainerInitializer.configure(context, null);
```
jglick, line 549: https://github.com/jenkinsci/jenkins/pull/6785#discussion_r914347371 FTR

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -796,6 +799,7 @@ public static ImmutablePair<Server, ServletContext> _createWebServer(String cont
         context.setConfigurations(new Configuration[]{new WebXmlConfiguration()});
         context.addBean(new NoListenerConfiguration(context));
         server.setHandler(context);
+        JettyWebSocketServletContainerInitializer.configure(context, null);
```
jglick, line 802: @basil FYI this JTH release breaks a WebSocket test in the `kubernetes` plugin:

```
java.lang.NullPointerException: Cannot invoke "org.eclipse.jetty.websocket.server.JettyWebSocketServerContainer.upgrade(org.eclipse.jetty.websocket.server.JettyWebSocketCreator, javax.servlet.http.HttpServletRequest, javax.servlet.http.HttpServletResponse)" because the return value of "org.eclipse.jetty.websocket.server.JettyWebSocketServerContainer.getContainer(javax.servlet.ServletContext)" is null
	at jenkins.websocket.Jetty10Provider.handle(Jetty10Provider.java:61)
	at jenkins.websocket.WebSockets.lambda$upgrade$0(WebSockets.java:70)
```

I am still trying to understand why; in the meantime, https://github.com/jenkinsci/kubernetes-plugin/pull/1275/commits/dcc8137937248b52ff7754287ccb777e1f503d9a

jglick: Ah, never mind, I think this is due to a stale copy in that plugin (https://github.com/jenkinsci/jenkins-test-harness/pull/289#issuecomment-809560892).

basil: Glad to hear it, especially given that I tested this case by running the Swarm plugin's `JenkinsRule`-based tests (which use WebSocket) against this PR prior to its merge and release.

PR #465, opened 2022-07-26
Handle empty port to fix RealJenkinsRule flakiness

Pldi23:
Cloudbees CI reported a number of flaky tests encountering in [operations-center-context plugin](https://github.com/cloudbees/operations-center-context/blob/master/src/test/java/com/cloudbees/opscenter/artifacts/RemoteArtifactCopyBuilderTest.java#L98)

Logs:
[RemoteArtifactCopyBuilderTest.txt](https://github.com/jenkinsci/jenkins-test-harness/files/9189993/RemoteArtifactCopyBuilderTest.txt)
[RemoteArtifactCopyBuilderTest_stacktrace.txt](https://github.com/jenkinsci/jenkins-test-harness/files/9189995/RemoteArtifactCopyBuilderTest_stacktrace.txt)

Changes made in scope of this PR:

1. Handle empty port. [514bfbf](https://github.com/jenkinsci/jenkins-test-harness/pull/465/commits/514bfbf4b6cc5d93e74d341de0c16f5146b78fe7)
2. Stop proc once. [38b7261](https://github.com/jenkinsci/jenkins-test-harness/pull/465/commits/38b726176ad11ae48b6575aa5260ac024e8cc3e8)

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -609,6 +609,10 @@ public final class RealJenkinsRule implements TestRule {
 
     private static int readPort(File portFile) throws IOException {
         String s = FileUtils.readFileToString(portFile, StandardCharsets.UTF_8);
+        if (s.isEmpty()) {
+            LOGGER.warning(() -> String.format("PortFile: %s exists, but value is still not written", portFile.getAbsolutePath()));
+            return 0;
+        }
         try {
             return Integer.parseInt(s);
         } catch (NumberFormatException e) {
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -611,7 +611,7 @@ public final class RealJenkinsRule implements TestRule {
         String s = FileUtils.readFileToString(portFile, StandardCharsets.UTF_8);
 
         // Work around to non-atomic write of port value in Winstone releases prior to 6.1.
-        // Todo When Winstone 6.2 has been widely adopted, this can be deleted.
+        // TODO When Winstone 6.2 has been widely adopted, this can be deleted.
         if (s.isEmpty()) {
             LOGGER.warning(() -> String.format("PortFile: %s exists, but value is still not written", portFile.getAbsolutePath()));
             return 0;
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -385,10 +385,11 @@ public final class RealJenkinsRule implements TestRule {
                     }
                     System.out.println("Will load plugins: " + Stream.of(plugins.list()).filter(n -> n.matches(".+[.][hj]p[il]")).sorted().collect(Collectors.joining(" ")));
                     base.evaluate();
+                } catch (Throwable t) {
+                    LOGGER.log(Level.WARNING, "Something went wrong", t);
+                    throw t;
                 } finally {
-                    if (proc != null) {
-                        stopJenkins();
-                    }
+                    stopJenkins();
                     try {
                         tmp.dispose();
                     } catch (Exception x) {
@@ -615,18 +616,27 @@ public final class RealJenkinsRule implements TestRule {
         }
     }
 
+    /**
+     * Stops Jenkins and releases any system resources associated
+     * with it. If Jenkins is already stopped then invoking this
+     * method has no effect.
+     */
     public void stopJenkins() throws Throwable {
-        endpoint("exit").openStream().close();
-        if (!proc.waitFor(60, TimeUnit.SECONDS) ) {
-            System.err.println("Jenkins failed to stop within 60 seconds, attempting to kill the Jenkins process");
-            proc.destroyForcibly();
-            throw new AssertionError("Jenkins failed to terminate within 60 seconds");
+        if (proc != null) {
+            Process _proc = proc;
+            proc = null;
+            endpoint("exit").openStream().close();
+
+            if (!_proc.waitFor(60, TimeUnit.SECONDS) ) {
+                System.err.println("Jenkins failed to stop within 60 seconds, attempting to kill the Jenkins process");
+                _proc.destroyForcibly();
+                throw new AssertionError("Jenkins failed to terminate within 60 seconds");
+            }
+            int exitValue = _proc.exitValue();
+            if (exitValue != 0) {
+                throw new AssertionError("nonzero exit code: " + exitValue);
+            }
         }
-        int exitValue = proc.exitValue();
-        if (exitValue != 0) {
-            throw new AssertionError("nonzero exit code: " + exitValue);
-        }
-        proc = null;
     }
 
     public void runRemotely(Step s) throws Throwable {
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -609,10 +609,14 @@ public final class RealJenkinsRule implements TestRule {
 
     private static int readPort(File portFile) throws IOException {
         String s = FileUtils.readFileToString(portFile, StandardCharsets.UTF_8);
+
+        // Work around to non-atomic write of port value in Winstone releases prior to 6.1.
+        // Todo When Winstone 6.2 has been widely adopted, this can be deleted.
         if (s.isEmpty()) {
             LOGGER.warning(() -> String.format("PortFile: %s exists, but value is still not written", portFile.getAbsolutePath()));
             return 0;
         }
+
         try {
             return Integer.parseInt(s);
         } catch (NumberFormatException e) {


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -620,7 +620,8 @@ public void stopJenkins() throws Throwable {
         if (!proc.waitFor(60, TimeUnit.SECONDS) ) {
             System.err.println("Jenkins failed to stop within 60 seconds, attempting to kill the Jenkins process");
             proc.destroyForcibly();
-            throw new AssertionError("Jenkins failed to terminate within 60 seconds");
+            proc = null;
```
Pldi23, line 623: What happens when test flakes and fails:

1. test runs well and going to [stopJenkins](https://github.com/jenkinsci/jenkins-test-harness/blob/6a1fdad76d6fa4998a5211f62008f234ea04bf87/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java#L442) at the end.
2. stopJenkins [hit timeout](https://github.com/jenkinsci/jenkins-test-harness/blob/6a1fdad76d6fa4998a5211f62008f234ea04bf87/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java#L620).
3. AssertExceptions thown as a result
4. [finally block](https://github.com/jenkinsci/jenkins-test-harness/blob/6a1fdad76d6fa4998a5211f62008f234ea04bf87/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java#L389-L396) executed and because proc != null tries to [stop Jenkins again](https://github.com/jenkinsci/jenkins-test-harness/blob/6a1fdad76d6fa4998a5211f62008f234ea04bf87/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java#L390)
5. ConnectionException throws [here](https://github.com/jenkinsci/jenkins-test-harness/blob/6a1fdad76d6fa4998a5211f62008f234ea04bf87/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java#L619)

Pldi23: My suggestion is to set proc = null (like [here](https://github.com/jenkinsci/jenkins-test-harness/blob/6a1fdad76d6fa4998a5211f62008f234ea04bf87/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java#L594-L595)) instead of throwing AssertError

mikecirioli (approved):

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -620,7 +620,8 @@ public void stopJenkins() throws Throwable {
         if (!proc.waitFor(60, TimeUnit.SECONDS) ) {
             System.err.println("Jenkins failed to stop within 60 seconds, attempting to kill the Jenkins process");
             proc.destroyForcibly();
-            throw new AssertionError("Jenkins failed to terminate within 60 seconds");
```
jtnord, line 623: so to me this is trading one set of flakes for another.

Jenkins could not shut down within the requested time - presumably there is some code that is running / blocking in a `@Terminator`  

Various things could be happening in this state including flushing state to disk, closing databases etc and if the process can not close cleanly these will no longer happen - and then you have failures later on because something that was expected to be persisted was not in fact cleanly persisted.

Did you investigate why Jenkins could not shut down within the 60 seconds?

Pldi23: > Did you investigate why Jenkins could not shut down within the 60 seconds?

I'll try to understand it, but locally when running the tests `waitFor` takes no more then 15-20 seconds (to reproduce the flake I had to set timeout to 3 seconds)

jtnord: from the attached logs -> `Shutting down a Jenkins instance that was still starting up`  is likely to cause issues.

Perhaps something is not waiting long enough for Jenkins to start?

basil: I like the idea of printing a thread dump for debugging purposes. To get the PID for pre-Java 11 versions of Jenkins you can call `GNUCLibrary.LIBC.getpid()` (guarded by `if (Functions.isGlibcSupported()) {`).

jglick (changes requested):
Setting `proc = null` seems uncontroversial, just a bug fix. Replacing the `AssertionError` with a potentially passing test seems like it is just masking an actual problem in the test.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jglick: From inside the Jenkins JVM you can use https://github.com/jenkinsci/jenkins-test-harness/blob/0038ef9b8a0bb2fec2ef4be67d1124f51db53076/src/main/java/org/jvnet/hudson/test/JenkinsRule.java#L458-L464

basil (changes requested):
Hello @Pldi23, thank you for your contribution to the open-source Jenkins test harness.

`RealJenkinsRule#then` always stops Jenkins after it is started, so the only reason for the `stopJenkins` call in `RealJenkinsRule#apply` must be as a fallback if someone started Jenkins by some other means but forgot to stop it: for example, by invoking `RealJenkinsRule#startJenkins` directly from a test but forgetting to call `RealJenkinsRule#stopJenkins`. It is certainly better to stop Jenkins late rather than never. (It might be good to log a warning before doing so, since this indicates a programming error.)

It does seem wrong that we would ever try to stop Jenkins via the `/exit` endpoint twice: it is likely to fail the second time as you observed. Taking inspiration from the Java [`Closeable#close`](https://docs.oracle.com/javase/8/docs/api/java/io/Closeable.html#close--) interface, I think these semantics would be desirable:

> If the stream is already closed then invoking this method has no effect.

In other words, the way `close()` is usually implemented in Java `Closeables` is to have some internal state regarding whether the resource has already been closed and to do nothing if it already has been closed. This is, in some way, indicated by `proc` being non-`null`. However, if we take the above design as the goal, then the current logic has two problems:

- The check for `proc` being `null` happens outside the call to `stopJenkins` rather than within it, violating the contract that invoking the method a second time has no effect
- Once we are past the critical point (calling `/exit`) no state is recorded that would prevent us from trying to invoke that endpoint a second time (erroneously!)

My concrete suggestion would be something like this:

```java
diff --git a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
index d856b02..ca04515 100644
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -180,12 +180,14 @@ public final class RealJenkinsRule implements TestRule {
 
-     private Process proc;
+    private volatile Process proc;
 
     // TODO may need to be relaxed for Gradle-based plugins
     private transient boolean supportsPortFileName;
 
+    private volatile boolean stopping = false;
+
     /**
      * Add some plugins to the test classpath.
      *
@@ -386,9 +388,7 @@ public final class RealJenkinsRule implements TestRule {
                     System.out.println("Will load plugins: " + Stream.of(plugins.list()).filter(n -> n.matches(".+[.][hj]p[il]")).sorted().collect(Collectors.joining(" ")));
                     base.evaluate();
                 } finally {
-                    if (proc != null) {
-                        stopJenkins();
-                    }
+                    stopJenkins();
                     try {
                         tmp.dispose();
                     } catch (Exception x) {
@@ -615,18 +615,31 @@ public final class RealJenkinsRule implements TestRule {
         }
     }
 
+    /*
+     * Stops Jenkins and releases any system resources associated
+     * with it. If Jenkins is already stopped then invoking this
+     * method has no effect.
+     */
     public void stopJenkins() throws Throwable {
-        endpoint("exit").openStream().close();
-        if (!proc.waitFor(60, TimeUnit.SECONDS) ) {
-            System.err.println("Jenkins failed to stop within 60 seconds, attempting to kill the Jenkins process");
-            proc.destroyForcibly();
-            throw new AssertionError("Jenkins failed to terminate within 60 seconds");
+        if (!stopping) {
+            stopping = true;
+            endpoint("exit").openStream().close();
         }
-        int exitValue = proc.exitValue();
-        if (exitValue != 0) {
-            throw new AssertionError("nonzero exit code: " + exitValue);
+        if (proc != null) {
+            try {
+                if (!proc.waitFor(60, TimeUnit.SECONDS)) {
+                    System.err.println("Jenkins failed to stop within 60 seconds, attempting to kill the Jenkins process");
+                    proc.destroyForcibly();
+                    throw new AssertionError("Jenkins failed to terminate within 60 seconds");
+                }
+                int exitValue = proc.exitValue();
+                if (exitValue != 0) {
+                    throw new AssertionError("nonzero exit code: " + exitValue);
+                }
+            } finally {
+                proc = null;
+            }
         }
-        proc = null;
     }
```

This accomplishes the following:
- Prevents `/exit` from being called twice by guarding it on a `stopping` boolean. We don't want it to be called more than once because the first call will prevent the second from succeeding.
- Sets `proc` to null even when an assertion is thrown if Jenkins failed to terminate within 60 seconds or terminates with a non-zero exit code. This allows the call to `stopJenkins` from `#then` to fail as expected and then allows the second call to `stopJenkins` from `RealJenkinsRule#apply` to not try to kill the process a second time (which is unnecessary and might confuse the user with a secondary failure mode).

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
Pldi23: @jtnord That’s interesting, this log I’ve attached, was collected March 22, then, I assume, the issue with 'port already in use'  was fixed, but Tests continue to be flaky (last failure Jul 23) but the log now is changed and provide much less details. 
[stacktrace.txt](https://github.com/jenkinsci/jenkins-test-harness/files/9196171/stacktrace.txt)
[st_out.txt](https://github.com/jenkinsci/jenkins-test-harness/files/9196173/st_out.txt)
[st_err.txt](https://github.com/jenkinsci/jenkins-test-harness/files/9196175/st_err.txt)
Reading the logs I even don't see that RealJenkinsRule started and ready, I'll try to dig deeper here, maybe I'll understand something

Pldi23:
Thanks @basil , I agree with your suggestion and applied it in [70722b9](https://github.com/jenkinsci/jenkins-test-harness/pull/465/commits/70722b9da015471f11a7266a9e4b676eeb6aae10). With this changes we could clearly see AssertionError (it is not masked now). The only question is, do we want `Process` to be volatile? From what I know it will not take any effect and do we really need `proc` to be thread safe here? If yes we need to use ThreadLocal or other mechanism to synchronise Process, but from what I see it is not needed to be atomic, but I might be missing something.

basil (approved):
Code change looks good. Regarding synchronization, I do not think it is needed but it could not hurt. Simplest would be to use `public synchronized void stopJenkins` and remove volatile from `stopping`. Or keep it simple and do nothing, since we do not expect this code to be used in a multithreaded context.

Now that `stopJenkins` is doing the check for `proc != null` internally the same check is redundant in the caller in `RealJenkinsRule#apply` and can be deleted:

```java
@@ -386,9 +388,7 @@ public final class RealJenkinsRule implements TestRule {
                     System.out.println("Will load plugins: " + Stream.of(plugins.list()).filter(n -> n.matches(".+[.][hj]p[il]")).sorted().collect(Collectors.joining(" ")));
                     base.evaluate();
                 } finally {
-                    if (proc != null) {
-                        stopJenkins();
-                    }
+                    stopJenkins();
                     try {
                         tmp.dispose();
                     } catch (Exception x) {
```

However given my earlier point …

> `RealJenkinsRule#then` always stops Jenkins after it is started, so the only reason for the `stopJenkins` call in `RealJenkinsRule#apply` must be as a fallback if someone started Jenkins by some other means but forgot to stop it: for example, by invoking `RealJenkinsRule#startJenkins` directly from a test but forgetting to call `RealJenkinsRule#stopJenkins`. It is certainly better to stop Jenkins late rather than never. (It might be good to log a warning before doing so, since this indicates a programming error.)

… it may be wise to log a warning in this case. This is basically serving the purpose of a finalizer as described in Joshua Bloch's _Effective Java_:

> So what, if anything, are finalizers good for? There are perhaps two legitimate uses. One is to act as a “safety net” in case the owner of an object forgets to call its explicit termination method. While there’s no guarantee that the finalizer will be invoked promptly, it may be better to free the resource late than never, in those (hopefully rare) cases when the client fails to call the explicit termination method. But the finalizer should log a warning if it finds that the resource has not been terminated, as this indicates a bug in the client code, which should be fixed. If you are considering writing such a safety-net finalizer, think long and hard about whether the extra protection is worth the extra cost. 

Putting Joshua Bloch's advice into practice:

```java
@@ -387,8 +387,9 @@ public final class RealJenkinsRule implements TestRule {
                     base.evaluate();
                 } finally {
                     if (proc != null) {
-                        stopJenkins();
+                        System.err.println("Jenkins still running: Prefer #then, or ensure that every call to #startJenkins has a matching call to #stopJenkins");
                     }
+                    stopJenkins();
                     try {
                         tmp.dispose();
                     } catch (Exception x) {
```

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -615,18 +617,31 @@ private static int readPort(File portFile) throws IOException {
...
+     * method has no effect.
+     */
     public void stopJenkins() throws Throwable {
-        endpoint("exit").openStream().close();
-        if (!proc.waitFor(60, TimeUnit.SECONDS) ) {
-            System.err.println("Jenkins failed to stop within 60 seconds, attempting to kill the Jenkins process");
-            proc.destroyForcibly();
-            throw new AssertionError("Jenkins failed to terminate within 60 seconds");
+        if (!stopping) {
+            stopping = true;
```
jtnord, line 627: this is technically not threadsafe and could in some really narrow edge case still be call stop twice.

if you care about thread safety here then this should use an `AtomicBoolean` and `if (!stopping.getAndSet(true))` 

But more importantly what sets this back to `false`?

AFAICT this needs to be reset when starting Jenkins as you can have multiple start/stops of a `RealJenkinsRule` using 
multiple `then` inside a given test which will no longer close / stop the subsequent Jenkins processes.

This is very likely the cause of the test failures here.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jtnord: it does appear as though Jenkins hasn't started up - I would guess the close is caused due to the teardown due to a test timeout.  
This to me signifies that this is running on a machine that is woefully underpowered, or there is a deadlock at startup, or the timeout is using a wall-clock and there has been a clock reset on the underlying hardware.  Getting the threadDump of the Jenkins process (and any locks) could reveal the smoking gun here.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
basil: > AFAICT this needs to be reset when starting Jenkins as you can have multiple start/stops of a RealJenkinsRule using

+1

Pldi23: Agree, AtomicBoolean is more reliable here, please see [55cc956](https://github.com/jenkinsci/jenkins-test-harness/pull/465/commits/55cc9566d0a7ccf9bdf177f734c4751d3336bcb7)

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -388,9 +388,14 @@ public RealJenkinsRule includeTestClasspathPlugins(boolean includeTestClasspathP
                     }
                     System.out.println("Will load plugins: " + Stream.of(plugins.list()).filter(n -> n.matches(".+[.][hj]p[il]")).sorted().collect(Collectors.joining(" ")));
                     base.evaluate();
+                } catch (Throwable t) {
+                    LOGGER.log(Level.WARNING, "Something went wrong", t);
```
Pldi23, line 392: I think we’re loosing a real issue here, let’s log and rethrow

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -607,6 +616,9 @@ private static boolean supportsPortFileName(String war) throws IOException {
     }
 
     private static int readPort(File portFile) throws IOException {
+        if (!portFile.canRead()) {
```
Pldi23, line 619: It’s just my assumption, because I still can’t reproduce the failure, but perhaps something wrong with the file read, could it be the case when file read happens at the same time when port value written and we receive IOException? (reading the logs more carefully I see that when flakes happens the test overall time about 2-3 seconds (usual time 15-20 sec) it means that the Jenkins is just above the start and it’s almost immediately stops)

Pldi23: We could remove this log before merge, I just want somehow to run failing tests on a ci using my fork and examine the logs then.

Pldi23: I managed to reproduce locally flaky failure and the reason is because port value is still empty we catch NumberFormatException and throw AssertionError (it was lost before and now visible because of [61659b8](https://github.com/jenkinsci/jenkins-test-harness/pull/465/commits/61659b8e909e2bd3ca76fa75a50122e6e91b6ea3)), what are your thoughts if we retry to read it until port gets populated? see my idea in [8106c20](https://github.com/jenkinsci/jenkins-test-harness/pull/465/commits/8106c20e0d8464e49a734fbb082a79411cab5d28).

jtnord: well that is an interesting finding :)

https://github.com/jenkinsci/winstone/blob/4754da67bf797b906a5c858b2de63b97ac609543/src/main/java/winstone/Launcher.java#L232-L234 is indeed a source of potential flakyness (that may or may not be your flake) but it looks like it could be.

The file and its contents are not created atomically - so if you get very unlucky the file will exist before it is being written to.
The upstream could be changed (maybe it should) to use an atomic write (write and flush to a intermediate file, then move the file in place with an atomic flag set, but this may really be overkill).  if you want to fix the flakes in the interim then probably if the port number is not valid sleep 5 seconds and try again once only.

Pldi23: IMO the right fix would be [49689e7](https://github.com/jenkinsci/jenkins-test-harness/pull/465/commits/49689e76453cdda1940301f0c2b96490e032358d) . It is the while loop which is waiting until Jenkins complete the start procedure, so the condition when file is ready to provide the port value is not fully correct. (I'm going to test my changes and run flaky tests with this changes to see if they will be still flaky).

jtnord: Yes that should do the trick!
I would probably move that logic to readPort and return 0 if the file is empty which would cause the loop to retry as currently there is a bit of a mix of concerns between the two functions.

Pldi23:
Ok, seems the issue is fixed with this changes, I’ve run flaky RemoteArtifactCopyBuilderTest hundreds of times and all runs are good.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -385,9 +389,14 @@ public RealJenkinsRule includeTestClasspathPlugins(boolean includeTestClasspathP
                     }
                     System.out.println("Will load plugins: " + Stream.of(plugins.list()).filter(n -> n.matches(".+[.][hj]p[il]")).sorted().collect(Collectors.joining(" ")));
                     base.evaluate();
+                } catch (Throwable t) {
+                    LOGGER.log(Level.WARNING, "Something went wrong", t);
+                    throw t;
```
jglick, line 394: I do not understand the purpose of this. If an error is thrown out of a test, the test fails and the error is displayed by the test runner right?

```suggestion
```

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -385,9 +389,14 @@ public RealJenkinsRule includeTestClasspathPlugins(boolean includeTestClasspathP
...
                 } finally {
-                    if (proc != null) {
+                    try {
                         stopJenkins();
+                    } catch (Exception e) {
+                        LOGGER.log(Level.WARNING, "Jenkins was not stopped correctly", e);
                     }
```
jglick, line 400: Why not simply

```suggestion
                    stopJenkins();
```

? If something failed, the test should fail. Why would you suppress this?

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -608,25 +618,42 @@ private static boolean supportsPortFileName(String war) throws IOException {
 
     private static int readPort(File portFile) throws IOException {
         String s = FileUtils.readFileToString(portFile, StandardCharsets.UTF_8);
+        if (StringUtils.isEmpty(s)) {
+            LOGGER.log(Level.WARNING, () -> String.format("PortFile: %s exists, but value is still not written", portFile.getAbsolutePath()));
```
jglick, line 622:
```suggestion
            LOGGER.warning(() -> String.format("PortFile: %s exists, but value is still not written", portFile.getAbsolutePath()));
```

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -608,25 +618,42 @@ private static boolean supportsPortFileName(String war) throws IOException {
 
     private static int readPort(File portFile) throws IOException {
         String s = FileUtils.readFileToString(portFile, StandardCharsets.UTF_8);
+        if (StringUtils.isEmpty(s)) {
```
jglick, line 621:
```suggestion
        if (s.isEmpty() {
```
?

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -608,25 +618,42 @@ private static boolean supportsPortFileName(String war) throws IOException {
...
+     * method has no effect.
+     */
     public void stopJenkins() throws Throwable {
-        endpoint("exit").openStream().close();
-        if (!proc.waitFor(60, TimeUnit.SECONDS) ) {
-            System.err.println("Jenkins failed to stop within 60 seconds, attempting to kill the Jenkins process");
-            proc.destroyForcibly();
-            throw new AssertionError("Jenkins failed to terminate within 60 seconds");
+        if (!stopping.get()) {
+            stopping.set(true);
```
jglick, line 639: If you are not actually using atomic swap operations, why is this field not simply a `boolean`?

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -608,25 +618,42 @@ private static boolean supportsPortFileName(String war) throws IOException {
...
+                    System.err.println("Jenkins failed to stop within 60 seconds, attempting to kill the Jenkins process");
+                    proc.destroyForcibly();
+                    throw new AssertionError("Jenkins failed to terminate within 60 seconds");
+                }
+                int exitValue = proc.exitValue();
+                if (exitValue != 0) {
+                    throw new AssertionError("nonzero exit code: " + exitValue);
+                }
+            } finally {
+                proc = null;
```
jglick, line 654: I suspect you could dispense with `stopping` altogether and just do something like

```java
if (proc != null) {
  Proc _proc = proc;
  proc = null;
  // …now terminate _proc as before
}
```

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -528,7 +538,7 @@ public void startJenkins() throws Throwable {
         new StreamCopyThread(description.toString(), proc.getErrorStream(), System.err).start();
         int tries = 0;
         while (true) {
-            if (port == 0 && portFile != null && portFile.exists()) {
+            if (port == 0 && portFile != null && portFile.canRead()) {
```
jglick, line 541: What is the practical difference here?

jglick (commented):
This seems to be mixing together several different changes:
* Only attempt to stop `proc` once. Seems like a clear bug fix.
* Handle an empty port file. OK, though why would the Surefire JVM be seeing a half-written file to begin with? Did whatever code writes this file neglect to save to a temporary file and atomically rename it?
* Suppressing some exceptions, which sounds wrong.
* Some other changes of unclear importance.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
Pldi23: I lost a lot of time understanding root cause with empty port because first exception was lost when finally throw it’s own exception, I still suggest leave it to at least log an exception.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
Pldi23: agree, done in [aa5dd8f](https://github.com/jenkinsci/jenkins-test-harness/pull/465/commits/aa5dd8fcded997b2fe2d308fd7bb6d57c01d1b29)

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
Pldi23: done in [b9589b3](https://github.com/jenkinsci/jenkins-test-harness/pull/465/commits/b9589b3ad5360095bd830191e7ed3a8073738008)

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
Pldi23: done in [b9589b3](https://github.com/jenkinsci/jenkins-test-harness/pull/465/commits/b9589b3ad5360095bd830191e7ed3a8073738008)

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
Pldi23: oh yes, done in [c3cfed6](https://github.com/jenkinsci/jenkins-test-harness/pull/465/commits/c3cfed66853ce70ca4403094fb67b15bf1760958)

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
Pldi23: The motivation for this change is simple, Jenkins-test-harness does not control the way portFile gets created, potentially it could be created with some security/access options, so I suggest this change, if we are sure that it always accessible for a read once created we could revert this change.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
Pldi23: I would say it should work, do you prefer this approach over atomic synchronisation using `stopping`?

Pldi23:
> This seems to be mixing together several different changes:
> 
> * Only attempt to stop `proc` once. Seems like a clear bug fix.
> * Handle an empty port file. OK, though why would the Surefire JVM be seeing a half-written file to begin with? Did whatever code writes this file neglect to save to a temporary file and atomically rename it?
> * Suppressing some exceptions, which sounds wrong.
> * Some other changes of unclear importance.

In general this PR fixes 3 different problems:

1. stop `proc` once.
2. fix empty port failure (the main reason tests are flaky)
3. avoid lose an exception when finally throws it’s own exception (I open for a discussion, if we don't want that change I could revert it)

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
Pldi23: I like this idea and applied changes in [2430d48](https://github.com/jenkinsci/jenkins-test-harness/pull/465/commits/2430d48729c5e81f362fba229c6f20bd0e316bf3)

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jtnord: > I lost a lot of time understanding root cause with empty port because first exception was lost when finally throw it’s own exception

Sounds like a bug somewhere, the furst exception should be a `supressed` exception of whatever was thrown in the `finally` 

Perhaps a bug in junit runner or surefire or somewhere else where it is ignoring supressed exceptions (I expect there could be multiple cases of this).

https://issues.apache.org/jira/browse/SUREFIRE-1772  possibly.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jglick: BTW you can use GitHub’s native mechanism for applying suggestions in batch, which has some nice properties:
* sets co-author
* automatically resolves comment

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jglick: > potentially it could be created with some security/access options

Why would that happen? Is this something you observed, or just speculated on?

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -608,25 +609,38 @@ private static boolean supportsPortFileName(String war) throws IOException {
 
     private static int readPort(File portFile) throws IOException {
         String s = FileUtils.readFileToString(portFile, StandardCharsets.UTF_8);
+        if (s.isEmpty()) {
+            LOGGER.warning(() -> String.format("PortFile: %s exists, but value is still not written", portFile.getAbsolutePath()));
+            return 0;
+        }
```
jglick, line 615: Again, where is the code that writes this file, and can it be improved to do so atomically? Please fix here if in JTH. This logic is OK as a workaround if that code is in Winstone and you file a patch for it but it would take a long time for actual tests to pick it up.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jglick: I think this thread can be resolved now?

jglick (approved):
My main concerns have been addressed here.

jtnord:
> * Handle an empty port file. OK, though why would the Surefire JVM be seeing a half-written file to begin with? Did whatever code writes this file neglect to save to a temporary file and atomically rename it?

yes, as per https://github.com/jenkinsci/jenkins-test-harness/pull/465#discussion_r932229715

>https://github.com/jenkinsci/winstone/blob/4754da67bf797b906a5c858b2de63b97ac609543/src/main/java/winstone/Launcher.java#L232-L234 is indeed a source of potential flakyness (that may or may not be your flake) but it looks like it could be.

> The file and its contents are not created atomically - so if you get very unlucky the file will exist before it is being written to.
The upstream could be changed (maybe it should) to use an atomic write (write and flush to a intermediate file, then move the file in place with an atomic flag set, but this may really be overkill). if you want to fix the flakes in the interim then probably if the port number is not valid sleep 5 seconds and try again once only.

given the lag for getting a winstone change then integrated into jenkins and then LTS etc, having a workaround here is believe ok.

Could this PR be split so it was more obvious what each individual change was doing,

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jtnord: https://github.com/jenkinsci/winstone/blob/4754da67bf797b906a5c858b2de63b97ac609543/src/main/java/winstone/Launcher.java#L232-L234

jtnord (approved):
I'm not convinced of the change to `canRead` from `exists` for the port file, but given I do not expect to be trying to read a port file with a different security setting I also do not expect to be timing out for ever without the diagnosis of the `IOException`

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
Pldi23: speculatively, I've not observed it. I don't mind and will revert it

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jglick: OK, so then that should be patched.

jglick:
> Could this PR be split so it was more obvious what each individual change was doing,

That would be OK, though I am also OK with the larger PR so long as the title explains that this is fixing flakiness related to port reading and the description enumerates the changes contributing to that fix.

Given that there have been significant changes since https://github.com/jenkinsci/jenkins-test-harness/pull/465#issuecomment-1202087307 I guess it would be prudent to reverify effectiveness.

Pldi23:
1. History squashed into 2 commits, PR name and description was also updated. (I decided to leave 1 PR and don't split it into 2 to make it easier to run flakebusters automations for flaky test)
2. Submitted a [bug](https://github.com/jenkinsci/winstone/issues/248) for Winstone to track.
3. Plan to re-run flaky tests with a patch.

timja:
Given the patch is likely that big for winstone I think it would be worth attempting a real fix there, we can get it out quickly and probably into the next LTS as well.

jglick (approved):
Has it been (re-)tested at scale yet?

Pldi23:
> Has it been (re-)tested at scale yet?

In progress, no failures at that moment https://gauntlet-2.cloudbees.com/rosie/job/playground/job/flakebusters/job/selectors/job/pct-test-selector/49/testReport/ .

[RemoteArtifactCopyBuilderTest](https://github.com/cloudbees/operations-center-context/blob/master/src/test/java/com/cloudbees/opscenter/artifacts/RemoteArtifactCopyBuilderTest.java#L98) are slow, I expect it to finish tomorrow.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
basil: If a PR was filed to fix the underlying problem in Winstone, I could accept this workaround as a temporary measure. But since no PR has been filed to fix the underlying problem in Winstone, I am concerned that accepting this workaround will decrease the incentive to implement a fix for the underlying problem in Winstone. In the absence of such a fix to Winstone this workaround may not be temporary but rather it may become permanent (i.e., technical debt). The Winstone fix seems trivial and could likely be coded in less than an hour, so my preference would be for it to be implemented prior to the addition of this workaround.

Pldi23: @basil I created a [PR#249](https://github.com/jenkinsci/winstone/pull/249 ) in Winstone to unblock

basil: Great, and thank you! Could we now add a comment to the (indeed, temporary!) workaround being added in this PR that says something to the effect of:

> Work around […] in Winstone releases prior to […]. When Winstone […] has been widely adopted, this can be deleted.

The reason I find it important to leave such comments is that without them, it is difficult to know when it is safe to remove the workaround in the future.

Pldi23: comment added in [1479165](https://github.com/jenkinsci/jenkins-test-harness/pull/465/commits/14791652245ac29f5a9ef658af3b0354146cbdd1)

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -609,10 +609,14 @@ private static boolean supportsPortFileName(String war) throws IOException {
 
     private static int readPort(File portFile) throws IOException {
         String s = FileUtils.readFileToString(portFile, StandardCharsets.UTF_8);
+
+        // Work around to non-atomic write of port value in Winstone releases prior to 6.1.
+        // Todo When Winstone 6.2 has been widely adopted, this can be deleted.
```
timja, line 614: generally todo is all caps
```suggestion
        // TODO When Winstone 6.2 has been widely adopted, this can be deleted.
```

Pldi23:
Re-Tested successfully, all of 7200 runs are green

jtnord (approved):

jglick (approved):

PR #491, opened 2022-09-21
Weed out flaky tests

basil:


basil (commented):
No flaky tests in this run or in the 5 most recent builds of the main branch, so let us prevent any from being introduced in the future by removing this option.

PR #492, opened 2022-09-27
Make `JenkinsRule#after` more robust

basil:
Observed this flake:

```
[ERROR] Tests run: 35, Failures: 0, Errors: 1, Skipped: 1, Time elapsed: 112.835 s <<< FAILURE! - in hudson.model.ViewTest
[ERROR] hudson.model.ViewTest.shouldNotAllowInconsistentViewName  Time elapsed: 4.426 s  <<< ERROR!
java.lang.NullPointerException: Cannot invoke "com.gargoylesoftware.htmlunit.html.DomElement.getElementsByTagName(String)" because "doc" is null
      at com.gargoylesoftware.htmlunit.html.HtmlPage.executeDeferredScriptsIfNeeded(HtmlPage.java:1498)
      at com.gargoylesoftware.htmlunit.html.HtmlPage.initialize(HtmlPage.java:282)
      at com.gargoylesoftware.htmlunit.WebClient.loadWebResponseInto(WebClient.java:708)
      at com.gargoylesoftware.htmlunit.WebClient.loadWebResponseInto(WebClient.java:590)
      at com.gargoylesoftware.htmlunit.WebClient.getPage(WebClient.java:508)
      at com.gargoylesoftware.htmlunit.WebClient.getPage(WebClient.java:415)
      at com.gargoylesoftware.htmlunit.WebClient.getPage(WebClient.java:553)
      at com.gargoylesoftware.htmlunit.WebClient.getPage(WebClient.java:535)
      at org.jvnet.hudson.test.JenkinsRule$WebClient.getPage(JenkinsRule.java:2622)
      at org.jvnet.hudson.test.JenkinsRule.after(JenkinsRule.java:511)
      at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:631)
```

The failure makes no sense at first blush: after parsing the empty string for "about:blank" into an empty DOM, we should trivially be able to fetch the document element in `com.gargoylesoftware.htmlunit.html.HtmlPage#getDocumentElement` and see that it has no `<script>` elements in `com.gargoylesoftware.htmlunit.html.HtmlPage#executeDeferredScriptsIfNeeded` before chugging right along. And indeed that is what happens when I step through in a debugger. So why is this getting clobbered with null in production?

Stepping through the code that led us up to that point sniffing for asynchronous activity, I found some as expected. The main teardown logic in HtmlUnit is `com.gargoylesoftware.htmlunit.WebWindowImpl#destroyChildren` which effectively calls `getJobManager().removeAllJobs()` and `getJobManager().shutdown()`. The former is documented: "removes all jobs from the execution queue. This doesn't interrupt any jobs that may be currently running." The latter is effectively a no-op. So calling this method prevents new asynchronous tasks from being queued, but it does nothing to stop the tasks that are already running. So I suspect that there is some asynchronous JavaScript task that is looping through the DOM and removing some elements, and it continues running even as the DOM is replaced with the shiny clean DOM from "about:blank". The asynchronous JavaScript task keeps running, removing the root element of the clean DOM until there is nothing left, then finally we reach `HtmlPage#initialize` to do what should be a mundane no-op but blow up because the DOM has been clobbered by now. I do not have evidence in support of this theory, but after reading most of the code I cannot see any other races by visual inspection. I have no idea what might be looping through and removing DOM nodes but the point is that any asynchronous activity that keeps running after reloading the page is a liability.

What to do about this? HtmlUnit does not seem to provide an API to atomically cancel existing tasks and stop accepting new ones, which is what we really want. And in fact `getJobManager().removeAllJobs()` makes it impossible to wait on the existing tasks. So our only recourse is to wait for existing tasks for a bounded amount of time and then stop accepting new ones, hoping that nothing gets scheduled between the time we finish waiting and the time we stop accepting new tasks.

While I was here I noticed that the `IOException` being caught could never have been thrown, as the "about:blank" page is an empty `String` without the potential for system calls. So rather than swallow it I am upgrading it to an `UncheckedIOException` which is effectively an assertion. Not really related to the core change, but should make the code easier to reason about.

### Testing done

None yet. I intend to run some core tests in a loop with and without this change to see if it makes a tangible difference in the rate of occurrence of this bug (which I haven't seen very often to begin with). But either way it seems like a generally positive change to make that can't hurt, as long as I don't notice any novel regressions.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -132,6 +132,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
+import java.io.UncheckedIOException;
 import java.lang.annotation.Annotation;
 import java.lang.management.ThreadInfo;
 import java.lang.reflect.Array;
@@ -504,15 +505,19 @@ public class JenkinsRule implements TestRule, MethodRule, RootAction {
                     tl.onTearDown();
             }
 
-            // cancel pending asynchronous operations, although this doesn't really seem to be working
+            // cancel asynchronous operations as best as we can
             for (WebClient client : clients) {
-                // unload the page to cancel asynchronous operations
+                // wait until current asynchronous operations have finished executing
+                WebClientUtil.waitForJSExec(client);
+                // unload the page to prevent new asynchronous operations from being scheduled
                 try {
                     client.getPage("about:blank");
                 } catch (IOException e) {
-                    // ignore
+                    // should never happen when loading "about:blank"
+                    throw new UncheckedIOException(e);
+                } finally {
+                    client.close();
                 }
-                client.close();
             }
             clients.clear();
 


basil:
This PR passed a normal Jenkins core PR build. As far as the `ViewTest` flake is concerned, I was able to reproduce the flake after running the test in a loop for 11 hours or so. So far I have run this test in a loop for 9 hours without being able to reproduce the flake. I just kicked off another 5 hour loop, so fingers crossed that the flake doesn't show up.

timja (approved):

PR #493, opened 2022-09-29
Improve `InboundAgentRule`

basil:
Various improvements to `InboundAgentRule`:

- Support WebSocket
- Support secrets
- Support skipping agent start
- Improve determinism by waiting for `SlaveComputer#_connect` to finish after adding the agent but before starting it
- Add work directory options
- Use thread-safe data structures

The motivation behind this is to enable most tests in Jenkins core to be converted to this rule. That way flakiness fixes can be implemented in one place and apply to all tests, which isn't possible with the copypasta we have today.

Once I get an incremental I'll post a core PR demonstrating how this is used.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
+++ b/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
@@ -211,7 +211,6 @@ public final class InboundAgentRule extends ExternalResource {
         }
         if (options.isStart()) {
             start(r, options);
-            r.waitOnline(s);
         }
         return s;
     }
--- a/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
+++ b/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
@@ -31,13 +31,15 @@ import hudson.model.Slave;
 import hudson.slaves.DumbSlave;
 import hudson.slaves.JNLPLauncher;
 import hudson.slaves.RetentionStrategy;
+import hudson.slaves.SlaveComputer;
 import hudson.util.StreamCopyThread;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import org.apache.commons.io.FileUtils;
 import org.apache.tools.ant.util.JavaEnvUtils;
 import org.junit.rules.ExternalResource;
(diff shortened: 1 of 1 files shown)


basil:
So far this has passed a regular Jenkins core test run as well as running all `InboundAgentRule` tests in a loop for 5 hours with no flakes. I'll keep on running the latter tests in a loop, but this PR looks promising so far.

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
```diff
@@ -50,32 +52,184 @@
...
-    private final Map<String, Process> procs = new HashMap<>();
+    private final ConcurrentMap<String, Process> procs = new ConcurrentHashMap<>();
+
+    /**
+     * The options used to (re)start an inbound agent.
+     */
+    public static class Options {
+        // TODO Java 14+ use records
+
+        @CheckForNull private String name;
```
timja, line 63: personal preference would be to move the annotation above it, it looks strange to me as it's in the same grouping of variables and feels mis-aligned. (feel free to ignore)

```suggestion
        @CheckForNull
        private String name;
```

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
```diff
@@ -50,32 +52,184 @@
...
+            this.start = start;
+        }
+
+        /**
+         * A builder of {@link Options}.
+         *
+         * <p>Instances of {@link Builder} are created by calling {@link
+         * InboundAgentRule.Options#newBuilder}.
+         */
+        public interface Builder {
```
timja, line 106: would you be able to explain why you went for an interface for a builder? It's not a pattern I'm familiar with (of course there's many slightly different ways to do builders).

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
```diff
@@ -50,32 +52,184 @@
...
-        s.setNumExecutors(1); // TODO 2.234+ delete (default switched from 2 to 1)
+        JNLPLauncher launcher = new JNLPLauncher(true);
+        launcher.setWebSocket(options.isWebSocket());
+        DumbSlave s = new DumbSlave(options.getName(), new File(r.jenkins.getRootDir(), "agent-work-dirs/" + options.getName()).getAbsolutePath(), launcher);
         s.setRetentionStrategy(RetentionStrategy.NOOP);
         r.jenkins.addNode(s);
-        start(r, name);
+        // SlaveComputer#_connect runs asynchronously. Wait for it to finish for a more deterministic test.
+        while (s.toComputer().getOfflineCause() == null) {
+            Thread.sleep(100);
```
timja, line 210: is there a chance this never happens and we end up with an infinite loop? I guess the test timeout would kick in eventually.

timja (approved):

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
basil: `google-java-format` puts them on the same line if the whole line would be within the margin, and I have seen another argument that they should be on the same line because type annotations should be close to their corresponding type. I do not care much either way.

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
jglick: I also struggled to understand this (looking into https://github.com/jenkinsci/jenkins-test-harness/pull/514#issuecomment-1319716770). Conventionally builders would be concrete `final` classes. AFAICT the only implementation is in fact `BuilderImpl`?

basil: If you do not like it, then change it.

PR #499, opened 2022-10-13
Add jenkins-core to dependencyManagement to avoid pulling all stapler versions

Vlatombe:
`maven-enforcer-plugin` `requireUpperBoundsCheck` rule will retrieve transitive dependencies of all plugins specified in the pom, which use various versions of jenkins core. These depend on old versions of stapler-adjunct-xxx, which used to depend on stapler using a version range. This causes indirectly to retrieve all versions of stapler, causing unnecessary additional time to download.

Adding `jenkins-core` to `dependencyManagement` directly allows to cut this since no older jenkins version will be resolved, and thus no old stapler artifacts.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,6 +80,12 @@ THE SOFTWARE.
         <artifactId>access-modifier-annotation</artifactId>
         <version>1.27</version>
       </dependency>
+      <dependency>
+        <!-- Avoid downloading multiple jenkins core versions and tons of stapler versions due to version range usage in older libs -->
+        <groupId>org.jenkins-ci.main</groupId>
+        <artifactId>jenkins-core</artifactId>
+        <version>${jenkins.version}</version>
+      </dependency>
     </dependencies>
   </dependencyManagement>
 


jglick:
> all plugins specified in the pom

Specified in which POM? This project’s POM does not mention any (Jenkins) plugins.

Is there a known way to reproduce this which would let us track down exactly where the old version of Jenkins core is coming from?

Somehow related to https://github.com/stapler/stapler-adjunct-jquery/pull/2 via https://github.com/jenkinsci/stapler/pull/186 I suppose.

jglick (approved):
Anyway, seems like it could not hurt. https://github.com/jenkinsci/jenkins-test-harness/blob/5dc1b6cd75adbea7695c435a1613f5d5485191f4/pom.xml#L93-L102 specifies `optional` so it is not enforcing a core version directly AFAICT.

jglick:
> it must look up the detached plugins

Seems plausible. Some of the detached versions of course have very old core deps.

basil:
Thanks for fixing this! This had been annoying me for some time.

PR #501, opened 2022-10-14
Wait for inbound agent to go offline

basil:
I have been seeing a lot of flakiness in `InboundAgentRule` based tests in `#after` methods and I think the cause could be that we aren't waiting for the inbound agent to go offline before starting to tear down the test. This should solve the problem by generalizing https://github.com/jenkinsci/jenkins/blob/5b5040298f5fbf9d5ff7ea7fbfd34f7169e94d75/test/src/test/java/jenkins/agents/WebSocketAgentsTest.java#L109-L113 to `InboundAgentRule`. The new `stop` method is analogous to the existing `#start(JenkinsRule, String)` method. Once I get an incremental build I plan to file a Jenkins core PR consuming this new functionality.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
+++ b/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
@@ -27,6 +27,7 @@ package org.jvnet.hudson.test;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import hudson.model.Computer;
 import hudson.model.Slave;
 import hudson.slaves.DumbSlave;
 import hudson.slaves.JNLPLauncher;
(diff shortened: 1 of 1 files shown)


timja (approved):

jglick:
FTR what did a flake here look like—some error thrown from `after`? It is a pity to force callers to use a `finally` block, so I am wondering if another part of this patch would suffice to solve the issue or if you really need to wait for `Computer.online` status for some reason.

jglick:
`JenkinsRule.after` could just ignore exceptions from `ChannelShutdownListener` I guess. The real problem is I think that `WebSocketAgents.Session.Transport.write` (or `WebSocketSession.sendBinary` or `Jetty9Provider$1.sendBinary`) should be checking the socket state and tolerating the case that it is already closed (which is presumably harmless once `Channel.close` has run). @Vlatombe @amuniz WDYT?

basil:
Yeah maybe but I'm not sure how realistic of a use case it is to have two threads racing to close at the same time. In any case I think the tests should be as deterministic as possible, so I think it is desirable to let the normal close sequence finish in every test rather than potentially trying to race with it in `ChannelShutdownListener#onTearDown` based on the speed of the test system.

PR #510, opened 2022-11-11
Accept multiple `Step`s in `then` and `runRemotely`

jglick:
Sometimes when defining complex suites of tests using `RealJenkinsRule` it is helpful to be able to define reusable `Step` implementations (classes or method handles) and run a series of them at once. This makes that process slightly more convenient and faster.

### DIFF
--- a/src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
+++ b/src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
@@ -85,12 +85,6 @@ public class RealJenkinsRuleTest {
         assertEquals("value", System.getenv("SOME_ENV_VAR"));
     }
 
-    @Test public void testFilter() throws Throwable{
-        rr.startJenkins();
-        rr.runRemotely(RealJenkinsRuleTest::_testFilter);
-        rr.runRemotely(RealJenkinsRuleTest::_htmlUnit1);
-    }
-
     @Test public void testReturnObject() throws Throwable {
         rr.startJenkins();
         assertEquals(rr.getUrl().toExternalForm(), rr.runRemotely(RealJenkinsRuleTest::_getJenkinsUrlFromRemote));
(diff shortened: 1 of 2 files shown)


basil (approved):
I don't suppose it's worth updating the tests to exercise this as in e.g.

```java
diff --git a/src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java b/src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
index 6471f85..1500063 100644
--- a/src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
+++ b/src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
@@ -87,8 +87,7 @@ public class RealJenkinsRuleTest {
 
     @Test public void testFilter() throws Throwable{
         rr.startJenkins();
-        rr.runRemotely(RealJenkinsRuleTest::_testFilter);
-        rr.runRemotely(RealJenkinsRuleTest::_htmlUnit1);
+        rr.runRemotely(RealJenkinsRuleTest::_testFilter, RealJenkinsRuleTest::_htmlUnit1);
     }
 
     @Test public void testReturnObject() throws Throwable {
```

jglick:
Will need to check if that specific example works (the behavior is different—just one HTTP request rather than two), but yes it would be good to have at least elementary test coverage.

jglick:
> if that specific example works

It does not; see #323.

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
```diff
@@ -85,12 +85,6 @@ private static void _smokes(JenkinsRule r) throws Throwable {
         assertEquals("value", System.getenv("SOME_ENV_VAR"));
     }
 
-    @Test public void testFilter() throws Throwable{
```
jglick, line 88: Had gotten separated from its helper methods.

basil (approved):

PR #511, opened 2022-11-11
Do not suspend the controller JVM on startup when debugging

jglick:
When debugging `RealJenkinsRule`, being forced to attach to the controller before proceeding is more annoying than helpful. (If you do want to debug something during startup, set a breakpoint.)

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -519,7 +519,7 @@ public final class RealJenkinsRule implements TestRule {
             argv.add("-Dwinstone.portFileName=" + portFile);
         }
         if (new DisableOnDebug(null).isDebugging()) {
-            argv.add("-agentlib:jdwp=transport=dt_socket,server=y");
+            argv.add("-agentlib:jdwp=transport=dt_socket,server=y,suspend=n");
         }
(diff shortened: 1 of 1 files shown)


basil (approved):
Never bothered me too much either way, but seems fine.

Vlatombe (commented):
I would be in favor of extracting this so that options can be easily changed without changing corresponding code. Sometimes I need to debug harness code, sometimes the Jenkins instances on startup.

jglick:
> sometimes the Jenkins instances on startup

And setting a breakpoint in controller code does not suffice for this purpose?

> extracting this so that options can be easily changed without changing corresponding code

Not sure what you had in mind. Examples?

Vlatombe:
If the VM is not suspended on startup, it may be too late by the time the debugger is attached to the controller VM to hit the breakpoint.

> Examples?

Extract debug options to a field which can be set by the test caller (e.g. calling `rjr.debugOptions("-agentlib:jdwp=transport=dt_socket,server=y,suspend=y")`. Still require some code change but only in the test code, not in the harness.

jglick:
> Still require some code change but only in the test code

Changes to the test code do not seem acceptable. You should be able to debug a test however you like without editing it.

I do not see a way forward that would satisfy your use case, so closing and I guess I will live with the annoyance.

PR #513, opened 2022-11-11
Allow `RealJenkinsRule` controllers to know `surefire.forkNumber`

jglick:
This Maven property can be helpful to avoid conflicts between parallelized tests when using test services, if you cannot reliably randomize ports. Needs https://github.com/jenkinsci/plugin-pom/pull/645 to ensure the property gets propagated from Maven to the Surefire booter JVM.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -530,14 +530,23 @@ public final class RealJenkinsRule implements TestRule {
                 "--httpPort=" + port, // initially port=0. On subsequent runs, the port is set to the port used allocated randomly on the first run.
                 "--httpListenAddress=127.0.0.1",
                 "--prefix=/jenkins"));
-        ProcessBuilder pb = new ProcessBuilder(argv);
-        System.out.println("Launching: " + pb.command());
-        pb.environment().put("JENKINS_HOME", home.getAbsolutePath());
+        Map<String, String> env = new TreeMap<>();
+        env.put("JENKINS_HOME", home.getAbsolutePath());
+        String forkNumber = System.getProperty("surefire.forkNumber");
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -530,14 +530,23 @@ public void startJenkins() throws Throwable {
                 "--httpPort=" + port, // initially port=0. On subsequent runs, the port is set to the port used allocated randomly on the first run.
                 "--httpListenAddress=127.0.0.1",
                 "--prefix=/jenkins"));
-        ProcessBuilder pb = new ProcessBuilder(argv);
-        System.out.println("Launching: " + pb.command());
-        pb.environment().put("JENKINS_HOME", home.getAbsolutePath());
+        Map<String, String> env = new TreeMap<>();
```
basil, line 533: Nice touch to use something better than `HashMap` here (and consistent with `hudson.EnvVars`, which also uses `TreeMap`). It's too bad that `ProcessBuilder#environment` uses a `HashMap` internally, scrambling the order of whatever the user passes in. I wish `hudson.EnvVars` had extended `LinkedHashMap` (insertion order), because that is how libc's `setenv(3)` works, but that ship sailed a long time ago.

basil (approved):

Vlatombe (approved):

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jglick: As far as I know the ultimate order is irrelevant, but it is much nicer to print the variables in order below.

basil: > but it is much nicer to print the variables in order below.

… which is why I wrote it was a nice touch above.

PR #514, opened 2022-11-11
Allow `RealJenkinsRule` instances to specify a `name` & `color`

jglick:
At CloudBees we often multiple `RealJenkinsRule` rules simultaneously to model the interaction among multiple controllers. It is much easier to follow log output when log lines are prefixed with a controller identifier, and (when running interactively) are colored.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -27,6 +27,7 @@ package org.jvnet.hudson.test;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.ExtensionList;
+import hudson.console.LineTransformationOutputStream;
 import hudson.model.UnprotectedRootAction;
 import hudson.security.ACL;
 import hudson.security.ACLContext;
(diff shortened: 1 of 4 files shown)


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -541,10 +575,41 @@ public void startJenkins() throws Throwable {
...
+                        out.write(27);
+                        out.write('[');
+                        out.write('0');
+                        out.write('m');
+                        out.write(b, preNlLen, len - preNlLen);
+                    } else {
+                        out.write(b, 0, len);
+                    }
+                }
+            };
```
basil, line 610: I don't suppose it would be worth extracting this into a static class and writing a unit test for some common cases (e.g., regular line, empty line, Windows line break, etc.).

basil (approved):
Fancy!

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jglick: > extracting this into a static class

Or just a `static` method. Yes, sounds prudent.

Vlatombe:
These name and color logic could be reused for `InboundAgentRule`.

jglick:
Yes, I think that would work.

jglick:
Core PR demonstrating this coming as soon as I get an incremental build…

jglick:
https://github.com/jenkinsci/jenkins/pull/7411

PR #519, opened 2022-11-18
Add `BuildWatcher` support to `RealJenkinsRule`

Vlatombe:
Calling `rjr.watchBuilds()` will register BuildWatcher and print build logs.

Result can be observed by running `mvn test -Dtest=RealJenkinsRuleTest#agentBuild`

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -28,6 +28,7 @@ import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.ExtensionList;
 import hudson.model.UnprotectedRootAction;
+import hudson.model.listeners.RunListener;
 import hudson.security.ACL;
 import hudson.security.ACLContext;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -857,6 +875,9 @@ public void doStep(StaplerRequest req, StaplerResponse rsp) throws Throwable {
         }
         public HttpResponse doExit(@QueryParameter String token) throws IOException {
             checkToken(token);
+            if (buildWatcher != null) {
+                buildWatcher.after();
+            }
```
Vlatombe, line 880: Probably not strictly necessary since the jvm will shutdown just after.

jglick (commented):
OK but there is actually a better way which does not require anything to run inside the controller JVM and is more flexible. `BuildWatcher` as currently designed does not handle the case that the build is already running when the controller starts (or, in this case, when the Surefire JVM connects to it). That means it does not support `@LocalData`, and it only supports watching builds across `JenkinsSessionRule` restarts insofar as the JVM does not exit. I have a plan to write this up properly and support RJR. Not sure if this patch would conflict with that.

Vlatombe:
Superceded by #532

PR #528, opened 2022-11-30
Require Java 11 or newer

basil:
The first of several PRs to require Java 11 or newer. More cleanup is planned after this PR now that the baseline has been increased to 2.361.


timja (approved):

jglick:
Like #453 I suppose. :rocket:

PR #541, opened 2023-01-10
`RealJenkinsRule`: option to provide a custom port/listen address

Vlatombe:
Kubernetes-plugin is running its tests within a kubernetes pod running in Kind.

During test execution, agents need to loopback to the running controller running in another container, and so the controller needs to bind another network interface than localhost.

A [special](https://github.com/jenkinsci/kubernetes-plugin/blob/54f40a5e7fb1798bad69977b00fa8634260d8852/src/test/java/org/jvnet/hudson/test/JenkinsRuleNonLocalhost.java) version of JenkinsRule is used for regular tests, considering prior attempts to introduce such option were not merged.

For RealJenkinsRule-based tests, it seems much simpler to have the option available directly in the framework.

See usage in https://github.com/jenkinsci/kubernetes-plugin/pull/1290

Relates to https://github.com/jenkinsci/jenkins-test-harness/pull/289#issuecomment-809560892, for RJR.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -160,6 +160,15 @@ public final class RealJenkinsRule implements TestRule {
      */
     private int port;
 
+    /**
+     * HTTP interface the server listens to.
+     * <p>
+     * Defaults to 127.0.0.1.
+     * <p>
+     * Should be overridden only in specific testing use cases, like testing inside a docker container.
+     */
+    private String httpListenAddress = "127.0.0.1";
+
     private File war;
 
     private boolean includeTestClasspathPlugins = true;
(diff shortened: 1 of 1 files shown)


jglick:
Any relation to https://github.com/jenkinsci/jenkins-test-harness/pull/289#issuecomment-809560892?

Vlatombe:
Ah, possibly, due to https://github.com/jenkinsci/jenkins-test-harness/blob/8bea09f5abe2038ce26822a6f752aff3279fe1c9/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java#L559

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -315,6 +324,15 @@ public RealJenkinsRule withPort(int port) {
         return this;
     }
 
+    /**
+     * Provides a custom interface to listen to. This should be restricted for specific testing use cases (like running inside a docker container)
```
jglick, line 328:
```suggestion
     * Provides a custom interface to listen to.
     * <p><em>Important:</em> for security reasons this should be overridden only in special scenarios,
     * such as testing inside a Docker container.
     * Otherwise a developer running tests could inadvertently expose a Jenkins service without password protection,
     * allowing remote code execution.
```

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -160,6 +160,15 @@ public final class RealJenkinsRule implements TestRule {
      */
     private int port;
 
+    /**
+     * HTTP interface the server listens to.
+     * <p>
+     * Defaults to 127.0.0.1.
+     * <p>
+     * Should be overridden only in specific testing use cases, like testing inside a docker container.
+     */
```
jglick, line 169: Leave it to the `public` method.

```suggestion
```

jglick (approved):

PR #544, opened 2023-01-17
Use `ProcessHandle.destroy` to implement `RealJenkinsRule.stopJenkins`

jglick:
As of https://github.com/jenkinsci/winstone/pull/310 we will get final log messages even without an explicit call path to `Jenkins.cleanUp`, so there is no need for a special `exit` endpoint.


dwnusbaum:
Note that when using `RealJenkinsRule` in tests against plugins which have picked up this PR (i.e. parent POM >= 4.54) but not https://github.com/jenkinsci/jenkins/pull/7632 (i.e. Jenkins >= 2.391), shutdown-related log output is lost.

jglick:
Correct.

PR #558, opened 2023-02-28
Fix potential infinite loop in `InboundAgentRule`

jtnord:
waiting for the agent could be an infinite loop if the agent had not connected when we enter the retry (ie computer was null).

Also spotbugs was complaining about a redundant null check from `URLConnection.getInputStream()` as that will return a stream or throw an `UnknownServiceException` if input is not supported - but never null.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
+++ b/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
@@ -455,7 +455,7 @@ public final class InboundAgentRule extends ExternalResource {
             return options.getName();
         }
 
-        @SuppressFBWarnings(value = {"PATH_TRAVERSAL_IN", "NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE"}, justification = "just for test code")
+        @SuppressFBWarnings(value = "PATH_TRAVERSAL_IN", justification = "just for test code")
         private static Slave createAgent(JenkinsRule r, Options options) throws Descriptor.FormException, IOException, InterruptedException {
             if (options.getName() == null) {
                 options.setName("agent" + r.jenkins.getNodes().size());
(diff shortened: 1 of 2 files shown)


src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
```diff
@@ -455,7 +455,7 @@ public String run(JenkinsRule r) throws Throwable {
             return options.getName();
         }
 
-        @SuppressFBWarnings(value = {"PATH_TRAVERSAL_IN", "NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE"}, justification = "just for test code")
```
jtnord, line 458: `NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE` was a lie at any rate - if `computer` was null we would never return!

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
```diff
@@ -469,6 +469,7 @@ private static Slave createAgent(JenkinsRule r, Options options) throws Descript
             Computer computer = s.toComputer();
             while (computer == null || computer.getOfflineCause() == null) {
                 Thread.sleep(100);
+                computer = s.toComputer();
```
jtnord, line 472: `computer` cloud be `null` - so we need to recheck to avoid an infinite loop.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -1297,16 +1297,9 @@ public JSONWebResponse postJSON(@NonNull String path, @NonNull Object json) thro
             }
 
             WebResponseData webResponseData;
-            InputStream responseStream = conn.getInputStream();
```
jtnord, line 1300: [`conn.getInputStream()`](https://docs.oracle.com/javase/8/docs/api/java/net/URLConnection.html#getInputStream--) will never return `null` - unless some code is violating the API contract.
Thus leading to spotbugs complaining about `RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE` 

this rest of the diff moves to `try-with-resources` to cleanup the code knowing that the null check is not needed.

jglick (approved):
Looks right.

Vlatombe (approved):

PR #559, opened 2023-03-06
Revert #544 so that `RealJenkinsRule` shuts down cleanly on Windows

dwnusbaum:
Based on CI build failures in https://github.com/jenkinsci/workflow-cps-plugin/pull/669 (see for example [this test output](https://ci.jenkins.io/job/Plugins/job/workflow-cps-plugin/job/PR-669/5/testReport/org.jenkinsci.plugins.workflow.cps/CpsFlowDefinitionRJRTest/windows_11___Build__windows_11____flushLogsOnShutdown/)), I am suspicious that beyond the lack of shutdown-related log messages on older versions of Jenkins, https://github.com/jenkinsci/jenkins-test-harness/pull/544 might not shut down Jenkins cleanly on Windows, in particular I think `Jenkins.cleanUp` might not run and terminators may be skipped.

This PR injects an `ItemListener` dynamically into Jenkins using `ExtensionList.add` that writes a file in `onBeforeShutdown`, which is called by `Jenkins.cleanUp`, and after the restart we check for the existence of this file. Maybe there is something else we can already observe without needing this, but without `LoggerRule` I couldn't think of anything obvious.

`RealJenkinsRuleTest.restart` passes for me locally on MacOS with this PR.

EDIT: The tests confirmed my suspicions, so this PR now reverts #544 so that `RealJenkinsRule` shuts down cleanly on Windows.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -95,6 +95,7 @@ import org.junit.rules.Timeout;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
 import org.jvnet.hudson.test.recipes.LocalData;
+import org.kohsuke.stapler.HttpResponse;
 import org.kohsuke.stapler.HttpResponses;
 import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.Stapler;
@@ -756,13 +757,16 @@ public final class RealJenkinsRule implements TestRule {
         if (proc != null) {
             Process _proc = proc;
             proc = null;
-            // Do not use Process.destroy as it closes streams, losing any final output:
-            _proc.toHandle().destroy();
+            endpoint("exit").openStream().close();
 
             if (!_proc.waitFor(60, TimeUnit.SECONDS) ) {
                 System.err.println("Jenkins failed to stop within 60 seconds, attempting to kill the Jenkins process");
                 _proc.destroyForcibly();
                 throw new AssertionError("Jenkins failed to terminate within 60 seconds");
+            }
+            int exitValue = _proc.exitValue();
+            if (exitValue != 0) {
+                throw new AssertionError("nonzero exit code: " + exitValue);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


dwnusbaum:
Hmm the CI build does fail on Windows but not on Linux. @jglick was https://github.com/jenkinsci/jenkins-test-harness/pull/544 necessary to fix an issue in some scenario, or was it just a simplification that could be reverted? As far as I can tell, that PR causes `RealJenkinsRule.stopJenkins` to not stop Jenkins cleanly on Windows, even when running Jenkins 2.391+.

jglick:
It could be reverted I think, but I would like to know stopping the process does not cause terminators to run—if true, that certainly sounds like a bug?

dwnusbaum:
> I would like to know why stopping the process does not cause terminators to run—if true, that certainly sounds like a bug?

Discussion in [JDK-4485742](https://bugs.openjdk.org/browse/JDK-4485742) and linked tickets indicates that `Process.destroy` terminates the process in a way that bypasses Java shutdown hooks on Windows, and that this will not be changed due to backwards compatibility implications. Looking at JDK source code, `ProcessHandle.destroy` seems to use the same native Windows API calls (compare [this](https://github.com/openjdk/jdk/blob/dfacda488bfbe2e11e8d607a6d08527710286982/src/java.base/windows/native/libjava/ProcessHandleImpl_win.c#L369) to [this](https://github.com/openjdk/jdk/blob/dfacda488bfbe2e11e8d607a6d08527710286982/src/java.base/windows/native/libjava/ProcessImpl_md.c#L454)), so I assume it has the same behavior.

jglick:
Also [JDK-8056139](https://bugs.openjdk.org/browse/JDK-8056139). Hmm, that is unfortunate. We could fully revert #544, or continue to send SIGTERM on Linux and just use the special `/exit` endpoint on Windows.

jglick:
> any reason to prefer `ProcessHandle.destroy`

Merely to avoid a gratuitous use of a different code path than what would often be used in a realistic context. I agree that having two modes is more confusing that it would be worth, so reverting #544 would be the simplest fix. Feel free to file that (it sounds like you already have some downstream work which would verify the effectiveness).

dwnusbaum:
Ok, I push the revert to this PR. 

> it sounds like you already have some downstream work which would verify the effectiveness

I think the updates to `RealJenkinsRuleTest.restart` here are enough the validate the fix and prevent future regressions.

jglick:
> the updates to `RealJenkinsRuleTest.restart` here are enough the validate the fix

You mean, when run on Windows.

jglick (approved):

dwnusbaum:
> You mean, when run on Windows.

Yes, but the CI build here does test against Windows.

PR #575, opened 2023-05-02
`WarExploder` does not find the Jenkins WAR in the Gradle cache if its SHA-1 hash begins with a 0

### LINKED ISSUE #574: WarExploder doesn't account for Gradle not always generating valid SHA1s in cache path
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  
  ```text
  Paste the output here
  ```

</details>

### What Operating System are you using (both controller, and any agents involved in the problem)?

N/A

### Reproduction steps

Try to run the test harness against `jenkins-war:2.332.3`.

### Expected Results

Test harness functions properly.

### Actual Results

`WarExploder` throws an `AssertionError` saying
> ~/.gradle/caches/modules-2/files-2.1/org.jenkins-ci.main/jenkins-war/2.332.3/48e994349a6ea9fff614dca79dd3c5f75c0f3d16/jenkins-war-2.332.3.war does not yet exist. Prime your development environment by running `mvn validate`.

This issue occurs because gradle's dependency caching trims leading zeros from the hash in the file cache. The SHA1 for jenkins-war 2.332.3 is `068b1e10fdddae7cad6f2b59c8c205d285321940` (see image)
image

but the file is stored on disk in `~/.gradle/caches/modules-2/files-2.1/org.jenkins-ci.main/jenkins-war/2.332.3/68b1e10fdddae7cad6f2b59c8c205d285321940/jenkins-war-2.332.3.war`

This issue was raised with the gradle team in 2014 [here](https://discuss.gradle.org/t/sha-1-checksum-portion-of-file-path-in-gradle-cache-is-missing-leading-zeros/1244/2) with the response effectively being that it's an internal API with no guarantees.

I'd suggest implementing a conditional check to see if the directory hash is missing a character (stringified SHA1s are 40 characters and SHA256s are 64) that you prepend a `0` and test again.

### Anything else?

_No response_

dgreene-r7:
Hi!

This PR fixes an issue where the `WarExploder` is unable to find a copy of `jenkins-war` in the Gradle cache on disk if the checksum for the specified version begins with a zero. For some reason the fine folks working on Gradle decided to drop leading zeros from checksums. According to https://discuss.gradle.org/t/sha-1-checksum-portion-of-file-path-in-gradle-cache-is-missing-leading-zeros/1244/2 it could be to save characters in order to respect file path limitations. Either way, because the Gradle cache is an internal API, there's no guarantees in the way the test harness is trying to use it.

### Testing done

I compiled my changes and included them into a Jenkins JobDSL repo we have where I noticed the issue earlier today. With version `1954.v2e3fd5465b_a_6` I saw the following error when attempting to run tests:

```
java.lang.AssertionError: /Users/dgreene/.gradle/caches/modules-2/files-2.1/org.jenkins-ci.main/jenkins-war/2.332.3/jenkins-war-2.332.3.war does not yet exist. Prime your development environment by running `mvn validate`.
	at org.jvnet.hudson.test.WarExploder.findJenkinsWar(WarExploder.java:115)
	at org.jvnet.hudson.test.WarExploder.explode(WarExploder.java:146)
	at org.jvnet.hudson.test.WarExploder.getExplodedDir(WarExploder.java:62)
	at org.jvnet.hudson.test.JenkinsRule._createWebServer(JenkinsRule.java:797)
	at org.jvnet.hudson.test.JenkinsRule.createWebServer(JenkinsRule.java:752)
	at org.jvnet.hudson.test.JenkinsRule.createWebServer(JenkinsRule.java:741)
	at org.jvnet.hudson.test.JenkinsRule.newHudson(JenkinsRule.java:693)
	at org.jvnet.hudson.test.JenkinsRule.before(JenkinsRule.java:407)
	at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:602)
	at java.base/java.lang.Thread.run(Thread.java:829) 
```

This is the contents of the directory on disk:
```
/Users/dgreene/.gradle/caches/modules-2/files-2.1/org.jenkins-ci.main/jenkins-war/2.332.3/
├── 68b1e10fdddae7cad6f2b59c8c205d285321940
│   └── jenkins-war-2.332.3.war
└── 94b5703d50ba22c43604ee89ea706535098d782d
    └── jenkins-war-2.332.3-sources.jar
```

I've included `jenkins-war` into the project by adding
```
implementation "org.jenkins-ci.main:jenkins-war:2.3332.3@war"
```
to my `build.gradle` file.

Using the snapshot build, the `jenkins-war` war file was correctly found on disk and was able to be exploded.

This _should_ be a no-op for Maven-based projects.

### Submitter checklist

- [X] Link to relevant issues in GitHub or Jira

Fixes #574

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/WarExploder.java
+++ b/src/main/java/org/jvnet/hudson/test/WarExploder.java
@@ -30,7 +30,6 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.net.URL;
-import java.nio.file.Path;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
@@ -98,10 +97,10 @@ public final class WarExploder {
                     coreArtifactDir = core.getParentFile().getParentFile();
                 }
                 if (core.getName().equals("jenkins-core-" + version + ".jar") && coreArtifactDir.getName().equals("jenkins-core")) {
-                    File versionedWarArtifactDir = Path.of(coreArtifactDir.getParentFile().getPath(), "jenkins-war", version).toFile();
-                    war = Path.of(versionedWarArtifactDir.getPath(), "jenkins-war-" + version + ".war").toFile();
+                    File warArtifactDir = new File(coreArtifactDir.getParentFile(), "jenkins-war");
+                    war = new File(new File(warArtifactDir, version), "jenkins-war-" + version + ".war");
                     if (!war.isFile()) {
-                        File[] hashes = versionedWarArtifactDir.listFiles();
+                        File[] hashes = new File(warArtifactDir, version).listFiles();
                         if (hashes != null) {
                             for (File hash : hashes) {
                                 if (HEX_DIGITS.matcher(hash.getName()).matches()) {
--- a/src/main/java/org/jvnet/hudson/test/WarExploder.java
+++ b/src/main/java/org/jvnet/hudson/test/WarExploder.java
@@ -30,6 +30,7 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.net.URL;
+import java.nio.file.Path;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
@@ -53,9 +54,9 @@ public final class WarExploder {
     private static final String JENKINS_WAR_PATH = System.getProperty(JENKINS_WAR_PATH_PROPERTY_NAME);
 
     /**
-     * A pattern that matches either SHA-1 or SHA-256 hashes.
+     * A pattern that matches hex encoded strings.
      */
-    private static final Pattern SHA_HASH = Pattern.compile("^[a-f0-9]{40}$|^[a-f0-9]{64}$");
+    private static final Pattern HEX_DIGITS = Pattern.compile("^[a-f0-9]+$");
 
     public static synchronized File getExplodedDir() throws Exception {
         if (EXPLODE_DIR == null) {
@@ -87,7 +88,7 @@ public final class WarExploder {
                 File core = Which.jarFile(Jenkins.class); // will fail with IllegalArgumentException if have neither jenkins-war.war nor jenkins-core.jar in ${java.class.path}
                 String version;
                 File coreArtifactDir;
-                if (SHA_HASH.matcher(core.getParentFile().getName()).matches()) {
+                if (HEX_DIGITS.matcher(core.getParentFile().getName()).matches()) {
                     // Gradle
                     version = core.getParentFile().getParentFile().getName();
                     coreArtifactDir = core.getParentFile().getParentFile().getParentFile();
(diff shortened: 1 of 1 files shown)


basil (commented):
Thanks for the PR.

Based on my reading of the Gradle source code, I think there can be more than one leading zero in which case all of them are stripped. So this PR doesn't fully solve the issue.

I think a simpler approach would be to loosen up the logic that looks for a SHA-1 or SHA-256 hash to account for hashes that have leading zeros stripped by eliminating the requirement for a particular number of digits. That is:

```java
     /**
-     * A pattern that matches either SHA-1 or SHA-256 hashes.
+     * A pattern that matches a string of hexadecimal digits. Note that since Gradle can omit the
+     * leading zeros, we do not look for any specific number of digits.
      */
-    private static final Pattern SHA_HASH = Pattern.compile("^[a-f0-9]{40}$|^[a-f0-9]{64}$");
+    private static final Pattern HEX_DIGITS = Pattern.compile("^[a-f0-9]+$");
```

and replacing any reference to `SHA_HASH` with a reference to `HEX_DIGITS`. If this resolves your issue, I think it would be preferable to the current solution as it is both simpler and more robust. This shouldn't match any Jenkins group IDs or artifact IDs (because they have non-hex characters) or versions (because they all have dots).

dgreene-r7:
I can definitely pivot to that approach.

> Based on my reading of the Gradle source code, I think there can be more than one leading zero in which case all of them are stripped

Just out of curiosity, where did you find the handling in the Gradle source? I spent a little time poking through yesterday and couldn't find where they handle constructing the path.

basil:
> I can definitely pivot to that approach.

Great!

> Just out of curiosity, where did you find the handling in the Gradle source?

https://github.com/gradle/gradle/blob/1875b96ea5b963a100a5bf27859e07a0e5d2bf16/subprojects/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/caching/CachedDependencyResolutionIntegrationTest.groovy#L221-L235

dgreene-r7:
I had a think on this a little bit and frankly the whole war finding aspect is fraught with brittleness because it's trying to assemble a path from unknowns rather than just relying on traversing the file system.

I wish there were a better way to determine the Jenkins version so we could search the gradle cache without making assumptions about its structure. I didn't put a lot of time into it so there might be but in the meantime I've updated this PR with a slightly different approach. Let me know if you'd prefer I just use the hex character regex and I'll go back to that.

basil:
As you wrote above, this seems to be trying to implement a better search algorithm but doesn't go all the way. It's still using the hash regex in some places. If we're going to use the hash regex, I think we should consistently use it in both places as I originally suggested. If you want to implement a better search algorithm, that is fine, but I think it would be more work than what is currently done in this PR and should not use the hash regex at all.

dgreene-r7:
Yeah that's fair. I've switched back to the hex regex. Ultimately I think a FS-traversal approach with whatever special considerations need to be made against Maven and Gradle is the best approach but maybe that's something I can look at another time.

basil (approved):
Thanks for the PR!

PR #577, opened 2023-05-08
Wait for `InboundAgentRule` to be online as part of starting it

jglick:


Vlatombe (approved):

jglick:
Ran all core tests using `InboundAgentRule` against this (merged with `master`) just to be sure. Note that https://github.com/jenkinsci/jenkins/blob/fd90fdf17f62e90d31ae21843c8f86e2f976a1ac/test/src/test/java/jenkins/slaves/restarter/JnlpSlaveRestarterInstallerTest.java#L67-L72 is not currently using the start option.

jglick:
Just noticed that `start(JenkinsRule, …)` is not using the same logic, which I guess it should be. Let me try that…

Vlatombe (approved):

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
```diff
@@ -321,6 +324,7 @@ public void start(@NonNull JenkinsRule r, Options options) throws Exception {
         Objects.requireNonNull(name);
         stop(r, name);
         start(GetAgentArguments.get(r, name), options);
+        WaitForAgentOnline.wait(r, name);
```
basil, line 327: The Javadoc at the top of this file, which instructs consumers to call `r.waitOnline(agent)`, is now out of date, as are all cases of this pattern in Jenkins core.

PR #587, opened 2023-05-21
Remove deprecated `hamcrest-core` and `hamcrest-library` modules

NotMyFault:
Both modules named are deprecated since 2019, in favor of `hamcrest` where the logic of both modules live on since then.
Consumers are advised to replace hamcrest-x with hamcrest, if not already done.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -188,16 +188,6 @@ THE SOFTWARE.
       <version>${hamcrest.version}</version>
     </dependency>
     <dependency>
-      <groupId>org.hamcrest</groupId>
-      <artifactId>hamcrest-core</artifactId>
-      <version>${hamcrest.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>org.hamcrest</groupId>
-      <artifactId>hamcrest-library</artifactId>
-      <version>${hamcrest.version}</version>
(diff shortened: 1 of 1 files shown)


MarkEWaite (approved):
Thanks!

MarkEWaite:
@NotMyFault will this affect users of hamcrest in automated tests of plugins?

If they already have a declared dependency on the deprecated libraries, will they be able to continue to use those libraries even though they are no longer a dependency of Jenkins test harness?

If they don't have a declared dependency on the deprecated libraries, but are using them, will they be able to resolve that by declaring the dependency themselves?  I assume they could also resolve the issue by switching to use `hamcrest` rather than the deprecated libraries.

NotMyFault:
> I assume they could also resolve the issue by switching to use hamcrest rather than the deprecated libraries.

That's the desired approach. It contains the code of both deprecated libraries. Consumers are advised to remove both from their pom and add `hamcrest` only if not already done.

You can continue using the deprecated libraries for sure, if you add a version to the deprecated libraries overriding the JTH (or any other) dependencies.
That's the undesirable approach, given the modules haven't been touched in 4 years and there's no reason to not move on to `hamcrest`.

PR #588, opened 2023-05-23
Replace hamcrest matchers with methods from non-deprecated packages

NotMyFault:
With https://github.com/jenkinsci/jenkins-test-harness/pull/587 in mind, we can make use of the methods from the `hamcrest` package, which is the drop-in replacement for the deprecated -core package.

The code from the deprecated package has been merged in, and the methods return the same, nonetheless I applied the `breaking` label because the method signature changes.
Testing with core in https://github.com/jenkinsci/jenkins/pull/8046 revealed no issues.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsMatchers.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsMatchers.java
@@ -3,10 +3,10 @@ package org.jvnet.hudson.test;
 import hudson.util.FormValidation;
 import hudson.util.Secret;
 import org.hamcrest.BaseMatcher;
-import org.hamcrest.CoreMatchers;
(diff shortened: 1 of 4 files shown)


basil:
> I applied the `breaking` label because the method signature changes

Can you give an example of a method whose signature has changed?

NotMyFault:
Methods now return Matchers.doSth instead of CoreMatchers.doSth

basil:
In which method is a return value of `CoreMatchers` part of the method signature?

NotMyFault:
Adjusted the labels, the underlying hamcrest methods return the same. Looks like I read it wrong before.

basil (approved):
Thanks for the PR!

PR #591, opened 2023-05-27
Dispose of Windows read-only temporary files

MarkEWaite:
## Dispose of Windows temporary read-only files

When the git plugin and the gitlab plugin upgraded to parent pom 4.64, their tests failed on Windows. The Java delete() method on Windows throws an access denied exception when asked to delete a read-only file.  Command line git creates read-only files in the objects directory when it clones a repository.

This test creates a read-only file and relies on the JenkinsRule cleanup process to remove the file.

### Testing done

[Draft pull request](https://github.com/jenkinsci/git-plugin/pull/1454) submitted to the git plugin confirming that clearing the "read-only" permission on Windows files allowed the 100+ failing tests to pass with parent pom 4.64.

Automated test written that confirms read-only files are deleted as expected on Unix and that read-only files are not deleted on Windows.

The [first commit](https://github.com/jenkinsci/jenkins-test-harness/pull/591/commits/9eee796708cc43f80d15a6b28aaf316cb84fc2ef) is a failing test.  [Second commit](https://github.com/jenkinsci/jenkins-test-harness/pull/591/commits/13e45e0632cdf41e5d862be32ffc1efce88da29e) is the code changes to pass the failing test.

### Submitter checklist

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
+++ b/src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
@@ -133,6 +133,11 @@ public class TemporaryDirectoryAllocator {
             }
         }
         try {
+            if (isWindows()) {
+                // Windows throws an access denied exception when deleting read-only files
+                boolean ok = p.toFile().setWritable(true);
+                LOGGER.fine(() -> "allow write to " + p + ", result: " + ok);
+            }
             Files.deleteIfExists(p);
         } catch (DirectoryNotEmptyException x) {
             try (Stream<Path> children = Files.list(p)) {
@@ -141,4 +146,7 @@ public class TemporaryDirectoryAllocator {
         }
     }
 
+    private boolean isWindows() {
+        return File.pathSeparatorChar == ';';
+    }
 }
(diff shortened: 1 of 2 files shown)


MarkEWaite:
https://github.com/jenkinsci/gitlab-plugin/pull/1480 is the plugin pull request that reverted the parent pom upgrade.  Once this is merged, released, updated in the parent pom, and the parent pom released, then the GitLab plugin can be updated to use it.

MarkEWaite:
@timja if your time would allow it, I'd love to have this included in the next release of the Jenkins test harness.  I can't upgrade the git plugin or the gitlab plugin to the most recent parent pom until that parent pom includes a Jenkins test harness with this change.

timja (approved):
Sure I think I hit this in the scripter plugin as well

src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
```diff
@@ -133,6 +133,11 @@ private void delete(Path p) throws IOException {
             }
         }
         try {
+            if (isWindows()) {
+                // Windows throws an access denied exception when deleting read-only files
+                boolean ok = p.toFile().setWritable(true);
```
jglick, line 138: Fine though if this ever fails it will do so in an opaque fashion; using NIO APIs might be better.

PR #210, opened 2020-04-09
waitForCompletion should wait for !Run.logUpdated even after !Run.building

jglick:
I think this is the root of many of the flakes surrounding `WorkflowRun.finish`, like https://github.com/jenkinsci/workflow-job-plugin/pull/158. For both traditional job types (https://github.com/jenkinsci/jenkins/blob/449c5aced523a6e66fe3d6a804e5dbfd5c5c67c6/core/src/main/java/hudson/model/Run.java#L520-L537) and Pipeline (https://github.com/jenkinsci/workflow-job-plugin/blob/5323c277aaa1194f899e8e320bb7246f2bbcc24f/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L778-L779), `!Run.logUpdated` is stronger than `!Run.building`, so we ought to wait for the build to _really_ be completed.

PR #596, opened 2023-05-30
Use `FutureImpl.get` from `waitForCompletion`

jglick:
Since https://github.com/jenkinsci/workflow-job-plugin/pull/357 I have noticed some flakes like

```
java.lang.AssertionError: expected:<SUCCESS> but was:<FAILURE>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:120)
	at org.junit.Assert.assertEquals(Assert.java:146)
	at org.jenkinsci.plugins.workflow.job.CpsPersistenceTest.lambda$completedExecutionButRunIncomplete$23(CpsPersistenceTest.java:416)
```

The build log shows

```
[Pipeline] End of Pipeline
Finished: SUCCESS
```

yet the Jenkins log shows

```
WARNING	hudson.model.Run#onEndBuilding: testJob #1: No build result is set, so marking as failure. This should not happen.
```

I cannot reproduce this locally but I suspect the cause is that `isLogUpdated` is returning false before `WorkflowRun.finish` has completed, sometime before it calls `save()` or `onEndBuilding()`. The right way to listen for a build to complete is via its `Executor`.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -99,6 +99,7 @@ import hudson.model.UpdateSite;
 import hudson.model.User;
 import hudson.model.View;
 import hudson.model.queue.QueueTaskFuture;
+import hudson.model.queue.WorkUnit;
 import hudson.remoting.Which;
 import hudson.security.ACL;
 import hudson.security.AbstractPasswordBasedSecurityRealm;
@@ -165,6 +166,7 @@ import java.util.TreeSet;
 import java.util.UUID;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
@@ -225,6 +227,7 @@ import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 import org.junit.internal.AssumptionViolatedException;
@@ -1531,7 +1534,20 @@ public class JenkinsRule implements TestRule, MethodRule, RootAction {
      * @since 1.607
      */
     public <R extends Run<?,?>> R waitForCompletion(R r) throws InterruptedException {
-        // Could be using com.jayway.awaitility:awaitility but it seems like overkill here.
+        Executor executor = r.getExecutor();
+        if (executor != null) {
+            WorkUnit workUnit = executor.getCurrentWorkUnit();
+            if (workUnit != null) {
+                try {
+                    Queue.Executable r2 = workUnit.context.future.get();
+                    assertSame(r, r2);
+                    return r;
+                } catch (ExecutionException x) {
+                    throw new RuntimeException(x);
+                }
+            }
+        }
+        // Fallback:
         while (r.isLogUpdated()) {
             Thread.sleep(100);
         }
--- a/src/test/java/org/jvnet/hudson/test/JenkinsRuleTest.java
+++ b/src/test/java/org/jvnet/hudson/test/JenkinsRuleTest.java
@@ -5,6 +5,8 @@ import com.gargoylesoftware.htmlunit.Page;
 import com.gargoylesoftware.htmlunit.WebRequest;
 
 import hudson.EnvVars;
+import hudson.model.FreeStyleBuild;
+import hudson.model.FreeStyleProject;
 import hudson.model.RootAction;
 import hudson.model.User;
 import jenkins.model.Jenkins;
@@ -413,4 +415,13 @@ public class JenkinsRuleTest {
         j.createSlave("agent", "agent", new EnvVars());
         j.jenkins.save();
     }
+
+    @Test
+    public void waitForCompletion() throws Exception {
+        FreeStyleProject p = j.createFreeStyleProject();
+        p.getBuildersList().add(new SleepBuilder(1000));
+        FreeStyleBuild b = p.scheduleBuild2(0).getStartCondition().get();
+        j.assertBuildStatusSuccess(j.waitForCompletion(b));
+    }
+
 }


timja (approved):

jtnord:
>  The right way to listen for a build to complete is via its Executor.

Shouldn't it be via `!isInProgress() && !isBuilding()`  (as opposed to just `isInProgress()` - the same thing was [done for the ATH](https://github.com/jenkinsci/acceptance-test-harness/pull/860/files). (the former happens after all listeners are called after the build is finalized).
see also https://github.com/jenkinsci/jenkins/pull/6829

rsandell (approved):

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -1531,6 +1534,18 @@ public static String getLog(Run run) throws IOException {
      * @since 1.607
      */
     public <R extends Run<?,?>> R waitForCompletion(R r) throws InterruptedException {
+        Executor executor = r.getExecutor();
```
jtnord, line 1537: Whilst this may be technically correct - if we need to look at this then in my opinion there is an in Jenkins core, and we should not work around it in isolation.

we have `isBuilding()` and `isInProgress()` which should be combinable - if this does not work - then the whole ATH is buggy also. (isInProgress()).

Both of these do use the State - (as does isLogUpdated() unless pipeline ooverrides it) so if the log is not being updated, then the build is complete ,

The contract for the state COMPLETE is `Build is completed now, and log file is closed.`  if pipeline is overriing isLogUpdated() then this would surely sound like a break of contract - as COMPLETE is the log is closed - nothing more can add to the log ever...

jtnord (changes requested):
the more I see this change the more I realize it is incorrect.

The API from workflow-job is saying "I an complete" - yet it is not - this is a production code issue in workflow job.  the API to obtain if a build is complete is there - it is just workflow job violates this.

As this code can be called by any plugin at any time this is then not a test issue - but a production one.

jglick:
Unfortunately there is no single and consistent definition of “complete”; there are a number of stages of “completion”. Other functional test utilities such as `buildAndAssertSuccess` wait for `QueueTaskFuture.get()` to return, which is what I was following here.

> Both of these do use the State - (as does isLogUpdated() unless pipeline ooverrides it)

`Run.State` is not used by `workflow-job`.

jglick:
For example, in freestyle builds, `isLogUpdated` compares `State` to `COMPLETED`. That is set by `onEndBuilding`; the logger is actually closed _after_ this, despite the Javadoc for `RunListener.onFinalized` (called during `onEndBuilding`) claiming that at this point

> `TaskListener` is no longer available

For Pipeline, `isLogUpdated()` does indeed reflect whether anything more can be printed to the log file, but that is not precisely the condition that all tests need to be checking, since some tests actually presume that `onFinalized` has been called, or certain fields have been written to `build.xml`, or `QueueTaskFuture` has completed. What is asserted here should cover the corner cases better.

Forgot to mention that this PR amends #210, and before that https://github.com/jenkinsci/jenkins/pull/1609, and before _that_ https://github.com/jenkinsci/pipeline-plugin/pull/92 which was originally for Pipeline in fact.

jglick:
> if this does not work - then the whole ATH is buggy also

I would not expect the needs of `acceptance-test-harness` to be identical. https://github.com/jenkinsci/acceptance-test-harness/pull/860 is editing code which polls the REST API (itself a relatively slow operation) at intervals. Functional tests are dealing with much tighter time scales and so are more likely to display subtle ordering effects, particularly when some stages of a test are triggered by callbacks rather than polling (the original motivation for https://github.com/jenkinsci/workflow-job-plugin/pull/357) or involve direct access to Java structures (like the use of `java.lang.reflect.Field` in the test mentioned in the PR description), neither of which is possible in a black-box test.

jtnord:
> I would not expect the needs of acceptance-test-harness to be identical

I expect a Jenkins API that any code can call be it REST or java API that can return "this build is finished"  without the risk of race condition.  It seems list you are saying that need is different - purely due to the window of opportunity for a race condition - but I disagree - as any window of opertunity is an issue no matter how small.  

The API is "is this building" and if the answer is "no" - then the build **must** be complete - be it called via REST or or by a Java API.  As demonstrated in the need for this PR that is not the case therefore the API is buggy and broken as it is returning "I am complete" when it is not infact complete.

if you can demine that something has finished building by the lack of an executor - then why should that not be moved into the workflow job as the actual implementation?

jglick:
> if you can determine that something has finished building by the lack of an executor - then why should that not be moved into the workflow job as the actual implementation?

Not the lack of an executor per se, but the `QueueTaskFuture.get()` returning. That _is_ the final notification currently in `workflow-job`, which is what this PR is waiting for.

> the API is buggy and broken as it is returning "I am complete"

If you are referring to `isLogUpdated()` which prior to this PR was the sole condition for `waitForCompletion` to return, then no, it is correctly indicating that there is no further text coming to the log file.

> I expect a Jenkins API that any code can call be it REST or java API that can return "this build is finished" without the risk of race condition.

There is not really such an API that satisfies all use cases.

> The API is "is this building" and if the answer is "no" - then the build **must** be complete

There are `isBuilding()` and `isInProgress()` APIs, both of which can return false while `isLogUpdated()`. What exactly you would use them for, I am not quite sure, but they do not mean “entirely complete”.

Regarding the log message from the test

```
WARNING	hudson.model.Run#onEndBuilding: testJob #1: No build result is set, so marking as failure. This should not happen.
```

this is expected because the test in question deliberately mutates internal fields to try to simulate corruption of on-disk data and assert that the system recovers gracefully after a restart; this would never be expected to happen in production.

rsandell:
But historically iirc just because the job is no longer building does not mean it is complete? First RunListeners et.al needs to run, then cleaning up. And after that the run is actually completed.

dwnusbaum (approved):

amuniz (approved):

jtnord:
> > the API is buggy and broken as it is returning "I am complete"
> 
> If you are referring to `isLogUpdated()` which prior to this PR was the sole condition for `waitForCompletion` to return, then no, it is correctly indicating that there is no further text coming to the log file.

more `isInProgress()` than `isLogUpdated()`  - given the result is not final (aka written) that to me would say the job is still running,  (and building in Jenkins is not really build but running due to the overload of advjectives/nouns etc due to history where previously a job did "build").


> But historically iirc just because the job is no longer building does not mean it is complete? First RunListeners et.al needs to run, then cleaning up. And after that the run is actually completed.

And that was what `isInProgress` was actually for - (according to the documentation) - it should only be set to false (once the build starts) once there is nothing else that can effect the state of the job (be it writing to the log, changing the result or anything else).

https://github.com/jmdesprez/jenkins/blob/master/core/src/main/java/hudson/model/Run.java#L536-L543

Now complete has a definition - https://github.com/jmdesprez/jenkins/blob/master/core/src/main/java/hudson/model/Run.java#L276-L279

Jesse says above that Pipeline jobs do not use state - that is fine - to me that is an implementation details - but the contract is to me being violated - if isInProgress is returning false and there is still a QueueTaskFuture that has not completed - ergo a task for a job exists so how can it be comeplted.


I seem to be in a minority here - so have removed my -1 but still I would say this is a bug in pipeline returning `false` for `isInProgress` when it is actually still in progress...


> Regarding the log message from the test
```WARNING	hudson.model.Run#onEndBuilding: testJob #1: No build result is set, so marking as failure. This should not happen.```
this is expected because the test in question deliberately mutates internal fields to try to simulate corruption of on-disk data and assert that the system recovers gracefully after a restart; this would never be expected to happen in production.

ACK. That was not clear to me from the intial PR comment.

**Comment by jglick** [comment]
> the contract is to me being violated - if isInProgress is returning false and there is still a QueueTaskFuture that has not completed - ergo a task for a job exists so how can it not be inProgress still?

Maybe? The contract

> Determine whether the run is being build [_sic_] right now.
> true if after started and before completed

is hardly clear to me. Conversely I would be confused if a functional test failed

```java
assertFalse(jenkinsRule.buildAndAssertSuccess(someProject).isInProgress());
```

which would be possible if the order of events were flipped. There is also `Run.getExecutor`

> an executor holds on to`Run` _some more time_ even after the build is finished [emphasis added]

I do not think it is possible to satisfy everyone’s intuition without
* a much clearer state transition diagram (designed with the needs of Pipeline in mind and not retrofitted to that)
* a transactional database to ensure that transitions are atomic and there are no undocumented intermediate states

> have removed my -1 but still I would say this is a bug in pipeline

Ack. I am going to go ahead with this PR since in and of itself it seems harmless, adjusts a utility method originally written for use with Pipeline to better match the behavior of other similar-seeming methods and the actual behavior of `WorkflowRun.finish`, and _may_ resolve flakiness in some PCT runs. Whether the behavior of various `WorkflowRun.isXXX()` overrides and `QueueTaskFuture` and `RunListener.onFinalized` timings are consistent with `AbstractBuild` and/or “correct” according to existing Javadoc or intuition is a broader question; any changes in that regard have to be studied pretty closely, and there is a significant chance of PCT regressions, including hard-to-reproduce race conditions.

PR #631, opened 2023-08-02
[JENKINS-73261] Drop support for Jetty 12 EE 8 in the test harness

basil:
Removes the reflection-based compatibility layer added in #770 and drops support for EE 8. With this PR:

- Cores 2.463 and older are supported by the Java 11 backport branch at https://github.com/jenkinsci/jenkins-test-harness/tree/2225.x
- Cores 2.463 through 2.475 (inclusive) are supported by versions 2228.v919b_5883c444 through 2270.v87a_0ea_b_54da_0 (inclusive). There is no backport branch planned for these cores.
- Cores 2.476 and newer will be supported by the main branch of this repository, (as of this PR) requiring EE 9 or newer.


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jglick, line 1: Note that CloudBees CI had some RJR-based upgrade tests which relied on using `withWar` to start a controller with an older core version. These can no longer work, at least until the older version is 2.476+.

basil: @jglick Are these relevant tests now on 2.476+ (i.e., can we revert #841) or do we need to still wait a bit longer?

jglick: Thanks for the reminder. One of the tests I know about (`cloudbees-replication`) is already on 2.479.2 so that would not be affected. The other (`operations-center`) is currently on 2.440.x and should not move to 2.479.x until roughly 2025-10-30, so I would prefer to wait a while unless #841 is causing problems.

PR #635, opened 2023-08-08
remove flakyness expolding war for test

jtnord:
exploding the war for test had a race condition that caused tests to randomly fail if they had multiple tests running in parallel using JenkinsRule.

this switches a poor mans locking with race conditions to use the OS level file locking facilities

### Testing done

in this project created 20 test classes like the following :

```
package org.jvnet.hudson.test.jr;

import org.junit.Rule;
import org.junit.Test;
import org.jvnet.hudson.test.JenkinsRule;

public class JR1Test {

    @Rule
    public JenkinsRule rule = new JenkinsRule();
    
    @Test
    public void anything() {
    }

}
```

ran  `rm -fr target\jenkins-for-test && mvn test -DforkCount=20 -Dtest=JR*Test`  (on windows)

Observed that the tests did not fail and the logging is expected

```
[INFO] --- surefire:3.1.2:test (default-test) @ jenkins-test-harness ---
[INFO] Using auto detected provider org.apache.maven.surefire.junitplatform.JUnitPlatformProvider
[INFO]
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.jvnet.hudson.test.jr.JR12Test
[INFO] Running org.jvnet.hudson.test.jr.JR19Test
[INFO] Running org.jvnet.hudson.test.jr.JR20Test
[INFO] Running org.jvnet.hudson.test.jr.JR2Test
[INFO] Running org.jvnet.hudson.test.jr.JR7Test
[INFO] Running org.jvnet.hudson.test.jr.JR15Test
[INFO] Running org.jvnet.hudson.test.jr.JR11Test
[INFO] Running org.jvnet.hudson.test.jr.JR10Test
[INFO] Running org.jvnet.hudson.test.jr.JR4Test
[INFO] Running org.jvnet.hudson.test.jr.JR14Test
[INFO] Running org.jvnet.hudson.test.jr.JR17Test
[INFO] Running org.jvnet.hudson.test.jr.JR8Test
[INFO] Running org.jvnet.hudson.test.jr.JR18Test
[INFO] Running org.jvnet.hudson.test.jr.JR6Test
[INFO] Running org.jvnet.hudson.test.jr.JR16Test
[INFO] Running org.jvnet.hudson.test.jr.JR5Test
[INFO] Running org.jvnet.hudson.test.jr.JR13Test
[INFO] Running org.jvnet.hudson.test.jr.JR9Test
[...]

PR #638, opened 2023-08-14
Remove unnecessary defaults

basil:
Amends #635

- The default value of `forkCount` is 1 so not necessary to define it to 1 here
- The default property for `<forkCount>` is `${forkCount}` so not necessary to define it to `${forkCount}` here


jtnord (commented):
👍

PR #654, opened 2023-09-15
Work around an apparent JDK file deletion bug on Windows

jglick:
https://github.com/jenkinsci/workflow-cps-global-lib-http-plugin/pull/167 suppresses a couple of tests on Windows assuming they are flakes introduced by #198. In fact the failures are caused by #198 but are not flakes; I am able to reproduce https://github.com/jenkinsci/workflow-cps-global-lib-http-plugin/runs/16806834502 consistently on Windows 10 running on a VirtualBox file share, and I see no evidence of unclosed file handles or background activity in that plugin. For some reason, the `libs` subdirectory of the build directory does not get deleted the first time you call `Files.delete` on it (after deleting its children). Asking to delete `libs` again works. No other directories are affected. Using `File.delete` rather than the NIO method also deletes `libs` the first time. I suspect a bug in `WindowsFileSystemProvider.implDelete` but have not yet tried to narrow it down to a reproducible test case.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
+++ b/src/main/java/org/jvnet/hudson/test/TemporaryDirectoryAllocator.java
@@ -144,6 +144,9 @@ public class TemporaryDirectoryAllocator {
                 throw new IOException(children.map(Path::toString).collect(Collectors.joining(" ")), x);
             }
         }
+        if (isWindows() && Files.deleteIfExists(p)) {
+            LOGGER.warning(() -> "had to try twice to delete " + p);
+        }
     }
 
     private boolean isWindows() {


timja (approved):

jglick:
Holding off on this in the hope that I will find time to dig into why just these two tests, and no others that I know of, would cause the odd behavior.

jtnord:
couple of things.
the /lib directory sounds like it is a git clone?  
if so - perhaps a bug in jgit that is not releasing a file correctly (or perhaps caused by a bug in the code somewhere that is not closing something expected, and then leaving it to GC).
That it works in the old `io` sounds more like it is a timing issue rather than a bug in the JDK, alternatively something in pipeline still around.

https://github.com/basil/workflow-cps-global-lib-http-plugin/blob/bbab97bf7e1352cf6d0279bc788489340b1f2dbf/src/test/java/com/amadeus/jenkins/plugins/workflow/libs/HttpRetrieverIntegrationTests.java#L240-L242 and https://github.com/basil/workflow-cps-global-lib-http-plugin/blob/bbab97bf7e1352cf6d0279bc788489340b1f2dbf/src/test/java/com/amadeus/jenkins/plugins/workflow/libs/HttpRetrieverIntegrationTests.java#L274-L276 looks to be the culript or at least suspicious on the surface wrong - it is creating a stream using `Files.list` but it is not closing it, instead relying on GC.

Terminal operators do not close the underlying stream.

jtnord (changes requested):
There is no bug to fix - the plugins test code was bad.

jglick:
> the /lib directory sounds like it is a git clone?

In this case, no, the clone happens to a shared location and then selected files are copied into the build directory with a `FilePath` method.

PR #657, opened 2023-09-21
Set up file logging when shutting down the Jenkins test instance.

Vlatombe:
Matching change to https://github.com/jenkinsci/jenkins/pull/8513

This should help to identify race conditions during tests without causing too much noise as it only kicks in just before cleaning up the instance.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -130,6 +130,7 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.AbstractMap;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Base64;
 import java.util.Collections;
 import java.util.Enumeration;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -563,7 +563,7 @@ public static void _stopJenkins(Server server, List<LenientRunnable> tearDowns,
                 }
             }
         }
-
+        tweakXmlFileLogger();
```
jtnord, line 566: Loggers are generally static (as it the case in XmlFile) and as we are using the same JVM for a class rather than each test - does this not pollute the logger for the next test run in the same class, potentially causing a lot of log spam?

Vlatombe: You're right, should be better now

jtnord (approved):

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -563,18 +564,46 @@ public static void _stopJenkins(Server server, List<LenientRunnable> tearDowns,
...
         }
-        ExtensionList.clearLegacyInstances();
-        DescriptorExtensionList.clearLegacyInstances();
-
         if (exception.getSuppressed().length > 0) {
             throw exception;
         }
     }
 
+    private static final class TemporaryConsoleLogTweak implements AutoCloseable {
```
basil, line 579: Why is this temporary, and how will we ensure that it does not become permanent?

Vlatombe: This is not meant to be temporary, I renamed the class to clarify the intent.

basil: > This is not meant to be temporary

What would its value be after we diagnose and fix the bug seen in https://github.com/jenkinsci/jenkins/pull/8511/checks?check_run_id=17000216956?

jglick (approved):
Looks helpful; has it be shown to work anywhere?

Vlatombe:
> Looks helpful; has it be shown to work anywhere?

I tested this change locally against the test failing in https://github.com/jenkinsci/jenkins/pull/8511/checks?check_run_id=17000216956, it shows correctly the `queue.xml` being written just before cleanup, so I think it would be enough to show the race condition if it rehappens.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
Vlatombe: If there is no more race condition, I would fine removing it, but the thing is with this kind of issue is that it always pop up at an unexpected time and you often don't have enough data to investigate. On the other hand its impact is very light (under normal condition this adds one stacktrace writing `queue.xml`) since you wouldn't expect much xmlfile write being donc between stopping the test Jenkins instance and the cleanup sequence.

jglick:
Ah so this is going to print

```
FINEST	hudson.XmlFile#write: Writing …/queue.xml
java.lang.Throwable
	at hudson.XmlFile.write(XmlFile.java:205)
	at hudson.model.Queue.save(Queue.java:472)
	at jenkins.model.Jenkins._cleanUpPersistQueue(Jenkins.java:3921)
	at jenkins.model.Jenkins.cleanUp(Jenkins.java:3676)
	at org.jvnet.hudson.test.JenkinsRule._stopJenkins(JenkinsRule.java:569)
	at org.jvnet.hudson.test.JenkinsRule.after(JenkinsRule.java:517)
	at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:665)
```

at the end of _every_ test, which is not really desirable, since it is adding noise about a normal event. We would certainly want to know about writes after `cleanUp`. What this addition does _not_ show is `config.xml` being written very late, which might be the issue according to the stack trace of the flake.

jglick (approved):
Might work. Hard to predict.

Vlatombe:
> Might work. Hard to predict.

Yeah, at least this is not polluting every test log anymore.

jglick:
I reran all of `QuietDownCommandTest` with this patch and did not see any `XmlFile` stack traces, so I think this is reasonable to merge as is, and we can hope it reveals something.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
basil: Fine with the merge of this PR if and only if it is reverted once its diagnostic value has been exhausted.

basil:
https://github.com/jenkinsci/jenkins-test-harness/pull/657#discussion_r1340455592

PR #659, opened 2023-09-27
`RealJenkinsRule.addSyntheticPlugin`

### LINKED ISSUE #623: Test-only plugin from source in `RealJenkinsRule`
`JenkinsRule` tests commonly use `@TestExtension` to define mock implementations of various APIs, which is a very powerful system. For `RealJenkinsRule`, there is no direct equivalent. As https://github.com/jenkinsci/jenkins-test-harness/blob/787a641a9b265eab61fbcf18c0532a0b26d76662/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java#L140 hints at, you _can_ define new types in test sources and load them in a Jenkins JVM, and even register them programmatically with `ExtensionList.add` (suppressing the deprecation warning), but they will be defined in an unusual class loader (not `UberClassLoader`) and there may be other reasons why the test cannot be made to pass or would not be convincing if it did.

:thinking: There could be a system whereby you could define a `src/test/java/some/dedicated/pkg/package-info.java` with a `SOURCE`-retention annotation directing a processor to create a `target/test-classes/plugins/some-dedicated-pkg.jpi` which could be included in `addPlugins`. https://github.com/jenkinsci/jenkins-test-harness/blob/787a641a9b265eab61fbcf18c0532a0b26d76662/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java#L222-L227 notes that when you do this today, you must commit the plugin binary to Git, making the test harder to review and maintain.

jglick:
Fixes #623. ~Unfortunately I have hit a roadblock: even when `RoundEnvironment.processingOver`, `*.class` files for sources have not yet been written to disk, so it is impossible to pack them into the plugin archive.~


jglick:
After sleeping on it, I think it is unnecessary to pack the plugin archive at build time. It would suffice for `RealJenkinsRule` to create a test plugin on demand.

Vlatombe:
Looks good, why is it still in draft?

jglick:
> why is it still in draft?

Because I never actually had a use case for it—the proprietary change for which I had planned to use it turned out to be testable by just injecting an extension loaded from the test class loader. (In that case `Descriptor` was not needed, nor did the test extension need to be deserialized, etc.) Do you have a use case for this? If so, I am happy to take it out of draft.

jglick:
I now have a definite use case for this utility in a proprietary plugin: a case where I need for the test to run a mock `@Extension` which must be active before Jenkins completes initialization (and thus before I can `runRemotely` something to install it).

timja (approved):

src/test/java/org/jvnet/hudson/test/sample/plugin/Stuff.java
```diff
@@ -0,0 +1,50 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2023 CloudBees, Inc.
```
Vlatombe, line 4:
```suggestion
 * Copyright 2024 CloudBees, Inc.
```

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -1617,4 +1652,121 @@ private static class OutputPayload implements Serializable {
...
+     */
+    public final class SyntheticPlugin {
+        private final String pkg;
+        private String shortName;
+        private String version = "1-SNAPSHOT";
+        private Map<String, String> headers = new HashMap<>();
+
+        SyntheticPlugin(String pkg) {
+            this.pkg = pkg;
+            shortName = pkg.replace('.', '-');
```
Vlatombe, line 1668: Can be distinguished from usual plugins because the convention is different, but in my mind it would be clearer if we used a clear prefix to identify synthetic plugins (e.g. `test-<shortName>` or `synthetic-<shortName>`)

Vlatombe (approved):

PR #671, opened 2023-10-26
`RealJenkinsRule.runRemotely` overloads taking 1–2 step args

jglick:
In @cloudbees we have a lot of `RealJenkinsRule`-based tests. We often prefer to use method handles for assertions, setup utilities, etc.; this worked for steps which returned values (`Step2`, as of #450) or did not (`Step`) but until now passing a parameter required implementing a `static` class, which is more verbose.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
+++ b/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
@@ -95,7 +95,7 @@ public final class InboundAgentRule extends ExternalResource {
         private boolean webSocket;
         @CheckForNull private String tunnel;
         private boolean start = true;
-        private final Map<String, Level> loggers = new LinkedHashMap<>();
+        private final LinkedHashMap<String, Level> loggers = new LinkedHashMap<>();
         private String label;
         private final PrefixedOutputStream.Builder prefixedOutputStreamBuilder = PrefixedOutputStream.builder();
 
@@ -267,7 +267,7 @@ public final class InboundAgentRule extends ExternalResource {
      * @param options the options
      */
     public Slave createAgent(@NonNull JenkinsRule r, Options options) throws Exception {
-        Slave s = CreateAgent.createAgent(r, options);
+        Slave s = createAgentJR(r, options);
         if (options.isStart()) {
             start(r, options);
         }
@@ -279,7 +279,7 @@ public final class InboundAgentRule extends ExternalResource {
     }
 
     public void createAgent(@NonNull RealJenkinsRule rr, Options options) throws Throwable {
-        String name = rr.runRemotely(new CreateAgent(options));
+        String name = rr.runRemotely(InboundAgentRule::createAgentRJR, options);
         options.name = name;
         if (options.isStart()) {
             start(rr, options);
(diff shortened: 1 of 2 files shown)


src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
```diff
@@ -95,7 +95,7 @@ public static final class Options implements Serializable {
         private boolean webSocket;
         @CheckForNull private String tunnel;
         private boolean start = true;
-        private final Map<String, Level> loggers = new LinkedHashMap<>();
+        private final LinkedHashMap<String, Level> loggers = new LinkedHashMap<>();
```
jglick, line 98: An interesting case for language nerds: we need to prove to the compiler that the map is `Serializable` in https://github.com/jenkinsci/jenkins-test-harness/blob/7a4133525c7591d025da037fd2ed3c19a34c9e8c/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java#L315 which works even though the method definition https://github.com/jenkinsci/jenkins-test-harness/blob/7a4133525c7591d025da037fd2ed3c19a34c9e8c/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java#L471 declares a non-`Serializable` argument type.

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
```diff
@@ -447,124 +447,74 @@ public AgentArguments(@NonNull String agentJnlpUrl, @NonNull File agentJar, @Non
...
     }
 
-    private static class GetAgentArguments implements RealJenkinsRule.Step2<AgentArguments> {
-
-        private final String name;
-
-        GetAgentArguments(String name) {
-            this.name = name;
+    private static AgentArguments getAgentArguments(JenkinsRule r, String name) throws IOException {
+        Node node = r.jenkins.getNode(name);
```
jglick, line 451: (ignore WS to review)

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
```diff
@@ -447,124 +447,74 @@ public AgentArguments(@NonNull String agentJnlpUrl, @NonNull File agentJar, @Non
...
         }
     }
 
-    private static class CreateAgent implements RealJenkinsRule.Step2<String> {
-        private final Options options;
-
-        CreateAgent(Options options) {
-            this.options = options;
-        }
+    private static String createAgentRJR(JenkinsRule r, Options options) throws Throwable {
```
jglick, line 496: not to conflict with https://github.com/jenkinsci/jenkins-test-harness/blob/7a4133525c7591d025da037fd2ed3c19a34c9e8c/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java#L269

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
```diff
@@ -447,124 +447,74 @@ public AgentArguments(@NonNull String agentJnlpUrl, @NonNull File agentJar, @Non
...
+        createAgentJR(r, options);
+        return options.getName();
```
jglick, line 498: still need two methods, because in RJR mode we can only return the agent name

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -564,6 +564,8 @@ public static List<String> getJacocoAgentOptions() {
      *     // as needed
      * }
      * </pre>
+     * If you need to pass and/or return values, you can still use a static method reference:
+     * try {@link #runRemotely(Step2)} or {@link #runRemotely(StepWithReturnAndOneArg, Serializable)} etc.
```
jglick, line 568: Could extend this to 3+ arguments if there were any use cases, of course.

amuniz (approved):

PR #687, opened 2023-11-28
Provide the means to customize the folder containing the java executable for RJR

raul-arabaolaza:
Can be used to launch the JuT using any JVM instead of always delegating to JAVA_HOME.

Useful due to the upcoming support process for [different JVM versions](https://www.jenkins.io/blog/2023/11/06/introducing-2-2-2-java-support-plan/) as it would allow to create RJR test configurations that share the same code but run with different JVMs, like 17 and 21

### Testing done
Manual tested by creating a test that uses a custom location on my local system.
There are unit tests that ensures the new code works as expected also.

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -167,6 +167,8 @@ public final class RealJenkinsRule implements TestRule {
 
     private File war;
 
+    private String altJavaHome;
+
     private boolean includeTestClasspathPlugins = true;
 
     private final String token = UUID.randomUUID().toString();
@@ -306,6 +308,14 @@ public final class RealJenkinsRule implements TestRule {
      */
     public RealJenkinsRule withWar(File war) {
         this.war = war;
+        return this;
+    }
+
+    /**
+     * Allows to specify an alternate, not the one specified in JAVA_HOME, folder containing the JVM to use to launch the instance
+     */
+    public RealJenkinsRule withAltJavaHome(String altJavaHome) {
+        this.altJavaHome = altJavaHome;
         return this;
     }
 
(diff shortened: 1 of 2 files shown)


raul-arabaolaza:
Coherent test issue, I will fix it

car-roll (approved):

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -309,6 +311,14 @@ public RealJenkinsRule withWar(File war) {
         return this;
     }
 
+    /**
+     * Allows to specify an alternate, not the one specified in JAVA_HOME, folder containing the JVM to use to launch the instance
+     */
+    public RealJenkinsRule withAltJavaHome(String altJavaHome) {
```
timja, line 317: is alt needed here? `withJavaHome` seems fine to me, and then the docs could be simplified I think.
(non blocking)

```suggestion
    public RealJenkinsRule withJavaHome(String altJavaHome) {
```

timja (approved):

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
raul-arabaolaza: For me is an alternative one given the code assumes there is a default java home to use... However I understand is a matter so if there is other person (so 2 to 1) who wants to change I will

raul-arabaolaza:
Can I get this merged, please? It seems no more feedback is coming and this is already approved

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
Vlatombe: `withJavaHome` seems more natural to me as well, with a mention that it defaults to `java.home` sysprop.

raul-arabaolaza: @Vlatombe @timja Refactored in #693

PR #690, opened 2023-11-30
Allow `RealJenkinsRule` to delay `JENKINS_HOME` initialization until first startup using new `prepareHomeLazily` method

### LINKED ISSUE #365: `RealJenkinsRule.addPlugins` / `.omitPlugins` take effect only if called before test method starts
See #346.

dwnusbaum:
Fixes #365.

In addition to what is described in #365, I have a case where I would like to use `RealJenkinsRule` dynamically in tests in a way that is not well suited to `@Rule`/`@ClassRule`. This PR adds a `deprovision` method, which deletes `JENKINS_HOME`, and an `isAlive` method, which checks whether the Jenkins process is alive. In combination with the new `withLazyProvisioning` method that delays `JENKINS_HOME` initialization until Jenkins first starts, this allows `RealJenkinsRule` to be instantiated, provisioned, started, stopped, and deprovisioned fully dynamically. The `deprovision` and `isAlive` methods can be ignored in most cases.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -197,7 +197,7 @@ public final class RealJenkinsRule implements TestRule {
     private boolean debugServer = true;
     private boolean debugSuspend;
 
-    private boolean lazyProvisioning;
+    private boolean prepareHomeLazily;
     private boolean provisioned;
 
     // TODO may need to be relaxed for Gradle-based plugins
@@ -445,14 +445,14 @@ public final class RealJenkinsRule implements TestRule {
     }
 
     /**
-     * Allows JENKINS_HOME initialization to be delayed until {@link #startJenkins} is called for the first time.
+     * Allows {@code JENKINS_HOME} initialization to be delayed until {@link #startJenkins} is called for the first time.
      * <p>
      * This allows methods such as {@link #addPlugins} to be called dynamically inside of test methods, which enables
      * related tests that need to configure {@link RealJenkinsRule} in different ways to be defined in the same class
      * using only a single instance of {@link RealJenkinsRule}.
      */
-    public RealJenkinsRule withLazyProvisioning(boolean lazyProvisioning) {
-        this.lazyProvisioning = lazyProvisioning;
+    public RealJenkinsRule prepareHomeLazily(boolean prepareHomeLazily) {
+        this.prepareHomeLazily = prepareHomeLazily;
         return this;
     }
 
@@ -482,7 +482,7 @@ public final class RealJenkinsRule implements TestRule {
                 }
                 try {
                     home.set(tmp.allocate());
-                    if (!lazyProvisioning) {
+                    if (!prepareHomeLazily) {
                         provision();
                     }
                     base.evaluate();
@@ -709,7 +709,7 @@ public final class RealJenkinsRule implements TestRule {
         if (proc != null) {
             throw new IllegalStateException("Jenkins is (supposedly) already running");
         }
-        if (lazyProvisioning && !provisioned) {
+        if (prepareHomeLazily && !provisioned) {
             provision();
         }
         String cp = System.getProperty("java.class.path");
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -197,6 +197,8 @@ public final class RealJenkinsRule implements TestRule {
     private boolean debugServer = true;
     private boolean debugSuspend;
 
+    private boolean lazyProvisioning;
+
     // TODO may need to be relaxed for Gradle-based plugins
     private static final Pattern SNAPSHOT_INDEX_JELLY = Pattern.compile("(file:/.+/target)/classes/index.jelly");
 
@@ -441,6 +443,18 @@ public final class RealJenkinsRule implements TestRule {
         return this;
     }
 
+    /**
+     * Allows JENKINS_HOME initialization to be delayed until {@link startJenkins} is called for the first time.
+     * <p>
+     * This allows methods such as {@link addPlugins} to be called dynamically inside of test methods, which enables
+     * related tests that need to configure {@link RealJenkinsRule} in different ways to be defined in the same class
+     * using only a single instance of {@link RealJenkinsRule}.
+     */
+    public RealJenkinsRule withLazyProvisioning(boolean lazyProvisioning) {
+        this.lazyProvisioning = lazyProvisioning;
+        return this;
+    }
+
     public static List<String> getJacocoAgentOptions() {
         RuntimeMXBean runtimeMxBean = ManagementFactory.getRuntimeMXBean();
         List<String> arguments = runtimeMxBean.getInputArguments();
@@ -466,7 +480,9 @@ public final class RealJenkinsRule implements TestRule {
                     return;
                 }
                 try {
-                    provision(description);
+                    if (!lazyProvisioning) {
+                        provision();
+                    }
                     base.evaluate();
                 } finally {
                     stopJenkins();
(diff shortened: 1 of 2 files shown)


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -572,6 +481,131 @@ public static List<String> getJacocoAgentOptions() {
...
+     */
+    @SuppressFBWarnings(value = "PATH_TRAVERSAL_IN", justification = "irrelevant")
+    public void provision(Description description) throws Exception {
+        this.description = description;
+        if (home.get() != null) {
+            throw new IllegalStateException(description + " was already provisioned");
+        }
+        if (war == null) {
+            war = findJenkinsWar();
+        }
```
dwnusbaum, line 497: This is mostly a copy-paste, but these lines are new.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -572,6 +481,131 @@ public static List<String> getJacocoAgentOptions() {
...
+        home.set(tmp.allocate());
+        LocalData localData = description.getAnnotation(LocalData.class);
+        if (localData != null) {
+            new HudsonHomeLoader.Local(description.getTestClass().getMethod(description.getMethodName()), localData.value()).copy(getHome());
+        }
+        File plugins = new File(getHome(), "plugins");
+        Files.createDirectories(plugins.toPath());
```
dwnusbaum, line 504: This line was changed to use NIO to get rid of a SpotBugs warning. Not sure why it wasn't erroring before.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -572,6 +481,131 @@ public static List<String> getJacocoAgentOptions() {
...
+    }
+
+    /**
+     * Deletes {@code JENKINS_HOME}.
+     *
+     * This method does not need to be invoked when using {@code @Rule} or {@code @ClassRule} to run {@code RealJenkinsRule}.
+     */
+    public void deprovision() throws Exception {
+        tmp.dispose();
+        home.set(null);
```
dwnusbaum, line 599: This line is new, and I think it only matters for dynamically-provisioned instances. In normal cases this happens after the test method completes.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -441,6 +443,18 @@ public RealJenkinsRule includeTestClasspathPlugins(boolean includeTestClasspathP
         return this;
     }
 
+    /**
+     * Allows JENKINS_HOME initialization to be delayed until {@link startJenkins} is called for the first time.
+     * <p>
+     * This allows methods such as {@link addPlugins} to be called dynamically inside of test methods, which enables
```
jglick, line 449:
```suggestion
     * Allows JENKINS_HOME initialization to be delayed until {@link #startJenkins} is called for the first time.
     * <p>
     * This allows methods such as {@link #addPlugins} to be called dynamically inside of test methods, which enables
```

dwnusbaum:
I think what we want is for the preparation of $JENKINS_HOME/plugins/ to be done lazily right before Jenkins is first started. (Not completely sure this is backward compatible—might need to be opt-in.)

Ok, I can adapt my use case to support this kind of approach instead of an explicit `provision` method. See https://github.com/jenkinsci/jenkins-test-harness/pull/690/commits/11b1199e527fb6ad5a481e978fb6df8db09b75e8.

I'm working on adapting the tests but there is some bug with `agentBuild` (and maybe other tests) hanging, which I think is related to `TailLog` but I'm not sure yet.

jglick:
`Thread.setDaemon` maybe?

dwnusbaum:
The problem was that `TailLog` needs `RealJenkinsRule.getHome`, but lazy provisioning didn't set `home` until after Jenkins started, so `TailLog` was looking for `path/to/whatever` in the CWD, which didn't exist. It seems fine to pre-allocate `home` to avoid that problem. Alternatively, when using lazy provisioning with `TailLog` you could use `startJenkins` and then `TailLog` and then `runRemotely`, but that seems awkward and easy to mess up.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -446,7 +447,7 @@ public RealJenkinsRule includeTestClasspathPlugins(boolean includeTestClasspathP
     /**
      * Allows JENKINS_HOME initialization to be delayed until {@link startJenkins} is called for the first time.
```
jglick, line 448:
```suggestion
     * Allows JENKINS_HOME initialization to be delayed until {@link #startJenkins} is called for the first time.
```

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
```diff
@@ -72,11 +72,10 @@
 
 public class RealJenkinsRuleTest {
 
-    // TODO addPlugins does not currently take effect when used inside test method
-    @Rule public RealJenkinsRule rr = new RealJenkinsRule().addPlugins("plugins/structs.hpi").withDebugPort(4001).withDebugServer(false);
-    @Rule public RealJenkinsRule rrWithFailure = new RealJenkinsRule().addPlugins("plugins/failure.hpi");
+    @Rule public RealJenkinsRule rr = new RealJenkinsRule().withLazyProvisioning(true).withDebugPort(4001).withDebugServer(false);
 
     @Test public void smokes() throws Throwable {
+        rr.addPlugins("plugins/structs.hpi");
```
jglick, line 78: Much better.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -441,6 +444,18 @@ public RealJenkinsRule includeTestClasspathPlugins(boolean includeTestClasspathP
...
     }
 
+    /**
+     * Allows JENKINS_HOME initialization to be delayed until {@link #startJenkins} is called for the first time.
+     * <p>
+     * This allows methods such as {@link #addPlugins} to be called dynamically inside of test methods, which enables
+     * related tests that need to configure {@link RealJenkinsRule} in different ways to be defined in the same class
+     * using only a single instance of {@link RealJenkinsRule}.
+     */
+    public RealJenkinsRule withLazyProvisioning(boolean lazyProvisioning) {
```
dwnusbaum, line 454: I don't particularly care about this name and it is somewhat verbose. Perhaps just `lazy` would be good enough.

jglick: `prepareHomeLazily`? No strong opinion.

dwnusbaum: https://github.com/jenkinsci/jenkins-test-harness/pull/690/commits/1dcdc9ef15a957cccaab772943e18d7f641e69b1

dwnusbaum:
Ok, I have tested this in a project that uses `RealJenkinsRule` extensively, which helped me find a mistake here (https://github.com/jenkinsci/jenkins-test-harness/pull/690/commits/46729d968a260ff936938a7fbbbe80c8187d9523), but now that that has been fixed everything works well, so I think this is ready for review.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jglick: Is that the only change being made as this code is moved out of `apply`? The way the diff is structured it is very hard to verify, unfortunately. (Sometimes it is possible to arrange new methods in such a way that `diff -w` is tricked into showing only small hunks, but I cannot see that working here.)

jglick: Ah never mind! It does work here if you just hide whitespace: https://github.com/jenkinsci/jenkins-test-harness/pull/690/files?diff=unified&w=1

jglick (approved):

PR #723, opened 2024-01-26
Find the right Submit button

### LINKED ISSUE #703: `HtmlFormUtil.getSubmitButton` picks wrong button for Jenkins config forms
Discovered in #702.

```diff
diff --git pom.xml pom.xml
index d4b87033d5..54dd8292c1 100644
--- pom.xml
+++ pom.xml
@@ -56,7 +56,7 @@ THE SOFTWARE.
 
   <properties>
     <changelist>999999-SNAPSHOT</changelist>
-    <jenkins.version>2.361</jenkins.version>
+    <jenkins.version>2.437</jenkins.version>
     <jmh.version>1.37</jmh.version>
     <gitHubRepo>jenkinsci/${project.artifactId}</gitHubRepo>
     <!-- Normally filled in by "maven-hpi-plugin" with the path to "org-netbeans-insane-hook.jar" extracted from this repository -->
diff --git src/main/java/org/htmlunit/html/HtmlFormUtil.java src/main/java/org/htmlunit/html/HtmlFormUtil.java
index c5bcccd75c..e696c9db67 100644
--- src/main/java/org/htmlunit/html/HtmlFormUtil.java
+++ src/main/java/org/htmlunit/html/HtmlFormUtil.java
@@ -50,6 +50,7 @@ public class HtmlFormUtil {
      */
     public static Page submit(final HtmlForm htmlForm) throws IOException {
         HtmlElement submitElement = getSubmitButton(htmlForm);
+        System.err.println("TODO submitElement=" + submitElement);
         return submit(htmlForm, submitElement);
     }
 
diff --git src/test/java/org/jvnet/hudson/test/SleepBuilderTest.java src/test/java/org/jvnet/hudson/test/SleepBuilderTest.java
index 1ae1b411a8..bab6aa6fef 100644
--- src/test/java/org/jvnet/hudson/test/SleepBuilderTest.java
+++ src/test/java/org/jvnet/hudson/test/SleepBuilderTest.java
@@ -14,8 +14,7 @@ public class SleepBuilderTest {
         FreeStyleProject project = j.createFreeStyleProject();
         SleepBuilder builder = new SleepBuilder(30);
         project.getBuildersList().add(builder);
-        j.configRoundtrip(project);
-        j.assertEqualDataBoundBeans(project.getBuildersList().get(SleepBuilder.class), builder);
+        j.configRoundtrip();
     }
 
 }
```

You might expect this to print the **Save** button. But that is not an `<input type="submit">`, it is some YUI craziness. Instead it picks

```
HtmlButton[<button tooltip="Copy home directory" text="/tmp/j h1749655323416403161" message="Copied" type="button" class="copy-button jenkins-button jenkins-button--tertiary jenkins-copy-button jenkins-!-margin-left-1" title="Copy home directory">]
```

And of course you do not want to click _that_. Yet it seems that `HtmlForm.submit` quietly accepts this and submits the form anyway.

amuniz:
Fixes #703

Try to find the right Submit button instead of returning whatever `<button>` comes first in the form DOM.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/htmlunit/html/HtmlFormUtil.java
+++ b/src/main/java/org/htmlunit/html/HtmlFormUtil.java
@@ -98,12 +98,17 @@ public class HtmlFormUtil {
     }
 
     /**
-     * Gets the first {@code <input type="submit">} element in this form.
+     * Gets the first {@code <input type="submit">} or {@code <button name="Submit">} element in this form.
      */
     public static HtmlElement getSubmitButton(final HtmlForm htmlForm) throws ElementNotFoundException {
         List<HtmlElement> submitButtons = getSubmitButtons(htmlForm);
         if (!submitButtons.isEmpty()) {
             return submitButtons.get(0);
+        }
+        for (HtmlElement element : htmlForm.getElementsByAttribute("button", "name", "Submit")) {
+            if(element instanceof HtmlButton) {
+                return element;
+            }
         }
         for (HtmlElement element : htmlForm.getElementsByTagName("button")) {
             if(element instanceof HtmlButton) {


src/main/java/org/htmlunit/html/HtmlFormUtil.java
```diff
@@ -98,13 +98,18 @@ public static List<HtmlElement> getSubmitButtons(final HtmlForm htmlForm) throws
...
     public static HtmlElement getSubmitButton(final HtmlForm htmlForm) throws ElementNotFoundException {
         List<HtmlElement> submitButtons = getSubmitButtons(htmlForm);
         if (!submitButtons.isEmpty()) {
             return submitButtons.get(0);
         }
+        for (HtmlElement element : htmlForm.getElementsByAttribute("button", "name", "Submit")) {
+            if(element instanceof HtmlButton) {
+                return element;
+            }
+        }
```
amuniz, line 112: Without this, any other button that happens to show first in the page is returned, which is likely not going to be the Submit (Save) button for the form.

src/main/java/org/htmlunit/html/HtmlFormUtil.java
```diff
@@ -98,13 +98,18 @@ public static List<HtmlElement> getSubmitButtons(final HtmlForm htmlForm) throws
...
-     * Gets the first {@code <input type="submit">} element in this form.
+     * Gets the first {@code <input type="submit">} or {@code <button name="Submit">} element in this form.
      */
     public static HtmlElement getSubmitButton(final HtmlForm htmlForm) throws ElementNotFoundException {
         List<HtmlElement> submitButtons = getSubmitButtons(htmlForm);
         if (!submitButtons.isEmpty()) {
             return submitButtons.get(0);
         }
+        for (HtmlElement element : htmlForm.getElementsByAttribute("button", "name", "Submit")) {
+            if(element instanceof HtmlButton) {
```
jglick, line 109:
```suggestion
            if (element instanceof HtmlButton) {
```

Yeah I know https://github.com/jenkinsci/jenkins-test-harness/blob/07a24aeb0e8ba913a776d34700f721e357581590/src/main/java/org/htmlunit/html/HtmlFormUtil.java#L114 maybe we should enable Spotless.

jglick (approved):
_Looks_ right, but it is hard to predict what weird effects there might be in the universe of tests. Can this be checked against, say, Jenkins core which I am guessing exercises this code multiple times?

src/main/java/org/htmlunit/html/HtmlFormUtil.java
jtnord: I think this works but may still be brittle, as the submit button could be named something else.
(e.g.  Deploy plugin in plugin manager).  -> https://github.com/jenkinsci/jenkins/blob/master/test/src/test/java/hudson/PluginManagerTest.java#L152-L159 

So I think this may be better looking to find this submit first, then a button that is [primary](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/lib/form/submit.jelly#L44-L49), and then fall back to the first button in the form?

Vlatombe: Looks fragile, as the submit button name can be customized.
 
As a workaround, ignore button with class `jenkins-copy-button` and we would need to add a new class e.g. `jenkins-submit-button` to https://github.com/jenkinsci/jenkins/blob/58a8f2e018f733cf4ff9155f2c4806e079cbd31d/core/src/main/resources/lib/form/submit.jelly#L57

jtnord (commented):
I think this is work for some cases, but not possibly not enough cases.

if the submit button was called something other than submit it can still pick the first button - even if that button is not primary (e.g. a form with `Apply` and `Save`) buttons would pick apply and not save (when Save is the normal primary button)

src/main/java/org/htmlunit/html/HtmlFormUtil.java
amuniz: > As a workaround, ignore button with class jenkins-copy-button

This would solve the specific case of `JenkinsRule#configRoundtrip` (the "copy" button in the one under `/configure` which copies the `JENKINS_HOME` path to the clipboard) but not other potential cases where there other buttons. So, as a workaround (until the core patch), I think @jtnord proposal might be the best (or less bad) one?

amuniz: Also, the _default_ value is actually "Submit", so I think this would be dodging most of the possible hits?

amuniz:
Just to be clear, this fix is not intended to be a complete fix, but a workaround until we get the special marker class added in the submit button in core.

src/main/java/org/htmlunit/html/HtmlFormUtil.java
jtnord: you are correct, I was confusing the text and the name.

jtnord (approved):

jglick:
> Can this be checked against, say, Jenkins core

Tip: create a draft PR against core setting the JTH version to the incremental version from here.

amuniz:
Core test PR: https://github.com/jenkinsci/jenkins/pull/8911

amuniz:
I couldn't find a plugin using `configRountrip` _a lot_, so I'm just picking one which uses it a couple of times and it's not in the `/configure` screen but at job level: https://github.com/jenkinsci/workflow-cps-plugin/pull/841

amuniz:
Test PRs passed. I'll merge this today if nobody objects. 
I'll also create a PR to core to add the special marker class and a separated PR to `jenkins-test-harness` to use that class (keeping backward compatibility so it can be merged without waiting for the core change).

PR #728, opened 2024-02-08
Print text lines atomically from `TailLog`

jglick:
I noticed that when starting multiple builds nearly simultaneously and running several copies of `TailLog`, often lines of output would get intermixed and mangled. After adding synchronization I have not seen this recur.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/TailLog.java
+++ b/src/main/java/org/jvnet/hudson/test/TailLog.java
@@ -99,12 +99,15 @@ public final class TailLog implements AutoCloseable {
             @Override
             public void handle(String line) {
                 if (ps == null) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/TailLog.java
```diff
@@ -99,12 +99,15 @@ public TailLog(File buildDirectory, String job, int number) {
             @Override
             public void handle(String line) {
                 if (ps == null) {
-                    ps = new PrintStream(new PlainTextConsoleOutputStream(prefixedOutputStreamBuilder.withName(job + '#' + number).build(System.out)), true, StandardCharsets.UTF_8);
+                    ps = new PrintStream(new PlainTextConsoleOutputStream(prefixedOutputStreamBuilder.withName(job + '#' + number).build(System.out)), false, StandardCharsets.UTF_8);
+                }
+                synchronized (System.out) {
+                    ps.append(DeltaSupportLogFormatter.elapsedTime());
```
jglick, line 105: ignore ws

Vlatombe (approved):

PR #730, opened 2024-02-13
Add synchronization to `PrefixedOutputStream`

jglick:
Very similar to #728, I have found that near-simultaneous printing of prefixed output lines can wind up causing display corruption and messing up ANSI display codes. This seems to fix it, at least in limited testing.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/PrefixedOutputStream.java
+++ b/src/main/java/org/jvnet/hudson/test/PrefixedOutputStream.java
@@ -92,33 +92,35 @@ public final class PrefixedOutputStream extends LineTransformationOutputStream.D
     }
 
     @Override protected void eol(byte[] b, int len) throws IOException {
-        if (name != null) {
-            out.write('[');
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/PrefixedOutputStream.java
```diff
@@ -92,33 +92,35 @@ private PrefixedOutputStream(OutputStream out, String name, AnsiColor color) {
...
-            out.write(27); // ESC
-            out.write('[');
-            out.write(color.getCode().getBytes(StandardCharsets.US_ASCII));
-            out.write('m');
-            // Preserving original line ending, so not using trimEOL:
-            int preNlLen = len;
-            while (preNlLen > 0 && (b[preNlLen - 1] == '\n' || b[preNlLen - 1] == '\r')) {
-                preNlLen--;
+        synchronized (out) {
+            if (name != null) {
```
jglick, line 96: ignore WS

MarkEWaite (approved):

julieheard (approved):

PR #731, opened 2024-02-18
Accept NPE in MockFolder.createProject() if wrong job type is passed

MarkEWaite:
## Accept NPE in MockFolder.createProject() if wrong job type is passed

The null pointer exception will happen when the test is run and will be an immediate sign to the developer that there is a mistake in the caller.

Detected while prototyping an upgrade of the minimum Jenkins version from 2.361.4 to 2.401.3.  Message is not reported with 2.361.4 as Jenkins minimum version but is reported with Jenkins 2.387.3 or later.

### Testing done

Confirmed that warning is visible when `jenkins.version` is 2.387.3 or higher before this change.  Confirmed that warning is suppressed after this change.  Confirmed that automated tests pass before and after the exclusion was added.

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/MockFolder.java
+++ b/src/main/java/org/jvnet/hudson/test/MockFolder.java
@@ -25,6 +25,7 @@
 package org.jvnet.hudson.test;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Extension;
 import hudson.model.AbstractItem;
 import hudson.model.Action;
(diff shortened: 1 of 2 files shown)


jglick:
What exactly is this about? There are two overloads in https://github.com/jenkinsci/jenkins-test-harness/blob/6fd31af28c2f36223bba6be904683183324f5cb5/src/main/java/org/jvnet/hudson/test/MockFolder.java#L156-L163 and it is not clear from the description which you are referring to, or what the stack trace of the NPE might be, or what sort of caller mistake might trigger the NPE. Rather than suppressing a SpotBugs warning (which BTW we would rather do with `@SuppressFBWarnings` with a `justification` on the affected method), can we just throw an appropriate `IllegalArgumentException` or similar?

MarkEWaite:
> What exactly is this about? There are two overloads in

When I was test driving an upgrade of the minimum Jenkins version from 2.361.x to a newer Jenkins LTS, spotbugs reported:

> [ERROR] Medium: Possible null pointer dereference in org.jvnet.hudson.test.MockFolder.createProject(Class, String) due to return value of called method [org.jvnet.hudson.test.MockFolder, org.jvnet.hudson.\
test.MockFolder] Method invoked at MockFolder.java:[line 162]Known null at MockFolder.java:[line 162] NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE

I've moved the exclusion to the specific method with the issue.  It wasn't obvious to me how I could easily detect the case where an illegal argument exception should be thrown, so I took the easy way out and excluded the warning.

basil:
> which BTW we would rather do with `@SuppressFBWarnings`

We _can_ do so, although I do not see why we would necessarily _rather_ do so. If this is something we would _rather_ do, then I think it should be documented in the developer documentation at https://www.jenkins.io/doc/developer/book/; otherwise, this is merely personal preference.

PR #742, opened 2024-03-22
Adapt to IPV6 envs

raul-arabaolaza:
While running `workflow-durable-task-step` tests on a proprietary environment with only IPV6 networking enabled I found some parts of RJR that were not fully IPV6 compatible. This PR tries to fix those parts so RJR can be used in both IPV4 and IPV6 environments.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -671,7 +671,7 @@ public final class RealJenkinsRule implements TestRule {
         if (port == 0) {
             throw new IllegalStateException("This method must be called after calling #startJenkins.");
         }
-        return new URL("http://" + host + ":" + port + "/jenkins/");
+        return new URL("http", host, port, "/jenkins/");
     }
 
     private URL endpoint(String method) throws MalformedURLException {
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -671,7 +671,7 @@ public URL getUrl() throws MalformedURLException {
         if (port == 0) {
             throw new IllegalStateException("This method must be called after calling #startJenkins.");
         }
-        return new URL("http://" + host + ":" + port + "/jenkins/");
```
raul-arabaolaza, line 674: This failed when `host` was an IPV6 ip

timja (approved):

rsandell (approved):

jtnord (approved):

jglick (approved):
Looks right but @PereBueno why did this not come up during testing of #712?

jtnord:
> Looks right but @PereBueno why did this not come up during testing of #712?

Most likely system configuration.

Some hosts/systems/OSes call `::1` `localhost-6` causing a miss-match between the name and the address when used before this change.

Other hosts have 2 entries for `localhost`, one for v4 and one for v6.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
Vlatombe: I think this line was the only necessary change. The other change caused test regressions in other parts, so I'm proposing a revert of them in https://github.com/jenkinsci/jenkins-test-harness/pull/776

PR #744, opened 2024-03-25
Fix nullability

jtnord:
User.Get has a return of CheckForNull yet the returned value is passed
directly as a parameter of a method marked NonNull.

Whilst the return in this case can never be `null` as the create parameter
is `true`, compilers and tools are not clever enough to understand the
human written contract.  so we assert the user in non null before then
passing it down the line, to keep compilers happy

### Testing done

none - just whatever the CI performs.

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -2846,6 +2846,7 @@ public class JenkinsRule implements TestRule, MethodRule, RootAction {
          */
         public @NonNull WebClient withBasicApiToken(@NonNull String userId){
             User user = User.getById(userId, true);
+            assert user != null;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -2846,6 +2846,7 @@ public URL createCrumbedUrl(String relativePath) throws IOException {
          */
         public @NonNull WebClient withBasicApiToken(@NonNull String userId){
             User user = User.getById(userId, true);
+            assert user != null;
             return withBasicApiToken(user);
```
jtnord, line 2850: `withBasicApiToken` has the `User` parameter marked as `NonNull`.

MarkEWaite (approved):

PR #770, opened 2024-05-16
[JENKINS-73260] Forward compatibility with EE 9 cores

basil:
### Problem

We plan to eventually migrate from Jetty 12 EE 8 to Jetty 12 EE 9, but the test harness only supports cores with Jetty 12 EE 8.

### Solution

Bundle both the Jetty 12 EE 8 and EE 9 modules into the test harness, and dynamically select the module to load at runtime based on the Jenkins core in use.

This change does not increase the minimum core version supported by the test harness. To detect newer cores with EE 9 support and invoke the appropriate methods, reflection is used.

### Implementation

This plugin finds any places where we invoked EE 8 functionality in Jenkins core and turns them into an `if` statement. The existing code is in the `else` block, and the `if` block contains new code that invokes the relevant EE 9 functionality through reflection.

`JavaNetReverseProxy2` and `NoListenerConfiguration2` classes have been created that duplicate the existing classes but with EE 9 imports instead.

### Testing done

Core and plugin test suites have been executed successfully with this PR, with no known failures.

### Notes for reviewers

This change is somewhat challenging to review. The changes to `JenkinsRule` and `HudsonTestCase` look confusing, because the imports have changed, and it looks like a lot of new code has been added, but that new code is really preserving the existing code but with the old imports. The "existing" code has now been changed to add a 2 to the method names and referring to the new imports. So the existing code looks the same but means something different, and the "new" code is actually the existing code, but for compatibility.

### Deployment

This change requires Java 17 or newer, so it will not be backported to the 2225.x branch of the test harness or released in a plugin parent POM until such a time that we require Java 17 or newer for Jenkins plugin development.

It _will_ be used in the core release that adopts Jetty 12 EE 9.

### Future work

Once we drop support for EE 8 in the test harness, the code in this change can be simplified. The EE 8 branches of `if` statements can all be deleted, and the reflection-based code can be transformed to directly invoke the relevant methods, much like the EE 8 code does today.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
+++ b/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
@@ -90,6 +90,8 @@ import hudson.util.PersistedList;
 import hudson.util.ReflectionUtils;
 import hudson.util.StreamTaskListener;
 import hudson.util.jna.GNUCLibrary;
+import jakarta.servlet.ServletContext;
+import jakarta.servlet.ServletContextEvent;
 import java.beans.PropertyDescriptor;
 import java.io.BufferedReader;
 import java.io.File;
@@ -100,6 +102,7 @@ import java.lang.management.ThreadInfo;
 import java.lang.reflect.Array;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
@@ -126,8 +129,6 @@ import java.util.logging.Filter;
 import java.util.logging.Level;
 import java.util.logging.LogRecord;
 import java.util.logging.Logger;
-import javax.servlet.ServletContext;
-import javax.servlet.ServletContextEvent;
 import jenkins.model.Jenkins;
 import jenkins.model.JenkinsAdaptor;
 import jenkins.model.JenkinsLocationConfiguration;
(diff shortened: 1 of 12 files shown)


src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
```diff
@@ -583,6 +657,52 @@ protected ClassLoader configureClassLoader(ClassLoader loader) {
         return context.getServletContext();
     }
 
+    /**
```
basil, line 660: Not new code, so please only review the _diff_ of this method with the `createWebServer2` method (left as an exercise for the reader).

src/main/java/org/jvnet/hudson/test/JavaNetReverseProxy2.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Acts as a reverse proxy, so that during a test we can avoid hitting updates.jenkins.io.
+ *
+ * <p>
+ * The contents are cached locally.
+ *
+ * @author Kohsuke Kawaguchi
+ */
+public class JavaNetReverseProxy2 extends HttpServlet {
```
basil, line 28: Not new code, so please only review the _diff_ of this class with the `JavaNetReverseProxy` class (left as an exercise for the reader).

src/main/java/org/jvnet/hudson/test/NoListenerConfiguration2.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Kills off the {@link WebAppMain} {@link ServletContextListener}.
+ *
+ * <p>
+ * This is so that the harness can create the {@link jenkins.model.Jenkins} object.
+ * with the home directory of our choice.
+ *
+ * @author Kohsuke Kawaguchi
+ */
+public class NoListenerConfiguration2 extends AbstractLifeCycle {
```
basil, line 41: Not new code, so please only review the _diff_ of this class with the `NoListenerConfiguration` class (left as an exercise for the reader).

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -895,6 +973,130 @@ protected ClassLoader configureClassLoader(ClassLoader loader) {
         return context;
     }
 
+    /**
```
basil, line 976: Not new code, so please only review the _diff_ of these methods with the "2" versions above (left as an exercise for the reader).

src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
```diff
@@ -1763,7 +1883,7 @@ public String getContextPath() throws IOException {
         public WebRequest addCrumb(WebRequest req) {
             NameValuePair crumb = new NameValuePair(
                     jenkins.getCrumbIssuer().getDescriptor().getCrumbRequestField(),
-                    jenkins.getCrumbIssuer().getCrumb( null ));
+                    jenkins.getCrumbIssuer().getCrumb((javax.servlet.ServletRequest) null));
```
basil, line 1886: Here and elsewhere, newer cores will have two versions of this method, one for `javax` and one for `jakarta`, so to be forward compatible we have to disambiguate the method call. Here we choose the older import, since this change does have to work on non-Jakarta cores.

MarkEWaite (approved):

PR #772, opened 2024-05-19
Update search function for Command palette

janfaracik:
Updates the search method to support the 'Command palette' introduced n https://github.com/jenkinsci/jenkins/pull/7569. Also includes backwards compatibility for the classic YUI search bar.

### Submitter checklist

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
+++ b/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
@@ -1594,9 +1594,15 @@ public abstract class HudsonTestCase extends TestCase implements RootAction {
 
         public HtmlPage search(String q) throws IOException, SAXException {
             HtmlPage top = goTo("");
-            HtmlForm search = top.getFormByName("search");
-            search.getInputByName("q").setValue(q);
-            return (HtmlPage)HtmlFormUtil.submit(search, null);
+
+            HtmlButton button = top.querySelector("#button-open-command-palette");
+            button.click();
+
+            HtmlInput search = top.querySelector("#command-bar");
+            search.setValue(q);
+
+            HtmlLink firstResult = top.querySelector(".jenkins-command-palette__results__item");
+            return firstResult.click();
         }
 
         /**
--- a/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
+++ b/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
@@ -1602,6 +1602,8 @@ public abstract class HudsonTestCase extends TestCase implements RootAction {
             while (element == null && System.currentTimeMillis() - startTime < maxWaitTime) {
                 // Try to find the element
                 try {
+                    System.out.println("Current HTML:");
+                    System.out.println(page.getWebResponse().getContentAsString());
                     element = page.querySelector(query);
                 } catch (Exception e) {
                     // Ignore exceptions and continue waiting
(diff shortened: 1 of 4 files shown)


src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
```diff
@@ -1592,11 +1593,25 @@ public void run() {
...
+                search.getInputByName("q").setValue(q);
+                return (HtmlPage)HtmlFormUtil.submit(search, null);
+            }
+
+            button.click();
+
+            HtmlInput search = top.querySelector("#command-bar");
+            search.setValue(q);
+
+            Thread.sleep(100);
```
timja, line 1612: avoid sleeps and wait for expectedconditions as different speed machines may take different amount of time

timja, line 1612: would be good if we could avoid this sleep, is there something like seleniums expectedconditions available? 

this may introduce flakiness if it takes longer due to something like running lots of tests or a slow machine

timja: Likely cause of failures in https://ci.jenkins.io/blue/organizations/jenkins/Core%2Fjenkins/detail/PR-7569/57/tests/

e.g. 

```
Error
Cannot invoke "org.htmlunit.html.HtmlLink.click()" because "firstResult" is null
Stacktrace
java.lang.NullPointerException: Cannot invoke "org.htmlunit.html.HtmlLink.click()" because "firstResult" is null
	at org.jvnet.hudson.test.JenkinsRule$WebClient.search(JenkinsRule.java:2647)
	at hudson.search.SearchTest.testFailure(SearchTest.java:74)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:656)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

timja: although looking more I think its that this function doesn't handle it when the test is trying to have no results

timja: (not convinced this works at all)

src/main/java/org/jvnet/hudson/test/QueryUtils.java
```diff
@@ -0,0 +1,96 @@
...
+     * @param page the page
+     * @param value the value to find
+     * @throws RuntimeException if string is not present after three seconds
+     */
+    public static void waitUntilStringIsPresent(HtmlPage page, String value) {
+        long maxWaitTime = TimeUnit.SECONDS.toMillis(3);
+        long startTime = System.currentTimeMillis();
+
+        while (System.currentTimeMillis() - startTime < maxWaitTime) {
+            if (page.querySelector("*").getTextContent().contains(value)) {
```
timja, line 23: can probably be simplified with `asText()`? https://stackoverflow.com/a/6538127/4951015

janfaracik: Couldn't find the `asText` method, `getVisibleText` exists however so I've updated it with that

timja (approved):

src/main/java/org/jvnet/hudson/test/QueryUtils.java
```diff
@@ -0,0 +1,96 @@
...
+        while (System.currentTimeMillis() - startTime < maxWaitTime) {
+            if (page.querySelector("*").getVisibleText().contains(value)) {
+                try {
+                    Thread.sleep(100);
+                } catch (InterruptedException e) {
+                    throw new RuntimeException(e);
+                }
+            } else {
+                System.out.println("Took '" + (System.currentTimeMillis() - startTime) + "ms' " +
+                        "until string '" + value + "' was no longer present");
```
daniel-beck: Leftover development output?

PR #776, opened 2024-05-31
IPV6 follow-up

Vlatombe:
Reverts part of #742

Only leaves the relevant part which was to switch to the URL(String,String,String,String) constructor.

The change from localhost to ip caused some test failures in Kubernetes context (It is invalid to provide an IP address as part of an ingress host name)

### Submitter checklist

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -185,7 +185,7 @@ public final class RealJenkinsRule implements TestRule {
 
     private int timeout = Integer.getInteger("jenkins.test.timeout", new DisableOnDebug(null).isDebugging() ? 0 : 600);
 
-    private String host = httpListenAddress;
+    private String host = "localhost";
 
     Process proc;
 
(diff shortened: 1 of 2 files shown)


timja (approved):

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -292,13 +292,13 @@ public RealJenkinsRule withTimeout(int timeout) {
 
     /**
      * Sets a custom host name for the Jenkins root URL.
-     * <p>By default, this is the same as the {@link #httpListenAddress}.
-     * But you may wish to set it to something else that resolves to the loopback address,
+     * <p>By default, this is just {@code localhost}.
+     * But you may wish to set it to something else that resolves to localhost,
      * such as {@code some-id.127.0.0.1.nip.io}.
```
jglick, line 297:
```suggestion
     * such as {@code some-id.localtest.me}.
```
while we are here.

jglick (approved):
OK I think? I guess you ran at least the new test in an IPv6 VM? Do we have any way of verifying that this does not regress some random plugin’s tests?

amuniz (approved):

Vlatombe:
> the new test in an IPv6 VM

I lack the setup to do that, but in my experience running with `-Djava.net.preferIPv4Stack=false -Djava.net.preferIPv6Addresses=true` behaves pretty much the same.

imonteroperez (approved):

raul-arabaolaza (approved):

PR #779, opened 2024-06-05
Update the minimum version of Jenkins used by RealJenkinsRule to prevent detached plugins being dragged in.

jtnord:
As observed in https://github.com/jenkinsci/jenkins/pull/9351/checks?check_run_id=25813003911 the RealJenkinsRule was targetting an version of Jenkins core that required detached plugins.

This updates the version in the `RealJenkinsRuleInit` plugin to a version that does not require detached plugins.

Whilst this is an interim solution to the problem, the currently selected version made no sense to me as it was lower than the Jenkins version specified for the actual test harness.

Longer term (if needed) the test HPI could be unpacked, and have its `MANIFEST.mf` optionally updated to match the Jenkins version under test

incremental consumed in https://github.com/jenkinsci/jenkins/pull/9351

see also #780

### Submitter checklist


RealJenkinsRuleInit/pom.xml
```diff
@@ -4,16 +4,15 @@
...
         <relativePath />
     </parent>
     <groupId>io.jenkins.plugins</groupId>
     <artifactId>RealJenkinsRuleInit</artifactId>
     <version>0-SNAPSHOT</version>
     <packaging>hpi</packaging>
     <properties>
-        <jenkins.version>2.249</jenkins.version>
-        <java.level>8</java.level>
+        <jenkins.version>2.361</jenkins.version>
```
jtnord, line 15: version [has no detached plugins as of today,](https://github.com/jenkinsci/jenkins/blob/7aad67832442b8cd1dfaf363a3630d2687ddf3e3/core/src/main/resources/jenkins/split-plugins.txt) and matches the Jenkins version that the j-t-h itself uses

RealJenkinsRuleInit/src/main/resources/index.jelly
jtnord, line 1: needed for the parent bump

RealJenkinsRuleInit/pom.xml
```diff
@@ -4,16 +4,15 @@
...
+        <version>4.83</version>
         <relativePath />
     </parent>
     <groupId>io.jenkins.plugins</groupId>
     <artifactId>RealJenkinsRuleInit</artifactId>
     <version>0-SNAPSHOT</version>
     <packaging>hpi</packaging>
     <properties>
-        <jenkins.version>2.249</jenkins.version>
-        <java.level>8</java.level>
```
jtnord, line 16: newer version uses java11 and can not set the `java.level` to `11` as it causes `unknown java version 1.11` when compiling.
so updated the parent

RealJenkinsRuleInit/pom.xml
jtnord: this reveals at least one test was relying on this dragging in detached plugins (`instance-identity`) as [it was creating](https://github.com/jenkinsci/jenkins/pull/9351/commits/fe2eaa52bb98d6a4038841aac42f5c28cc70b861) classic inbound agents which would fail with this update until `instance-identity` was added back as a test plugin.

timja (approved):

MarkEWaite (approved):

RealJenkinsRuleInit/pom.xml
MarkEWaite: I was surprised that the earlier version was not based on an LTS and neither is the new version.  Is there a specific reason this does not use an LTS version?

PR #781, opened 2024-06-05
Add option to bump baseline of rjrinit plugin default to true

jtnord:
https://github.com/jenkinsci/jenkins-test-harness/pull/780#discussion_r1627958793

to get an incremental build that can then be used in the bom
https://github.com/jenkinsci/bom/pull/3284

### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #782, opened 2024-06-06
Dynamically create the RealJenkinsRuleInit plugin

jtnord:
Switch to dynamically creating the plugin used by RealJenkinsRule

The plugin will have its Jenkis baseline set to the version of the war to ensure that no detached plugins are pulled in.

Previously if the version of Jenkins used by RealJenkinsRuleInit has detached plugins in the version of Jenknis used for the test then this can uncessescarily pull in detached plugins.

In most cases this is harmless, but in some cases these plugins may be prevented from running under the test (e.g. testing FIPS and one of the detached plugins like eddsa is not FIPS complaint and blocks startup).

The plugin itself has limited exposure to the Jenkins api (via hudson.plugin.Plugin and RealJenkinsRule$Init2).

Code running from the plugin tests inside RealJenkinsRule useis the uberClassloader and if the plugin was targetting a lower core (where the dependencies have been deteached, they should still be included as this change only affects the version of the RealJenkinsRuleInit plugin.

based on a suggestion by @jglick https://github.com/jenkinsci/jenkins-test-harness/pull/780#discussion_r1628438844

obsoletes: #781 

### Submitter checklist


src/main/java/org/jvnet/hudson/test/PluginUtils.java
```diff
@@ -0,0 +1,81 @@
...
+        Path tmpClassesJar = Files.createTempFile("rjr", "jar");
+        try (FileOutputStream fos = new FileOutputStream(tmpClassesJar.toFile());
+                JarOutputStream classesJarOS = new JarOutputStream(fos, mf)) {
+
+            // the actual class
+            try (InputStream classIS = RealJenkinsRuleInit.class.getResourceAsStream(RealJenkinsRuleInit.class.getSimpleName() + ".class")) {
+                String path = RealJenkinsRuleInit.class.getPackageName().replace('.', '/');
+                createJarEntry(classesJarOS, path + '/' + RealJenkinsRuleInit.class.getSimpleName() + ".class", classIS);
+            }
+            createJarEntry(classesJarOS, "META-INF/services/hudson.Plugin", RealJenkinsRuleInit.class.getName().getBytes(StandardCharsets.UTF_8));
```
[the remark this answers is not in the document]
jtnord: https://github.com/jenkinsci/jenkins/blame/f8e55c78f8b89678131450d302e31c4f7805eac4/core/src/main/java/hudson/PluginWrapper.java#L912-L914C6 shows Plugin-Class has been there for 15 years...
https://github.com/jenkinsci/jenkins/blame/f8e55c78f8b89678131450d302e31c4f7805eac4/core/src/main/java/jenkins/PluginSubtypeMarker.java creates the service entry...  (again for a long time).

https://github.com/jenkinsci/jenkins/blame/f8e55c78f8b89678131450d302e31c4f7805eac4/core/src/main/java/hudson/util/Service.java#L46 was deprecated 7 years ago and https://github.com/search?q=repo%3Ajenkinsci%2Fjenkins%20ServiceLoader.load&type=code doesn't show anything attempting to load a `hudson.Plugin` service.   https://github.com/jenkinsci/jenkins/blob/f8e55c78f8b89678131450d302e31c4f7805eac4/core/src/main/java/hudson/PluginManager.java#L1334 is also seemingly uncalled so I _guess_ it is not needed

But whilst the service entry is still generated by Jenkins I think we should possibly keep it.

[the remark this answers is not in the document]
jtnord: aah.. https://github.com/jenkinsci/maven-hpi-plugin/blob/4847bf204e60204e708df5c12f5f9a8d3a6bf6d3/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsManifestMojo.java#L140-L149

so the services entry is just there to support the hpi plugin getting the info to write a manifest...

src/main/java/org/jvnet/hudson/test/RealJenkinsRuleInit.java
```diff
@@ -27,16 +27,27 @@
...
+
 import jenkins.model.Jenkins;
 
+/**
+ * Plugin for use by {@link RealJenkinsRule}.
+ * <p>
+ * <stong>NOTE</strong>: this and only this class is added into a dynamically generated plugin, see {@link PluginUtils#createRealJenkinsRulePlugin(java.io.File, String)}.
+ * In order for this to occur correctly there need to be no inner classes or other code dependencies here (except what can be loaded by reflection).
+ */
+@Restricted(NoExternalUse.class)
```
jglick, line 42: This will have no effect since the access control checker only runs on `*-plugin/src/main/java/**/*.java` not `*-plugin/src/test/java/**/*Test.java`. Might be useful as a reminder to humans alongside the Javadoc but that is it. You could consider prefixing the class name with `_` or something to emphasize that code is not intended to link against it.

src/main/java/org/jvnet/hudson/test/RealJenkinsRuleInit.java
```diff
@@ -27,16 +27,27 @@
...
+
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+
 import jenkins.model.Jenkins;
 
+/**
+ * Plugin for use by {@link RealJenkinsRule}.
+ * <p>
+ * <stong>NOTE</strong>: this and only this class is added into a dynamically generated plugin, see {@link PluginUtils#createRealJenkinsRulePlugin(java.io.File, String)}.
```
jglick, line 39:
```suggestion
 * <strong>NOTE</strong>: this and only this class is added into a dynamically generated plugin, see {@link PluginUtils#createRealJenkinsRulePlugin(java.io.File, String)}.
```

jglick (approved):

src/main/java/org/jvnet/hudson/test/RealJenkinsRuleInit.java
jtnord: move package to "do_not_consume" ?

jglick: Maybe, or some `impl` package. I am not particularly concerned. Might suffice to clarify in the Javadoc summary sentence (prior to first `.` IIRC) that test code should not refer to the class.

src/main/java/org/jvnet/hudson/test/RealJenkinsRuleInit.java
```diff
@@ -27,16 +27,27 @@
...
 import java.net.URL;
 import java.net.URLClassLoader;
+
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+
 import jenkins.model.Jenkins;
 
+/**
+ * Plugin for use by {@link RealJenkinsRule}.
```
jtnord, line 37:
```suggestion
 * Plugin for use <b>internally only</b> by {@link RealJenkinsRule}, do not use this from plugin test code!
```

src/main/java/org/jvnet/hudson/test/RealJenkinsRuleInit.java
jtnord: defc0f73e753897698cb9fe3dac086b4461f7fa8

jglick:
This seems to break tests which deliberately run an older Jenkins version (`withWar`) for purposes of testing compatibility. Maybe we can get a new method in RJR to override `targetJenkinsVersion`?

PR #795, opened 2024-07-05
Use `ProcessTree.killAll` from `InboundAgentRule`

jglick:
Ensures that even if an inbound agent starts some subprocesses, they are reliably killed off when the test completes.


dwnusbaum (approved):

basil:
Breaks the following tests (and thus my EE 9 test runs):

- `org.jenkinsci.plugins.workflow.support.steps.AgentErrorConditionTest#retryNodeBlockSynch`
- `org.jenkinsci.plugins.workflow.support.steps.ExecutorStepTest#buildShellScriptAcrossDisconnect`
- `org.jenkinsci.plugins.workflow.support.steps.ExecutorStepTest#contextualizeFreshFilePathAfterAgentReconnection`

PR #797, opened 2024-07-09
Move compatibility layer into separate package

basil:


basil:
Closing in favor of #798

PR #802, opened 2024-07-10
Limiting `ProcessTree.killAll` to `InboundAgentRule.after`

jglick:
https://github.com/jenkinsci/jenkins-test-harness/pull/795#issuecomment-2219071287 I neglected to consider direct call to `InboundAgentRule.stop` from test code, which should _not_ kill subprocesses, only the agent JVM itself. Tested in `workflow-durable-task-step` on all test cases in `AgentErrorConditionTest` & `ExecutorStepTest`; also verified that the proprietary test suite originally motivating #795 continues to pass with this, and that subprocesses are still terminated at the end of the test.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
+++ b/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
@@ -43,13 +43,13 @@ import java.io.File;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
-import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
```diff
@@ -80,7 +80,7 @@ public final class InboundAgentRule extends ExternalResource {
     private static final Logger LOGGER = Logger.getLogger(InboundAgentRule.class.getName());
 
     private final String id = UUID.randomUUID().toString();
-    private final ConcurrentMap<String, Process> procs = new ConcurrentHashMap<>();
+    private final Map<String, Process> procs = Collections.synchronizedMap(new HashMap<>());
```
jglick, line 83: (to allow `null` values)

basil (approved):

Vlatombe (approved):

basil:
Thank you very much! Confirming that my Jetty 12 test matrix is back to green after this PR.

PR #825, opened 2024-08-31
[JEP-237] Add new method to RealJenkinsRule to start in FIPS mode

olamy:
https://github.com/jenkinsci/jep/tree/master/jep/237

Signed-off-by: Olivier Lamy <olamy@apache.org>

### Submitter checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -316,6 +316,32 @@ THE SOFTWARE.
               <overWriteSnapshots>true</overWriteSnapshots>
             </configuration>
           </execution>
+          <execution>
+            <id>copy-fips-libs</id>
+            <goals>
+              <goal>copy</goal>
+            </goals>
+            <configuration>
+              <artifactItems>
+                <artifactItem>
+                  <groupId>org.bouncycastle</groupId>
+                  <artifactId>bc-fips</artifactId>
+                  <version>1.0.2.4</version>
+                </artifactItem>
+                <artifactItem>
+                  <groupId>org.bouncycastle</groupId>
+                  <artifactId>bcpkix-fips</artifactId>
+                  <version>1.0.7</version>
+                </artifactItem>
+                <artifactItem>
+                  <groupId>org.bouncycastle</groupId>
+                  <artifactId>bctls-fips</artifactId>
+                  <version>1.0.17</version>
+                </artifactItem>
+              </artifactItems>
+              <stripVersion>true</stripVersion>
+            </configuration>
+          </execution>
         </executions>
       </plugin>
       <plugin>
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -498,13 +498,11 @@ public final class RealJenkinsRule implements TestRule {
                             fipsLibrariesPath.resolve("bctls-fips.jar").toFile(),
                             fipsLibrariesPath.resolve("bcpkix-fips.jar").toFile());
         try {
-            javaOptions("--add-exports", "java.base/sun.security.provider=ALL-UNNAMED",
-                            "-Dsecurity.overridePropertiesFile=true",
+            javaOptions(    "-Dsecurity.overridePropertiesFile=true",
                             "-Djava.security.properties=" + writeFIPSJavaSecurityFile().toUri(),
                             "-Dorg.bouncycastle.fips.approved_only=true",
                             "-Djavax.net.ssl.trustStoreType=PKCS12",
-                            "-Djenkins.security.FIPS140.COMPLIANCE=true",
-                            "-Dcom.redhat.fips=false");
+                            "-Djenkins.security.FIPS140.COMPLIANCE=true");
         } catch (IOException e) {
             throw new RuntimeException(e);
         }
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -873,9 +873,6 @@ public final class RealJenkinsRule implements TestRule {
                     + (debugPort > 0 ? ",address=" + httpListenAddress + ":" + debugPort : ""));
         }
         if(!bootClasspathFiles.isEmpty()) {
-            //String fileList = String.join(File.pathSeparator, bootClasspathFiles.stream()
-            //        .map(file -> FilenameUtils.separatorsToSystem(file.getAbsolutePath())).toList());
-
             String fileList = bootClasspathFiles.stream().map(File::getAbsolutePath).collect(Collectors.joining(File.pathSeparator));
             argv.add("-Xbootclasspath/a:" + fileList);
 
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -28,6 +28,7 @@ import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.ExtensionList;
+import hudson.Functions;
 import hudson.model.UnprotectedRootAction;
 import hudson.security.ACL;
 import hudson.security.ACLContext;
@@ -97,6 +98,7 @@ import jenkins.model.Jenkins;
 import jenkins.model.JenkinsLocationConfiguration;
 import jenkins.util.Timer;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.FilenameUtils;
 import org.junit.Assume;
 import org.junit.AssumptionViolatedException;
 import org.junit.rules.DisableOnDebug;
@@ -495,11 +497,16 @@ public final class RealJenkinsRule implements TestRule {
                 !Files.exists(this.fipsLibrariesPath.resolve("bcpkix-fips.jar"))) {
             throw fipsIllegalSetupException(this.fipsLibrariesPath);
         }
+
         Path bcFips = fipsLibrariesPath.resolve("bc-fips.jar");
         Path bcTLFips = fipsLibrariesPath.resolve("bctls-fips.jar");
         Path bcpkixFips = fipsLibrariesPath.resolve("bcpkix-fips.jar");
 
-        String xBootClasspath = "-Xbootclasspath/a:"+ bcFips.toFile().getAbsolutePath() +":" + bcTLFips.toFile().getAbsolutePath() +":"+bcpkixFips.toFile().getAbsolutePath();
+        String pathSeparator = Functions.isWindows() ? ";" : ":";
+
+        String xBootClasspath = "-Xbootclasspath/a:"+ FilenameUtils.separatorsToSystem(bcFips.toFile().getAbsolutePath())
+                + pathSeparator + FilenameUtils.separatorsToSystem(bcTLFips.toFile().getAbsolutePath())
+                + pathSeparator + FilenameUtils.separatorsToSystem(bcpkixFips.toFile().getAbsolutePath());
         try {
             javaOptions(xBootClasspath, "--add-exports", "java.base/sun.security.provider=ALL-UNNAMED",
                             "-Dsecurity.overridePropertiesFile=true",
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -25,6 +25,7 @@
 package org.jvnet.hudson.test;
 
 import edu.umd.cs.findbugs.annotations.CheckForNull;
+import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.ExtensionList;
 import hudson.model.UnprotectedRootAction;
@@ -65,6 +66,8 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Properties;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
@@ -94,6 +97,7 @@ import jenkins.model.Jenkins;
 import jenkins.model.JenkinsLocationConfiguration;
 import jenkins.util.Timer;
 import org.apache.commons.io.FileUtils;
+import org.junit.Assume;
 import org.junit.AssumptionViolatedException;
 import org.junit.rules.DisableOnDebug;
 import org.junit.rules.TestRule;
@@ -200,6 +204,9 @@ public final class RealJenkinsRule implements TestRule {
 
     private boolean prepareHomeLazily;
     private boolean provisioned;
+    private boolean fipsEnabled;
+    private Path fipsLibrariesPath;
+
 
     // TODO may need to be relaxed for Gradle-based plugins
     private static final Pattern SNAPSHOT_INDEX_JELLY = Pattern.compile("(file:/.+/target)/classes/index.jelly");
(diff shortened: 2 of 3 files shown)


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -457,6 +464,116 @@ public RealJenkinsRule prepareHomeLazily(boolean prepareHomeLazily) {
...
+     *   <li>-Djenkins.security.FIPS140.COMPLIANCE=true</li>
+     *   <li>-Dcom.redhat.fips=false</li>
+     * </ul>
+     * @param fipsLibrariesPath path to directory with FIPS BouncyCastle jars
+     * @return
+     */
+    public RealJenkinsRule withFIPSEnabled(@NonNull Path fipsLibrariesPath) {
+        this.fipsLibrariesPath = Objects.requireNonNull(fipsLibrariesPath);
+        // check expected BouncyCastle fips jars are here
+        if  (!Files.exists(this.fipsLibrariesPath) ||
```
olamy, line 492: another option would have been to use maven-resolver API and download files for the user.  But this would mean we force the jars version here.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -744,8 +870,14 @@ public void startJenkins() throws Throwable {
                     + ",suspend=" + (debugSuspend ? "y" : "n")
                     + (debugPort > 0 ? ",address=" + httpListenAddress + ":" + debugPort : ""));
         }
-        argv.addAll(javaOptions);
+        if(!bootClasspathFiles.isEmpty()) {
+            String pathSeparator = Functions.isWindows() ? ";" : ":";
+            String fileList = String.join(pathSeparator, bootClasspathFiles.stream()
+                    .map(file -> FilenameUtils.separatorsToSystem(file.getAbsolutePath())).toList());
+            javaOptions("-Xbootclasspath/a:" + fileList);
```
olly-cb, line 877: or `argv.add()`

olamy, line 877: or `argv.add()`

pom.xml
```diff
@@ -316,6 +316,32 @@ THE SOFTWARE.
...
+            <id>copy-fips-libs</id>
+            <goals>
+              <goal>copy</goal>
+            </goals>
+            <configuration>
+              <artifactItems>
+                <artifactItem>
+                  <groupId>org.bouncycastle</groupId>
+                  <artifactId>bc-fips</artifactId>
+                  <version>1.0.2.4</version>
```
jtnord, line 329: you may want a TODO here, as this is not the latest version of `bc-fips` but for a valid reason.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -457,6 +464,125 @@ public RealJenkinsRule prepareHomeLazily(boolean prepareHomeLazily) {
         return this;
     }
 
+    /**
+     * Use {@link #withFIPSEnabled(Path)} with default value of {@code target/dependency}
+     * @return
```
jtnord, line 469: @return on its own is an invalid javadoc comment and will produce warnings, please add some text after it or remove it.

```suggestion
     * @return This real Jenkins rule.
```

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -457,6 +464,125 @@ public RealJenkinsRule prepareHomeLazily(boolean prepareHomeLazily) {
...
+     * <ul>
+     *   <li>-Xbootclasspath with FIPS BouncyCastle jars</li>
+     *   <li>java.security.properties to a new temp file configuring the security provider to FIPS BouncyCastle</li>
+     *   <li>-Dsecurity.overridePropertiesFile=true</li>
+     *   <li>-Djavax.net.ssl.trustStoreType=PKCS12</li>
+     *   <li>-Djenkins.security.FIPS140.COMPLIANCE=true</li>
+     *   <li>-Dcom.redhat.fips=false</li>
+     * </ul>
+     * @param fipsLibrariesPath path to directory with FIPS BouncyCastle jars
+     * @return
```
jtnord, line 487: @return on its own is an invalid javadoc comment and will produce warnings, please add some text after it or remove it.

```suggestion
     * @return This real Jenkins rule.
```

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jtnord: the java security and version of the jars may well be intertwined, so I am not sure that forcing the versions here would be a bad thing.
Additionally with the current setup all users need to duplicate the BC FIPS libraries in their pom config?

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -457,6 +464,125 @@ public RealJenkinsRule prepareHomeLazily(boolean prepareHomeLazily) {
...
+                            "-Djavax.net.ssl.trustStoreType=PKCS12",
+                            "-Djenkins.security.FIPS140.COMPLIANCE=true",
+                            "-Dcom.redhat.fips=false");
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+        return this;
+    }
+
+    public RealJenkinsRule withBootClasspath(File...files) {
```
jtnord, line 516: NIT: why not just a list of `Path`s here and save all the conversions?

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -457,6 +464,125 @@ public RealJenkinsRule prepareHomeLazily(boolean prepareHomeLazily) {
...
+                "      <id>copy-libs</id>\n" +
+                "      <goals>\n" +
+                "        <goal>copy</goal>\n" +
+                "      </goals>\n" +
+                "      <configuration>\n" +
+                "        <artifactItems>\n" +
+                "          <artifactItem>\n" +
+                "            <groupId>org.bouncycastle</groupId>\n" +
+                "            <artifactId>bc-fips</artifactId>\n" +
+                "            <version>1.0.2.4</version>\n" +
```
jtnord, line 539: :-)   you didn't want to hard code the version in an above method, but the suggestions are hard coded here!

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -457,6 +464,125 @@ public RealJenkinsRule prepareHomeLazily(boolean prepareHomeLazily) {
...
+            properties.put("security.provider.2", "org.bouncycastle.jsse.provider.BouncyCastleJsseProvider fips:BCFIPS");
+            properties.put("security.provider.3", "sun.security.provider.Sun");
+            properties.put("fips.provider.1", "org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider C:HYBRID;ENABLE{All};");
+            properties.put("fips.provider.2", "org.bouncycastle.jsse.provider.BouncyCastleJsseProvider fips:BCFIPS");
+            properties.put("keystore.type", "BCFKS");
+            //properties.put("securerandom.strongAlgorithms", "PKCS11:SunPKCS11-NSS-FIPS");
+            properties.put("security.useSystemPropertiesFile", "false");
+            properties.put("ssl.KeyManagerFactory.algorithm", "PKIX");
+            properties.put("fips.keystore.type", "BCFKS");
+            properties.store(outputStream, "");
```
jtnord, line 581: this is missing a lot more, you would need to remove all security providers except the 3 (and same for fips..) to get something that does not have the non FIPS providers.

Additionally, the changes here may also be JRE version (and vendor) specific (we needed to make some changes between java 17 and 21 IIRC)

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleFIPSTest.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Arrays;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+
+public class RealJenkinsRuleFIPSTest {
+
+    @Rule public RealJenkinsRule rr = new RealJenkinsRule().prepareHomeLazily(true)
+            .withDebugPort(4001).withDebugServer(false)
```
jtnord, line 43: is this part of the test?

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -457,6 +464,125 @@ public RealJenkinsRule prepareHomeLazily(boolean prepareHomeLazily) {
...
+                            fipsLibrariesPath.resolve("bctls-fips.jar").toFile(),
+                            fipsLibrariesPath.resolve("bcpkix-fips.jar").toFile());
+        try {
+            javaOptions("--add-exports", "java.base/sun.security.provider=ALL-UNNAMED",
+                            "-Dsecurity.overridePropertiesFile=true",
+                            "-Djava.security.properties=" + writeFIPSJavaSecurityFile().toUri(),
+                            "-Dorg.bouncycastle.fips.approved_only=true",
+                            "-Djavax.net.ssl.trustStoreType=PKCS12",
+                            "-Djenkins.security.FIPS140.COMPLIANCE=true",
+                            "-Dcom.redhat.fips=false");
```
jtnord, line 509: if you are disabling the RHEL patches, then you should not need to adjust `fips.provider.*` (or the add-exports option).

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -744,8 +870,14 @@ public void startJenkins() throws Throwable {
                     + ",suspend=" + (debugSuspend ? "y" : "n")
                     + (debugPort > 0 ? ",address=" + httpListenAddress + ":" + debugPort : ""));
         }
-        argv.addAll(javaOptions);
+        if(!bootClasspathFiles.isEmpty()) {
+            String pathSeparator = Functions.isWindows() ? ";" : ":";
```
jtnord, line 874: [`File.pathSeparator`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/File.html#pathSeparator) :)

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -744,8 +870,14 @@ public void startJenkins() throws Throwable {
...
+            String fileList = String.join(pathSeparator, bootClasspathFiles.stream()
+                    .map(file -> FilenameUtils.separatorsToSystem(file.getAbsolutePath())).toList());
```
jtnord, line 876: this makes my head hurt.  bootClasspathFiles is a list of files (could just be paths, see above)
and you are using a stream to create a list to then join.. but the stream can do it all.
Additionally I am not sure why FilenameUtils.separatorsToSystem is being used here - as the File/Path should already be correct as the path comes from the system and has been made absolute)?

For example

```java
String fileList = bootClasspathFiles.stream().map(File::getAbsolutePath).collect(Collectors.joining(File.pathSeparator));
```

jtnord (commented):
Thanks for this @olamy 

There's some issues with the java.security file (contents and lack of cleanup, and the likelihood this this is java version sepcific?) and some generic code comments

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
olamy: it's suggestions only. users are free to test latest version of bc if they need.

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleFIPSTest.java
olamy: nope but for me this doesn't have any effect except running the test using `-Dmaven.surefire.debug` or debug in your IDE, at least you want to debug everything is ready and no need to change code for it. 
Convenient as if you want to debug a RealJenkinsRule you really want to debug the server

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
olamy: this `@returns` is here just because of the IDE automatically adding.
TBH  it has almost no value. So I prefer to remove it and follow the pattern already used in the class.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
olamy: arf

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
olamy: this hurt my head a lot to get this working on windows :)
maybe this would work on windows I can have a try

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
olamy: fixed

jtnord:
Mentioning @basil in case he wants to review as he left comments on the testing section of the fips jep (which I still need to update) which this indirectly addresses

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -457,6 +464,125 @@ public RealJenkinsRule prepareHomeLazily(boolean prepareHomeLazily) {
...
+    }
+
+    private Path writeFIPSJavaSecurityFile() throws IOException {
+        String javaHome = System.getProperty("java.home");
+        if(javaHome == null) {
+            throw new IllegalArgumentException("Cannot find java.home property");
+        }
+        Path javaSecurity = Paths.get(javaHome, "conf", "security", "java.security");
+        Properties properties = new Properties();
+        Path securityFile = Files.createTempFile("java", ".security");
```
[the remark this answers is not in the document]
olamy: Good point. 
something as simple as `securityFile.toFile().deleteOnExit();` should work

LOL there is another culprit somewhere :) 
```
olamy@pop-os:~/dev/sources/jenkins/jenkins-test-harness$ ls -lrt /tmp/java*.security | wc -l
120
olamy@pop-os:~/dev/sources/jenkins/jenkins-test-harness$ ls -lrt /tmp/jenkins* | wc -l
980
```

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
olamy: removal done.
I have tested this with both 17 and 21 and btw this is building on ci.jenkins.io with 17 and 21.
The idea is to load the current java.security file, modify it to our needs, and then use this modified one only. 
Maybe this will have some issues on some exotic JVM distro manually tailored, but the goal is to have it working for most Jenkins developers and most CIs using this. 
And it's probably the case with this basic, simple approach. 
Let's do this first and let's see if we have issues reported.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
olamy: Yes but at least users have the freedom to eventually test some upgrades to new version (note it's in the plugin configuration so hopefully dependabot or other tools will not propose any upgrade)
The problem with using something to download dependencies (because dependencies may not be in the local repository) is that all the setup must be available here (such as the path to the used settings.xml, which can be different from `~/.m2/settings.xml`), cli request properties, the content of  `.mvn/maven.config`)
That makes something complicated to achieve and maintain (as Maven3 and Maven4 might have differences) for minimal gain. 
Here, we provide a pom snippet to copy/paste. We can upgrade the pom snippet if we need to update the recommended versions.

olamy:
oh crap.....
I pushed my mistake the branch to what is called origin repo for me (https://github.com/jenkinsci/jenkins-test-harness)
Then I waned to delete but delete to my fork and not here...

olamy:
reopened here https://github.com/jenkinsci/jenkins-test-harness/pull/829

olamy:
reopened

olamy:
grrhhh, I restored the branch but this doesn't;t re open the PR

PR #829, opened 2024-09-05
[JEP-237] Add new method to RealJenkinsRule to start in FIPS mode

olamy:
### Submitter checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -225,8 +225,8 @@ THE SOFTWARE.
     </dependency>
     <dependency>
       <groupId>io.jenkins.test.fips</groupId>
-      <artifactId>lib-fips-bundle-test</artifactId>
-      <version>-rc16.e95456a_916d7</version>
+      <artifactId>fips-bundle-test</artifactId>
+      <version>23.v76d4fd57f5b_d</version>
     </dependency>
     <dependency>
       <groupId>org.junit.jupiter</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -321,32 +321,6 @@ THE SOFTWARE.
               <overWriteSnapshots>true</overWriteSnapshots>
             </configuration>
           </execution>
-          <execution>
-            <id>copy-fips-libs</id>
-            <goals>
-              <goal>copy</goal>
-            </goals>
-            <configuration>
-              <artifactItems>
-                <artifactItem>
-                  <groupId>org.bouncycastle</groupId>
-                  <artifactId>bc-fips</artifactId>
-                  <version>1.0.2.4</version>
-                </artifactItem>
-                <artifactItem>
-                  <groupId>org.bouncycastle</groupId>
-                  <artifactId>bcpkix-fips</artifactId>
-                  <version>1.0.7</version>
-                </artifactItem>
-                <artifactItem>
-                  <groupId>org.bouncycastle</groupId>
-                  <artifactId>bctls-fips</artifactId>
-                  <version>1.0.17</version>
-                </artifactItem>
-              </artifactItems>
-              <stripVersion>true</stripVersion>
-            </configuration>
-          </execution>
         </executions>
       </plugin>
       <plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -316,6 +316,32 @@ THE SOFTWARE.
               <overWriteSnapshots>true</overWriteSnapshots>
             </configuration>
           </execution>
+          <execution>
+            <id>copy-fips-libs</id>
+            <goals>
+              <goal>copy</goal>
+            </goals>
+            <configuration>
+              <artifactItems>
+                <artifactItem>
+                  <groupId>org.bouncycastle</groupId>
+                  <artifactId>bc-fips</artifactId>
+                  <version>1.0.2.4</version>
+                </artifactItem>
+                <artifactItem>
+                  <groupId>org.bouncycastle</groupId>
+                  <artifactId>bcpkix-fips</artifactId>
+                  <version>1.0.7</version>
+                </artifactItem>
+                <artifactItem>
+                  <groupId>org.bouncycastle</groupId>
+                  <artifactId>bctls-fips</artifactId>
+                  <version>1.0.17</version>
+                </artifactItem>
+              </artifactItems>
+              <stripVersion>true</stripVersion>
+            </configuration>
+          </execution>
         </executions>
       </plugin>
       <plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -224,6 +224,11 @@ THE SOFTWARE.
       </exclusions>
     </dependency>
     <dependency>
+      <groupId>io.jenkins.test.fips</groupId>
+      <artifactId>fips-bundle-test</artifactId>
+      <version>1.0.0-20240910.032433-1</version>
+    </dependency>
+    <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter-api</artifactId>
     </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -225,8 +225,8 @@ THE SOFTWARE.
     </dependency>
     <dependency>
       <groupId>io.jenkins.test.fips</groupId>
-      <artifactId>fips-bundle-test</artifactId>
-      <version>1.0.0-20240910.032433-1</version>
+      <artifactId>lib-fips-bundle-test</artifactId>
+      <version>-rc16.e95456a_916d7</version>
     </dependency>
     <dependency>
       <groupId>org.junit.jupiter</groupId>
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -498,13 +498,11 @@ public final class RealJenkinsRule implements TestRule {
                             fipsLibrariesPath.resolve("bctls-fips.jar").toFile(),
                             fipsLibrariesPath.resolve("bcpkix-fips.jar").toFile());
         try {
-            javaOptions("--add-exports", "java.base/sun.security.provider=ALL-UNNAMED",
-                            "-Dsecurity.overridePropertiesFile=true",
+            javaOptions(    "-Dsecurity.overridePropertiesFile=true",
                             "-Djava.security.properties=" + writeFIPSJavaSecurityFile().toUri(),
                             "-Dorg.bouncycastle.fips.approved_only=true",
                             "-Djavax.net.ssl.trustStoreType=PKCS12",
-                            "-Djenkins.security.FIPS140.COMPLIANCE=true",
-                            "-Dcom.redhat.fips=false");
+                            "-Djenkins.security.FIPS140.COMPLIANCE=true");
         } catch (IOException e) {
             throw new RuntimeException(e);
         }
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -497,10 +497,10 @@ public final class RealJenkinsRule implements TestRule {
                             fipsLibrariesPath.resolve("bcpkix-fips.jar").toFile());
         try {
             // please note == is not a typo, but it makes our file completely override the jvm security file
-            javaOptions(    "-Djava.security.properties==" + writeFIPSJavaSecurityFile().toUri(),
-                            "-Dorg.bouncycastle.fips.approved_only=true",
-                            "-Djavax.net.ssl.trustStoreType=PKCS12",
-                            "-Djenkins.security.FIPS140.COMPLIANCE=true");
+            javaOptions("-Djava.security.properties==" + writeFIPSJavaSecurityFile().toUri(),
+                    "-Dorg.bouncycastle.fips.approved_only=true",
+                    "-Djavax.net.ssl.trustStoreType=PKCS12",
+                    "-Djenkins.security.FIPS140.COMPLIANCE=true");
         } catch (IOException e) {
             throw new RuntimeException(e);
         }
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -873,9 +873,6 @@ public final class RealJenkinsRule implements TestRule {
                     + (debugPort > 0 ? ",address=" + httpListenAddress + ":" + debugPort : ""));
         }
         if(!bootClasspathFiles.isEmpty()) {
-            //String fileList = String.join(File.pathSeparator, bootClasspathFiles.stream()
-            //        .map(file -> FilenameUtils.separatorsToSystem(file.getAbsolutePath())).toList());
-
             String fileList = bootClasspathFiles.stream().map(File::getAbsolutePath).collect(Collectors.joining(File.pathSeparator));
             argv.add("-Xbootclasspath/a:" + fileList);
 
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -28,6 +28,7 @@ import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.ExtensionList;
+import hudson.Functions;
 import hudson.model.UnprotectedRootAction;
 import hudson.security.ACL;
 import hudson.security.ACLContext;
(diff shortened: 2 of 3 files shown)


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -457,6 +462,127 @@ public RealJenkinsRule prepareHomeLazily(boolean prepareHomeLazily) {
...
+     *   <li>-Dsecurity.overridePropertiesFile=true</li>
+     *   <li>-Djavax.net.ssl.trustStoreType=PKCS12</li>
+     *   <li>-Djenkins.security.FIPS140.COMPLIANCE=true</li>
+     *   <li>-Dcom.redhat.fips=false</li>
+     * </ul>
+     * @param fipsLibrariesPath path to directory with FIPS BouncyCastle jars
+     */
+    public RealJenkinsRule withFIPSEnabled(@NonNull Path fipsLibrariesPath) {
+        this.fipsLibrariesPath = Objects.requireNonNull(fipsLibrariesPath);
+        // check expected BouncyCastle fips jars are here
```
olamy, line 487: discussion started here https://github.com/jenkinsci/jenkins-test-harness/pull/825#discussion_r1739610865

pom.xml
```diff
@@ -316,6 +316,32 @@ THE SOFTWARE.
...
+            <id>copy-fips-libs</id>
+            <goals>
+              <goal>copy</goal>
+            </goals>
+            <configuration>
+              <artifactItems>
+                <artifactItem>
+                  <groupId>org.bouncycastle</groupId>
+                  <artifactId>bc-fips</artifactId>
+                  <version>1.0.2.4</version>
```
olamy, line 329: discussion from https://github.com/jenkinsci/jenkins-test-harness/pull/825#discussion_r1743366773

olamy: @jtnord do you mean when Jenkins will have a class called FIPS200? :)

jtnord: > you may want a TODO here, as this is not the latest version of bc-fips but for a valid reason.

the latest bc-fips 1.x is `1.0.2.5`  but it should not be updated as it is not yet certified.
there is also a bc-fips `2.0.0.0` which is certified but at FIPS 140-3 (rather than 140-2) which is the current target of the JEP.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jtnord: > [olamy](https://github.com/olamy) [5 days ago](https://github.com/jenkinsci/jenkins-test-harness/pull/825#discussion_r1739610865)
another option would have been to use maven-resolver API and download files for the user. But this would mean we force the jars version here.

> @[jtnord](https://github.com/jtnord) jtnord [yesterday](https://github.com/jenkinsci/jenkins-test-harness/pull/825#discussion_r1743379921)
the java security and version of the jars may well be intertwined, so I am not sure that forcing the versions here would be a bad thing.
Additionally with the current setup all users need to duplicate the BC FIPS libraries in their pom config?

> @[olamy](https://github.com/olamy) olamy [7 hours ago](https://github.com/jenkinsci/jenkins-test-harness/pull/825#discussion_r1744688493)
Yes but at least users have the freedom to eventually test some upgrades to new version (note it's in the plugin configuration so hopefully dependabot or other tools will not propose any upgrade)
The problem with using something to download dependencies (because dependencies may not be in the local repository) is that all the setup must be available here (such as the path to the used settings.xml, which can be different from ~/.m2/settings.xml), cli request properties, the content of .mvn/maven.config)
That makes something complicated to achieve and maintain (as Maven3 and Maven4 might have differences) for minimal gain.
Here, we provide a pom snippet to copy/paste. We can upgrade the pom snippet if we need to update the recommended versions.

jtnord: This would be a double edged sword.
On the one hande this may be nice for plugin authors, but on the other tools like the PCT would not be able to check for regressions in a sweeping manner by just bumping the test-harness.
Newer versions (e.g. the 2.x line) of BC-FIPS *may* (I have no idea) require different changes to the `java.security` file.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -457,6 +462,127 @@ public RealJenkinsRule prepareHomeLazily(boolean prepareHomeLazily) {
...
+                !Files.exists(this.fipsLibrariesPath.resolve("bcpkix-fips.jar"))) {
+            throw fipsIllegalSetupException(this.fipsLibrariesPath);
+        }
+
+        withBootClasspath(fipsLibrariesPath.resolve("bc-fips.jar").toFile(),
+                            fipsLibrariesPath.resolve("bctls-fips.jar").toFile(),
+                            fipsLibrariesPath.resolve("bcpkix-fips.jar").toFile());
+        try {
+            // please note == is not a typo, but it makes our file completely override the jvm security file
+            javaOptions(    "-Djava.security.properties==" + writeFIPSJavaSecurityFile().toUri(),
```
jtnord, line 500: the java security and version of the jars may well be intertwined, so I am not sure that forcing the versions here would be a bad thing.
Additionally with the current setup all users need to duplicate the BC FIPS libraries in their pom config?

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -457,6 +462,127 @@ public RealJenkinsRule prepareHomeLazily(boolean prepareHomeLazily) {
...
+        }
+        Path javaSecurity = Paths.get(javaHome, "conf", "security", "java.security");
+        Properties properties = new Properties();
+        Path securityFile = Files.createTempFile("java", ".security");
+        securityFile.toFile().deleteOnExit();
+        try (InputStream inputStream = Files.newInputStream(javaSecurity);
+             OutputStream outputStream = Files.newOutputStream(securityFile)) {
+            properties.load(inputStream);
+            properties.keySet().removeIf(o -> ((String)o).startsWith("security.provider"));
+            properties.put("security.provider.1", "org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider C:HYBRID;ENABLE{All};");
```
jtnord, line 571: Note for others, the hybrid is a psedo random number that is initially seeded from a strong instance and is regularly reseeded from that instance to maintain strength and prevent any predictability.
This should ensure we do not run out of entropy on VMs, however as it does need the initial seed and some subsequent reseeding that is not gauranteed.  (Depending on how often the seeds secure random's source is required.)
in testing a product deployed with this I have not noticed any delays, but if we start seeing timeouts and the SR in the stack trace we would need the infra team to deploy something like [`haveged`](https://linux.die.net/man/8/haveged) on the nodes (both the VM and k8s node pool nodes).  (/cc @dduportal)

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleFIPSTest.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Arrays;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+
+public class RealJenkinsRuleFIPSTest {
+
+    @Rule public RealJenkinsRule rr = new RealJenkinsRule().prepareHomeLazily(true)
+            .withDebugPort(4001).withDebugServer(false)
```
jtnord, line 43: is this some left over testing?

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jtnord, line 500: you have some funky indentation here!  (the following lines may also need adjusting if spotless complains
```suggestion
            javaOptions("-Djava.security.properties==" + writeFIPSJavaSecurityFile().toUri(),
```

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -457,6 +462,127 @@ public RealJenkinsRule prepareHomeLazily(boolean prepareHomeLazily) {
...
+
+    private Path writeFIPSJavaSecurityFile() throws IOException {
+        String javaHome = System.getProperty("java.home");
+        if(javaHome == null) {
+            throw new IllegalArgumentException("Cannot find java.home property");
+        }
+        Path javaSecurity = Paths.get(javaHome, "conf", "security", "java.security");
+        Properties properties = new Properties();
+        Path securityFile = Files.createTempFile("java", ".security");
+        securityFile.toFile().deleteOnExit();
```
jtnord, line 566: NIT: There is actually a `TemporaryDirectoryAllocator tmp`  (although it is currently only cleans up if home is `non-null` but that could be changed?)

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
olamy: fixed.
but frankly would be better to have spotless activated (I will propose it once this one is merged)
we have plenty more other interesting things to do in our life than discussing about spaces :)

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleFIPSTest.java
olamy: comment from https://github.com/jenkinsci/jenkins-test-harness/pull/825#discussion_r1743517662 

sort of but for me this doesn't have any effect except running the test using -Dmaven.surefire.debug or debug in your IDE, at least you want to debug everything is ready and no need to change code for it.
Convenient as if you want to debug a RealJenkinsRule you really want to debug the server

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
olamy: Yes it's not perfect but at least users have the freedom to eventually test some upgrades to new version (note it's in the plugin configuration so hopefully dependabot or other tools will not propose any upgrade)
The problem with using something (such maven-resolver) to download dependencies (because dependencies may not be in the local repository) is that all the setup must be available here (such as the path to the used settings.xml, which can be different from ~/.m2/settings.xml), cli request properties, the content of .mvn/maven.config)
That makes something complicated (almost impossible as we may not have the information to mimic all the maven configuration) to achieve and maintain (as Maven3 and Maven4 might have differences) for minimal gain.
Here, we provide a pom snippet to copy/paste. We can upgrade the pom snippet to update the recommended versions.
Another solution, coulld to be bundle the fips jars within an artifact.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
olamy: > This would be a double edged sword. On the one hande this may be nice for plugin authors, but on the other tools like the PCT would not be able to check for regressions in a sweeping manner by just bumping the test-harness. Newer versions (e.g. the 2.x line) of BC-FIPS _may_ (I have no idea) require different changes to the `java.security` file.

at least it works for the current version we marked in core as named `FIPS140`
the release here are quick and cheap  why trying to solve we still have no ideas if it will be a problem or not.

jtnord: my concern is that pushing this down to plugins adds technical debt.  (e.g. rather than just being able to bump a jenkins-test-harness dep (possibly via a pom update) you end up needing to change the libraries and the test-harness in lock-step and once that is used in numerous plugins is more work that we would want it to be. (and changing how the test harness gets the libraries would similarly be more work that we would want it to be).

So it was a question of "are we definiately going to have this issue where a plugin can not just bump the bcfips libraries for testing because it also needs changes to the `java.security` file (in which case the way we have it today would be wrong and imply that is could indeed work.

I managed to track down the 2.0.0.0 [porting guide](https://downloads.bouncycastle.org/fips-java/docs/BC-FJA%202.0.0%20Porting%20Guide.pdf) and the [user guide](https://downloads.bouncycastle.org/fips-java/docs/BC-FJA-UserGuide-2.0.0.pdf) so I think this concern does not matter, witht he exception that for 2.0 a new library is needed in addition to the existing set.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
dduportal: It's worth adding it (I'll open an issue on jenkins-infra/helpdesk) right now on the agent templates (in https://github.com/jenkins-infra/packer-images/blob/main/provisioning/ubuntu-provision.sh). For VM agents, the service will be started on boot of each agent to help.
We already have the RNGD tools on the non-agent VMs (controllers, web services): https://github.com/jenkins-infra/jenkins-infra/blob/production/dist/profile/manifests/rngd.pp.

For the k8s node pool I have no idea if it is already present or not.
- We are using AKS with managed nodes with Azure Linux OS on the node: https://github.com/jenkins-infra/azure/blob/47e44a47bcd5330dbdffc443eb4b2f5a0c4b398a/ci.jenkins.io-kubernetes-agents.tf#L102-L134
- There is an (old: 2018) issue for AKS about this: https://github.com/Azure/AKS/issues/219 but it's unclear if it was fixed or not.
- On the upstream we have this issue: https://github.com/kubernetes/kubernetes/issues/60751 but again unclear

=> I believe we might need to search for a daemonset-based installation to ensure it's loaded.

dduportal: @jtnord is the "entropy increase" need to be set up (somehow) on Windows agents? (I do not know this platform well enough)

jtnord: > @jtnord is the "entropy increase" need to be set up (somehow) on Windows agents? (I do not know this platform well enough)

JDK on windows uses [`CryptGenRandom`](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptgenrandom) despite it being deprecated 😱 
the documentation here does talk about the sources and this includes process related things as well as keyboard/mouse.

The new alternative (AFAICT) is [`ProcessPrng`](https://learn.microsoft.com/en-us/windows/win32/seccng/processprng) which is lacking any and all details.
That said windows requirements these days are for TPM which contains a teue random number generator, so I don't think either of these would block like /dev/random.  (and yes there are various TPMs that exists whose TRNG is not actually random!).

also here is a [random Register article](https://www.theregister.com/2007/11/13/windows_random_number_gen_flawed/) saying windows is flawed.  
(which also implies at least in the windows 2000 era, that this was a PRNG not a TRNG...)

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
olamy: Perfectly understandable.
What about having another artifact bundling all together the fips bc libraries we know working together for a version x of the specifications? And so we can reuse that here eventually (unpacking them for users of the testing tool) having some constants/enum which can be used to generate the content of the security file?

jtnord: > so I think this concern does not matter, with the exception that for 2.0 a new library is needed in addition to the existing set.

which is actually a concern, so not sure why I wrote that!

> What about having another artifact bundling all together the fips bc libraries we know working together for a version x of the specifications?

I think there are a few things that we want to think about here.

1) how can we update the libraries to test a plugin automatically (e.g. during a PCT execution of a plugin (this could be bumping a J-T-H version which has the newer versions, but newer j-t-h's are not always backwards compatable).
2) how can we change what libraries are used (e.g. add the new library needed for BC-FIPS 2.0.0.0)
3) how can we make it easy for plugins to get the "right combination" (if plugins are to do this not the j-t-h)
4) how can we do this without introducing debt we need to maintain.

I think 2 can be solved by just looking for *all* libraries in a well known location during the RealJenkinsRule startup.
(this punts the need to provide this to the plugin so violates 3)

olamy: Right another solution. https://github.com/olamy/fips-bundle-test.
The generated jars will contain known working/certified/validated/right/etc.... combinations of BC-FIPS jars and generated `security.properties` according to those jars.
As explained, this used version can be overridden by an environment variable or system property; the PCT hook can certainly do this.
This makes this very simple to use (no need to add some complex pom fragments) 
If we need to add BC-FIPS 2.0.0.0, we simply need to add this new version to the bundle jar.

jtnord: that looks like a good idea.
I am also wondering if we can use the same method for the acceptance-test-harness to configure a JVM in a FIPSlike way.

olamy: sure but only for WinstoneController :) (we can forget JBossController and TomcatController)

olamy:
This PR now have dependency to another project: https://github.com/olamy/fips-bundle-test
hosting request: https://github.com/jenkins-infra/repository-permissions-updater/issues/4082

PR #835, opened 2024-09-18
`TailLog` to support reopened files

jglick:
I have some `RealJenkinsRule` tests which involve recreation of the build `log` file, but the default implementation of `Tailer` seems to be bound to an open file handle and thus does not notice any new content. Opening the file each time is of course less efficient but this utility is typically used only on very small log files in otherwise expensive tests so it should not matter. (Extracted from #827.)

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/TailLog.java
+++ b/src/main/java/org/jvnet/hudson/test/TailLog.java
@@ -29,8 +29,13 @@ import hudson.console.PlainTextConsoleOutputStream;
 import hudson.model.Job;
 import hudson.model.Run;
 import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.IOException;
 import java.io.PrintStream;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.attribute.FileTime;
+import java.time.Duration;
 import java.util.concurrent.Semaphore;
 import org.apache.commons.io.input.Tailer;
 import org.apache.commons.io.input.TailerListenerAdapter;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jvnet/hudson/test/TailLog.java
```diff
@@ -94,7 +99,45 @@ public TailLog withColor(PrefixedOutputStream.AnsiColor color) {
...
+                    long ptr;
+                    @Override public long getPointer() throws IOException {
+                        return ptr;
+                    }
+                    @Override public void seek(long pos) throws IOException {
+                        ptr = pos;
+                    }
+                    @Override public int read(byte[] b) throws IOException {
+                        // Unlike RandomAccessFileBridge, not sensitive to file handle:
+                        try (var is = Files.newInputStream(log)) {
```
Vlatombe, line 128: Is this correct? If the log file gets recreated, shouldn't `ptr` be reset ?

jglick: No, because `ptr` refers to how much content has already been printed; we do not want to restart at the top of the log file. See the associated test case, which prints 0–19 in order and then finishes.

Vlatombe: Ok so even if the file is recreated, you are sure it contains all the previous content already read.

jglick: Yes that is the scenario being tolerated here, since this utility is designed to work on build logs, which are append-only, never rotated.

Vlatombe (approved):

PR #838, opened 2024-09-18
Improved use of `java.io.tmpdir`

jglick:
Ensure that temporary files are written to `target/tmp/` (where `mvn clean` can delete them) rather than the system temporary directory, including in `RealJenkinsRule` forked JVMs; note that in the test JVM it seems to be necessary to look up this system property explicitly, I think since Surefire sets the system property after startup rather than using `-D`, and something reads and caches the value soon after startup. Ensure that all temporary directories are deleted when the test exits. Use temporary directories instead of `$JENKINS_HOME` for anything added by the test harness rather than Jenkins itself. (Extracted from #827.)


src/main/java/org/jvnet/hudson/test/recipes/WithPluginManager.java
```diff
@@ -90,7 +90,7 @@ public void setup(JenkinsRule jenkinsRule, WithPluginManager recipe) throws Exce
         public void decorateHome(JenkinsRule jenkinsRule, File home) throws Exception {
             Class<? extends PluginManager> c = recipe.value();
             Constructor<? extends PluginManager> ctr = c.getDeclaredConstructor(File.class);
-            jenkinsRule.setPluginManager(ctr.newInstance(home));
+            jenkinsRule.setPluginManager(ctr.newInstance(new File(home, "plugins")));
```
jglick, line 93: To restore behavior for those tests requesting `UnitTestSupportingPluginManager`.

src/main/java/org/jvnet/hudson/test/JavaNetReverseProxy2.java
```diff
@@ -90,7 +101,6 @@ private String getMimeType(String path) {
      */
     public static synchronized JavaNetReverseProxy2 getInstance() throws Exception {
         if (INSTANCE == null) {
-            // TODO: think of a better location --- ideally inside the target/ dir so that clean would wipe them out
```
jglick, line 93: Already worked. https://github.com/jenkinsci/plugin-pom/blob/e807519c3869c298fca47578e1f38c2bdc3cc735/pom.xml#L155-L156 https://github.com/jenkinsci/plugin-pom/blob/e807519c3869c298fca47578e1f38c2bdc3cc735/pom.xml#L806

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -504,27 +506,31 @@ public static List<String> getJacocoAgentOptions() {
         return new Statement() {
             @Override public void evaluate() throws Throwable {
                 System.out.println("=== Starting " + description);
-                if (war == null) {
-                    war = findJenkinsWar();
-                }
-                if (home.get() != null) {
+                try {
+                    jenkinsOptions("--webroot=" + createTempDirectory("webroot"), "--pluginroot=" + createTempDirectory("pluginroot"));
```
jglick, line 510: Various controller packagings set these options to temp dirs, so I think it is realistic to set them here automatically. Otherwise the fallback is inside `$JENKINS_HOME` which is undesirable.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -504,27 +506,31 @@ public static List<String> getJacocoAgentOptions() {
...
-                    home.set(tmp.allocate());
-                    if (!prepareHomeLazily) {
-                        provision();
-                    }
-                    base.evaluate();
                 } finally {
-                    stopJenkins();
                     try {
-                        deprovision();
+                        tmp.dispose();
```
jglick, line 533: May be confusing depending on whether or not you have whitespace displayed: now just deleting all temp dirs when the test completes, regardless of #610 mode. (The other code in `deprovision` did not seem important since a rule instance is only expected to be used once.)

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
```diff
@@ -440,6 +442,11 @@ public boolean isAlive(String name) {
             }
         }
         procs.clear();
+        try {
+            FileUtils.deleteDirectory(Path.of(System.getProperty("java.io.tmpdir"), "agent-work-dirs").toFile());
```
Vlatombe, line 446: Extract to constant?

Vlatombe (approved):

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
jglick: Actually when I look at this, it could be deleting work dirs still in use by other rules or even other tests, which is wrong.

Vlatombe (approved):

basil:
```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertTrue(Assert.java:53)
	at hudson.PluginManagerTest.uberClassLoaderIsAvailableDuringStart(PluginManagerTest.java:213)
```

```
org.jvnet.hudson.test.RealJenkinsRule$StepException: 
Remote step threw an exception: java.lang.AssertionError: 
Expected: is "/home/jenkins/agent/workspace/Core_jenkins_PR-9756/test/target/j h7251048736527506496/agent.jar"
     but: was "/home/jenkins/agent/workspace/Core_jenkins_PR-9756/test/target/agent.jar"
	at org.jvnet.hudson.test.RealJenkinsRule.runRemotely(RealJenkinsRule.java:998)
	at org.jvnet.hudson.test.RealJenkinsRule.runRemotely(RealJenkinsRule.java:983)
	at hudson.slaves.JNLPLauncherRealTest.then(JNLPLauncherRealTest.java:81)
	at hudson.slaves.JNLPLauncherRealTest.webSocket(JNLPLauncherRealTest.java:70)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
	at org.jvnet.hudson.test.RealJenkinsRule$1.evaluate(RealJenkinsRule.java:527)
Caused by: java.lang.AssertionError: 
Expected: is "/home/jenkins/agent/workspace/Core_jenkins_PR-9756/test/target/j h7251048736527506496/agent.jar"
     but: was "/home/jenkins/agent/workspace/Core_jenkins_PR-9756/test/target/agent.jar"
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
	at hudson.slaves.JNLPLauncherRealTest$RunJobStep.run(JNLPLauncherRealTest.java:107)
	at org.jvnet.hudson.test.RealJenkinsRule$StepsToStep2.run(RealJenkinsRule.java:1419)
	at org.jvnet.hudson.test.RealJenkinsRule$Endpoint.lambda$doStep$0(RealJenkinsRule.java:1332)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
```diff
@@ -465,7 +481,7 @@ private static AgentArguments getAgentArguments(JenkinsRule r, String name) thro
         if (!launcher.getWorkDirSettings().isDisabled()) {
             commandLineArgs = launcher.getWorkDirSettings().toCommandLineArgs(c);
         }
-        File agentJar = new File(r.jenkins.getRootDir(), "agent.jar");
+        File agentJar = new File(System.getProperty("java.io.tmpdir"), "agent.jar");
```
jtnord, line 484: this appears to completely break some things:

e.g. `-DforkCount=<number_greater_than_one>` as the file may not exist for 2 forks ([which use the same surefire temp director](https://github.com/jenkinsci/plugin-pom/blob/55de1268f3ffe29a657dc7b233c676697f2a1c4d/pom.xml#L806)y) , is completely written by one and the other and then used whilst the other is overwriting it.

2. the ability to quickly switch between version using `-DJenkins.version=x` (you now need to `mvn clean` as opposed to just using the incremental compilation which is faster)

jglick: Yeah a temp file would be better for this case. I can prepare a patch for it.

PR #843, opened 2024-09-20
`InboundAgentRule` should use a temp file for `agent.jar`

jglick:
https://github.com/jenkinsci/jenkins-test-harness/pull/838#discussion_r1768928055

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
+++ b/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
@@ -88,6 +88,7 @@ public final class InboundAgentRule extends ExternalResource {
     private final String id = UUID.randomUUID().toString();
     private final Map<String, Process> procs = Collections.synchronizedMap(new HashMap<>());
     private final Set<String> workDirs = Collections.synchronizedSet(new HashSet<>());
+    private final Set<File> jars = Collections.synchronizedSet(new HashSet<>());
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
```diff
@@ -481,10 +494,8 @@ private static AgentArguments getAgentArguments(JenkinsRule r, String name) thro
         if (!launcher.getWorkDirSettings().isDisabled()) {
             commandLineArgs = launcher.getWorkDirSettings().toCommandLineArgs(c);
         }
-        File agentJar = new File(System.getProperty("java.io.tmpdir"), "agent.jar");
-        if (!agentJar.isFile()) {
-            FileUtils.copyURLToFile(new Slave.JnlpJar("agent.jar").getURL(), agentJar);
-        }
+        File agentJar = Files.createTempFile(Path.of(System.getProperty("java.io.tmpdir")), "agent", ".jar").toFile();
+        FileUtils.copyURLToFile(new Slave.JnlpJar("agent.jar").getURL(), agentJar);
```
jtnord, line 498: I'm guessing that this is to protect against a single Test having multiple threads that call start() on 2 different agents sumultaneously to avoid any synchronization around the file.

jtnord (approved):

Sarrac3873 (approved):

PR #858, opened 2024-10-04
Offer a way to start the RealJenkinsRule instance with https only

Vlatombe:
This allows to start `RealJenkinsRule` instances listening on https instead of http.
A self-signed certificate matching the given host name (or localhost if unspecified) is automatically generated (using bouncycastle)

Various methods are offered to allow callers to access the RealJenkinsRule clients to access the instance without disabling security.

Also added utilities related to key store and self-signed certificate management.

### Submitter checklist


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -211,6 +216,8 @@ public final class RealJenkinsRule implements TestRule {
     private static final Pattern SNAPSHOT_INDEX_JELLY = Pattern.compile("(file:/.+/target)/classes/index.jelly");
 
     private final PrefixedOutputStream.Builder prefixedOutputStreamBuilder = PrefixedOutputStream.builder();
+    private boolean https;
+    private SSLContext sslContext;
```
jglick, line 220: Could just make `sslContext` be `@CheckForNull` and omit `https` as implied by a non-null value.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -728,7 +735,16 @@ public URL getUrl() throws MalformedURLException {
         if (port == 0) {
             throw new IllegalStateException("This method must be called after calling #startJenkins.");
         }
-        return new URL("http", host, port, "/jenkins/");
+        return new URL(https ? "https" : "http", host, port, "/jenkins/");
+    }
+
+    public RealJenkinsRule https(File keyStoreFile, String keyStorePassword, SSLContext sslContext) {
```
jglick, line 741: I do not think we need to deal with keystore passwords, do we? Presuming they are just test stores anyway.

Also it would be better to let the rule generate a keystore on the fly.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -891,6 +907,15 @@ public void startJenkins() throws Throwable {
         addTimeout();
     }
 
+    private Collection<String> getPortOptions() {
+        // initially port=0. On subsequent runs, the port is set to the port used allocated randomly on the first run.);
```
jglick, line 911:
```suggestion
        // Initially port=0. On subsequent runs, this is set to the port allocated randomly on the first run.
```

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -1021,6 +1046,13 @@ public <T extends Serializable> T runRemotely(Step2<T> s) throws Throwable {
         }
     }
 
+    private HttpURLConnection decorateConnection(@NonNull URLConnection urlConnection) {
+        if (https && sslContext != null) {
+            ((HttpsURLConnection) urlConnection).setSSLSocketFactory(sslContext.getSocketFactory());
```
jglick, line 1051: If this is all we are using, does it make more sense to configure an optional https://docs.oracle.com/en/java/javase/17/docs/api/java.base/javax/net/ssl/SSLSocketFactory.html ?

pom.xml
```diff
@@ -96,6 +96,11 @@ THE SOFTWARE.
         <artifactId>access-modifier-annotation</artifactId>
         <version>1.34</version>
       </dependency>
+      <dependency>
+        <groupId>org.bouncycastle</groupId>
+        <artifactId>bcpkix-jdk18on</artifactId>
+        <version>1.78.1</version>
+      </dependency>
```
jglick, line 103: Does not need to be in `dependencyManagement`; you can just specify the `version` directly in `dependencies`.

pom.xml
```diff
@@ -212,6 +217,10 @@ THE SOFTWARE.
       <artifactId>jmh-generator-annprocess</artifactId>
       <version>${jmh.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.bouncycastle</groupId>
+      <artifactId>bcpkix-jdk18on</artifactId>
```
jglick, line 222: My concern is that this will somehow clash with the BC pulled in via https://plugins.jenkins.io/bouncycastle-api/ in various plugins. Not sure how the dependency tree is resolved—could the JTH version take precedence, thus causing compilation errors or (worse) subtle behavioral differences in `JenkinsRule`-based tests?

src/main/java/jenkins/test/https/KeyStoreManager.java
```diff
@@ -0,0 +1,261 @@
...
+    private final String password;
+    @NonNull
+    private final KeyStore keyStore;
+    @NonNull
+    private final String type;
+
+    /**
+     * Creates a new instance using the default keystore type.
+     * @param path path of the keystore file. If it exists, it will be loaded automatically.
+     * @param password password for the keystore file.
```
jglick, line 68: Can this be optional, since for this kind of test we do not care about keystore passwords?

src/main/java/jenkins/test/https/SelfSignedCertificates.java
```diff
@@ -0,0 +1,167 @@
...
+import org.bouncycastle.operator.ContentSigner;
+import org.bouncycastle.operator.OperatorCreationException;
+import org.bouncycastle.operator.jcajce.JcaContentSignerBuilder;
+
+/**
+ * A utility class to generate self-signed certificates.
+ * <ul>
+ *   <li>A root CA
+ *   <li>An intermediate CA signed by the root CA
+ * </ul>
```
jglick, line 61: Better to use `@param`.

pom.xml
jtnord: especially as the upper-bounds rule in plugins does *not* look at test scope deps (which j-t-h will be) but the PCT does and upgrades test scope deps to be aligned.  (the dependency on jupiter-api iun this plugin already causes an issue)

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -771,6 +877,17 @@ private static File findJenkinsWar() throws Exception {
         return WarExploder.findJenkinsWar();
     }
 
+    /**
+     * Create a client configured to trust the self-signed certificate used by this instance.
```
jglick, line 881:
```suggestion
     * Create a client configured to trust any self-signed certificate used by this instance.
```
(since it also works when HTTPS is not enabled)

jtnord:
> A self-signed certificate matching the given host name (or localhost if unspecified) is automatically generated (using bouncycastle)

is there a reason you need to access the RealJenkinsRule via the host name?  if testing multiple RealJenkinsRules side by side and you want a nice jenkins-1 using some form of nip.io then would it not behove us to add a static SSL cert generated once and checked in (and skip generating on demand) that has subject altanmes for 
`localhost`, `::1`, `127.0.0.1`, `*.nip.io`, `*.localhost.me` ..,.. (insert any other providers here).

src/main/java/jenkins/test/https/SelfSignedCertificates.java
```diff
@@ -0,0 +1,167 @@
...
+                    Date.from(LocalDate.now()
+                            .plusDays(1)
+                            .atStartOfDay(ZoneId.systemDefault())
+                            .toInstant()),
+                    new X500Name("CN=endUserCert"),
+                    keyPair.getPublic());
+            builder.addExtension(Extension.keyUsage, true, new KeyUsage(KeyUsage.digitalSignature));
+            builder.addExtension(Extension.basicConstraints, false, new BasicConstraints(false));
+            var altNames = List.of(new GeneralName(GeneralName.dNSName, dnsName));
+            var subjectAltNames = GeneralNames.getInstance(new DERSequence(altNames.toArray(new GeneralName[0])));
```
jglick, line 150: or just
```suggestion
            var altName = new GeneralName(GeneralName.dNSName, dnsName);
            var subjectAltNames = GeneralNames.getInstance(new DERSequence(altName));
```

src/main/java/jenkins/test/https/SelfSignedCertificates.java
```diff
@@ -0,0 +1,167 @@
...
+ * <ul>
+ *   <li>A root CA
+ *   <li>An intermediate CA signed by the root CA
+ * </ul>
+ * Then you can generate user certificates for a given DNS name signed by the intermediate CA.
+ *
+ * @see #createRootCAs()
+ * @see #createUserCert(String, CertificateKeyPair)
+ */
+public record SelfSignedCertificates(CertificateKeyPair root, CertificateKeyPair intermediate) {
```
jtnord, line 67: YAGNI? create a single cert with subject alt-names and check it in.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -307,8 +330,13 @@ public RealJenkinsRule withTimeout(int timeout) {
      * might accidentally be shared across otherwise distinct services.
      * <p>Calling this method does <em>not</em> change the fact that Jenkins will be configured to listen only on localhost for security reasons
      * (so others in the same network cannot access your system under test, especially if it lacks authentication).
+     * <p>
+     * When using HTTPS, use @{link #https(String)} instead.
      */
     public RealJenkinsRule withHost(String host) {
+        if (https) {
+            throw new IllegalStateException("Don't call this method when using HTTPS");
+        }
```
jtnord, line 339: so why go to all the lengths to generate an SSL cert if it is always going to be localhost?

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -728,7 +756,85 @@ public URL getUrl() throws MalformedURLException {
...
+    public RealJenkinsRule https() {
+        return https("localhost");
+    }
+
+
+    /**
+     * Sets up HTTPS for the current instance, and disables plain HTTP.
+     * This generates a self-signed certificate for the given host name. The corresponding root CA that needs to be trusted by HTTP client can be obtained using {@link #getRootCA()}.
+     * <p>
+     * You don't need to call {@link #withHost(String)} when calling this method.
```
jtnord, line 778: I think you should limit this to be either localhost or something matching `*.<list_of_domains>` and skip the SSL Cert creation at runtime, and ditch the bouncycastle dependency.

jglick: Perhaps. Though we have had trouble in the past with test certs expiring and no one noticing until it became a nuisance.

pom.xml
jglick: (checking effects in https://github.com/jenkinsci/bom/pull/3720)

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jtnord: and if someone has a need for something that can not be solved like that - it would still seemingly need to be fixed so that tests would work relaible across different systems, so there could always be an option to have `withCertificate(Path cert)` and push the issue down to the plugin that is insisting on using some whacky name..

pom.xml
jtnord: or just use the java api from `sun.security.x509` and add `Add-Exports: java.base/sun.security.x509` to the librarieys manifiest?

src/main/java/jenkins/test/https/KeyStoreManager.java
Vlatombe: Not for now; https://github.com/jenkinsci/winstone/pull/417

Vlatombe:
> is there a reason you need to access the RealJenkinsRule via the host name?

The point here is to test close to real life scenarios:
* a reverse proxy set up with https (public-jenkins.acme.com)
* a RealJenkinsRule instance set up with https (internal-jenkins.acme.com)

Each one using different https certificates (so external users must trust CA for the reverse proxy, then reverse proxy must trust CA from RJR instance).

Checking in a static certificate would do if I had only one RealJenkinsRule instance, but I want guarantee that trust stores are setup appropriately so for test purpose, I think it is better to have a single certificate generated for each instance.

pom.xml
Vlatombe: > `sun.security.x509`

Looks bad, as this could break anytime when upgrading JDK, whereas we can control BC version (and at least it is a proper API). See also https://bugs.openjdk.org/browse/JDK-8058778, closed as won't fix.

jtnord: >whereas we can control BC version

but leak it to plugins test classpaths over what would or would not come from bouncycastle-api and will get automatically updated by PCT to a potentially different version than provided by the bouncycastle-api plugin.

jtnord:
> * a reverse proxy set up with https (public-jenkins.acme.com)
> * a RealJenkinsRule instance set up with https (internal-jenkins.acme.com)

A very specific setup useful for a very narrow set of plugins (lets call it one).
checkin the 3 certs in that project and expose a method to use a specific cert for a specific RealJenkinsRule instance then, and let the default checked in cert here work all the other plugins that just want a valid TLS endpoint.

pom.xml
jglick: An alternate would be to have JTH itself only support running with a specified cert; but move the code to generate certs on demand to `bouncycastle-api-plugin/src/test/java/` with

```xml
<no-test-jar>false</no-test-jar>
```

so that plugins interested in this could share an implementation.

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleHttpsTest.java
```diff
@@ -0,0 +1,46 @@
...
+        var options = InboundAgentRule.Options
+                .newBuilder()
+                .name("remote")
+                .webSocket()
+                .color(PrefixedOutputStream.Color.YELLOW);
+        iar.createAgent(rr, options.build());
+    }
+
+    private static void log(JenkinsRule r) throws IOException {
+        LOGGER.info("Running on " + r.getURL().toExternalForm());
```
jglick, line 44: Could strengthen this slightly to return the URL and show that it is `https`.

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleHttpsTest.java
```diff
@@ -0,0 +1,46 @@
...
+    @Before
+    public void setUp() throws Throwable {
+        rr.startJenkins();
+    }
+
+    @Test
+    public void runningStepAndUsingHtmlUnit() throws Throwable {
+        // We can run steps
+        rr.runRemotely(RealJenkinsRuleHttpsTest::log);
+        // replica1 directly
```
jglick, line 27:
```suggestion
        // web client trusts the cert
```

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleHttpsTest.java
```diff
@@ -0,0 +1,46 @@
...
+            wc.getPage(rr.getUrl());
+        }
+    }
+
+    @Test
+    public void inboundAgent() throws Throwable {
+        var options = InboundAgentRule.Options
+                .newBuilder()
+                .name("remote")
+                .webSocket()
```
jglick, line 38: I guess either way this exercises the changes, since even a TCP agent would make an initial HTTP handshake by default. (`InboundAgentRule` does not support direct mode AFAIK.)

src/main/keystore/generate_self_signed_cert.sh
```diff
@@ -0,0 +1,47 @@
...
+emailAddress			= noreply@jenkins.io
+
+[ req_ext ]
+keyUsage				= digitalSignature
+basicConstraints		= CA:false
+subjectAltName 			= @alternate_names
+subjectKeyIdentifier 	= hash
+
+[ alternate_names ]
+DNS.1         = localhost
```
jglick, line 32: I thought we would include at least `*.localtest.me`?

src/main/keystore/generate_self_signed_cert.sh
```diff
@@ -0,0 +1,47 @@
...
+EOF
+
+mkdir -p "$tmpDir/output"
+key="$tmpDir/output/key.pem"
+cert="$tmpDir/output/cert.pem"
+certP12="$tmpDir/output/cert.p12"
+# Generate a self-signed cert valid for 100 years
```
jglick, line 39:
```suggestion
```
(redundant)

pom.xml
```diff
@@ -220,7 +220,7 @@ THE SOFTWARE.
     <dependency>
       <groupId>org.jenkins-ci.modules</groupId>
       <artifactId>instance-identity</artifactId>
-      <version>3.1</version>
+      <version>201.vd2a_b_5a_468a_a_6</version>
```
jglick, line 223: Is this still relevant, given that we are testing WS agents?

src/main/java/jenkins/test/https/KeyStoreManager.java
```diff
@@ -0,0 +1,286 @@
...
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.TrustManagerFactory;
+import javax.net.ssl.X509TrustManager;
+import org.htmlunit.WebClient;
+
+/**
+ * Allows to manage a java keystore file more easily than base JDK.
+ */
+public class KeyStoreManager {
```
jglick, line 56:
```suggestion
public final class KeyStoreManager {
```

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -307,8 +329,13 @@ public RealJenkinsRule withTimeout(int timeout) {
      * might accidentally be shared across otherwise distinct services.
      * <p>Calling this method does <em>not</em> change the fact that Jenkins will be configured to listen only on localhost for security reasons
      * (so others in the same network cannot access your system under test, especially if it lacks authentication).
+     * <p>
+     * When using HTTPS, use @{link #https(String)} instead.
```
jglick, line 333:
```suggestion
     * When using HTTPS, use {@link #https(String)} instead.
```

jglick (approved):
Looks good AFAICT

pom.xml
Vlatombe: I was attempting to align bc dependencies (was bringing `jdk15on` flavor, not `jdk18on`) at the time. I will revert.

src/main/keystore/generate_self_signed_cert.sh
Vlatombe: Not needed for now.

pom.xml
Vlatombe, line 223:
```suggestion
      <version>3.1</version>
```

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleHttpsTest.java
Vlatombe: Prints something like
```
[INFO] [stdout] 2024-10-10 12:53:16.687+0000 [id=94]    INFO    o.j.h.t.RealJenkinsRuleHttpsTest#log: Running on https://localhost:54912/jenkins/
```

jglick: Yes, I just meant that the test JVM could assert that.

jglick (approved):

jglick (approved):

jtnord (approved):

PR #868, opened 2024-10-30
add checks for inline javascript

mawinter69:
Detects usage of inline event handlers like onclick, usage of script elements (except for json type), usage of checkUrl without checkDependsOn

Same as https://github.com/jenkins-infra/repository-permissions-updater/pull/4049 but for maven

### Testing done
Manually tested that a plugin that contains any of the offending patterns properly fails the tests.

### Submitter checklist

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JellyTestSuiteBuilder.java
+++ b/src/main/java/org/jvnet/hudson/test/JellyTestSuiteBuilder.java
@@ -26,8 +26,11 @@ package org.jvnet.hudson.test;
 import java.io.File;
 import java.io.IOException;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.jar.JarEntry;
@@ -36,7 +39,11 @@ import junit.framework.TestCase;
 import junit.framework.TestResult;
 import junit.framework.TestSuite;
 import org.apache.commons.io.FileUtils;
+import org.dom4j.Attribute;
 import org.dom4j.Document;
+import org.dom4j.Element;
+import org.dom4j.Namespace;
+import org.dom4j.Node;
 import org.dom4j.ProcessingInstruction;
 import org.dom4j.io.SAXReader;
 import org.jvnet.hudson.test.junit.GroupedTest;
@@ -88,6 +95,7 @@ public class JellyTestSuiteBuilder {
         private final URL jelly;
         private final JellyClassLoaderTearOff jct;
         private final boolean requirePI;
+        private List<String> errors = new ArrayList<>();
 
         JellyCheck(URL jelly, String name, JellyClassLoaderTearOff jct, boolean requirePI) {
             super(name);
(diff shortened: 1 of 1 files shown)


mawinter69:
Should this be guarded by a switch so that the checks can be disabled (like the requirePI switch)?
Would we need some way to handle false positives (not sure if there could be any), e.g. by adding some comment or attribute?
Enabling that feature as is would probably break the build for a lot of plugins, but it would force plugin owners to react when they want to upgrade

timja:
> Should this be guarded by a switch so that the checks can be disabled (like the requirePI switch)? Would we need some way to handle false positives (not sure if there could be any), e.g. by adding some comment or attribute? Enabling that feature as is would probably break the build for a lot of plugins, but it would force plugin owners to react when they want to upgrade

Best to run it through bom by force updating the test harness version in pct.sh (similar to how this updated the hpi plugin):
https://github.com/jenkinsci/bom/blob/81216a901a1846f7c78f88dd4f7e46021c1a4403/pct.sh#L37

To see what the impact is

src/main/java/org/jvnet/hudson/test/JellyTestSuiteBuilder.java
```diff
@@ -103,11 +111,52 @@ protected void runTest() throws Exception {
...
+            allNodes.forEach(n -> {
+                Element element = (Element) n;
+                Attribute onclick = element.attribute("onclick");
+                Attribute checkUrl = element.attribute("checkUrl");
+                Attribute checkDependsOn = element.attribute("checkDependsOn");
+                if (checkUrl != null  && checkDependsOn == null) {
+                    errors.add("Usage of 'checkUrl' without 'checkDependsOn' in "+jelly);
+                }
+                if (onclick != null && element.getNamespace() != Namespace.NO_NAMESPACE) {
+                    errors.add("Usage of 'onclick' from a taglib in "+jelly);
```
timja, line 137: could we link to a page explaining how to fix this?

src/main/java/org/jvnet/hudson/test/JellyTestSuiteBuilder.java
```diff
@@ -103,11 +111,52 @@ protected void runTest() throws Exception {
...
+
+        private void checkJavaScriptAttributes(Document dom) {
+            List<Node> allNodes = dom.selectNodes("//*");
+            allNodes.forEach(n -> {
+                Element element = (Element) n;
+                Attribute onclick = element.attribute("onclick");
+                Attribute checkUrl = element.attribute("checkUrl");
+                Attribute checkDependsOn = element.attribute("checkDependsOn");
+                if (checkUrl != null  && checkDependsOn == null) {
+                    errors.add("Usage of 'checkUrl' without 'checkDependsOn' in "+jelly);
```
timja, line 134: I'm wondering if this is clear enough to people, e.g. in most cases people need to add a `checkDependsOn=""`.

Personally I think this isn't clear enough on its own.

(It would be good if this could be fixed so it doesn't need the empty checkDependsOn)

timja:
Could you add sample output to the description showing what it looks like to a developer please?

src/main/java/org/jvnet/hudson/test/JellyTestSuiteBuilder.java
mawinter69: When used via a taglib it mean the resulting tag will also have that onclick handler. Afaik for all places in core that define an `onclick` in a taglib it is marked as deprecated. So the usage of onclick from a taglib is not really different from an onclick on a plain html element.

src/main/java/org/jvnet/hudson/test/JellyTestSuiteBuilder.java
mawinter69: Probably when we have CSP enabled by default and all plugins are adjusted we could remove an empty `checkDependsOn`, until then we need it to find out if we have to run the `eval` on the `checkUrl`.
We could add a link to https://www.jenkins.io/doc/developer/security/csp/ whenever there is an inline javascript found.

mawinter69:
> > Should this be guarded by a switch so that the checks can be disabled (like the requirePI switch)? Would we need some way to handle false positives (not sure if there could be any), e.g. by adding some comment or attribute? Enabling that feature as is would probably break the build for a lot of plugins, but it would force plugin owners to react when they want to upgrade
> 
> Best to run it through bom by force updating the test harness version in pct.sh (similar to how this updated the hpi plugin): https://github.com/jenkinsci/bom/blob/81216a901a1846f7c78f88dd4f7e46021c1a4403/pct.sh#L37
> 
> To see what the impact is

That'll be tricky. The project here is already requiring 2.479 of core. So when I want to build a plugin I need the plugin parent pom in version 5.x at least. Without that it's not building

timja:
> > > Should this be guarded by a switch so that the checks can be disabled (like the requirePI switch)? Would we need some way to handle false positives (not sure if there could be any), e.g. by adding some comment or attribute? Enabling that feature as is would probably break the build for a lot of plugins, but it would force plugin owners to react when they want to upgrade
> > 
> > 
> > Best to run it through bom by force updating the test harness version in pct.sh (similar to how this updated the hpi plugin): https://github.com/jenkinsci/bom/blob/81216a901a1846f7c78f88dd4f7e46021c1a4403/pct.sh#L37
> > To see what the impact is
> 
> That'll be tricky. The project here is already requiring 2.479 of core. So when I want to build a plugin I need the plugin parent pom in version 5.x at least. Without that it's not building

Could do a backport PR?

timja:
But i guess that will cause problems for ones that have upgraded

mawinter69:
> > > > Should this be guarded by a switch so that the checks can be disabled (like the requirePI switch)? Would we need some way to handle false positives (not sure if there could be any), e.g. by adding some comment or attribute? Enabling that feature as is would probably break the build for a lot of plugins, but it would force plugin owners to react when they want to upgrade
> > > 
> > > 
> > > Best to run it through bom by force updating the test harness version in pct.sh (similar to how this updated the hpi plugin): [jenkinsci/bom@`81216a9`/pct.sh#L37](https://github.com/jenkinsci/bom/blob/81216a901a1846f7c78f88dd4f7e46021c1a4403/pct.sh#L37)
> > > To see what the impact is
> > 
> > 
> > That'll be tricky. The project here is already requiring 2.479 of core. So when I want to build a plugin I need the plugin parent pom in version 5.x at least. Without that it's not building
> 
> Probably worth another go? Hopefully most plugins are on 5.x now.

I did a test with https://github.com/jenkinsci/bom/pull/5988 already. Need to repeat this now due to changes here and in hpi plugin

mawinter69:
With the updated pct the bom looks good https://ci.jenkins.io/job/Tools/job/bom/job/PR-5988/7/
The one failure is due to duplicate keys in a properties file so not related.

PR #874, opened 2024-11-14
Tolerate missing `DEBUG_YUI` field

basil:
Gracefully deal with the case where `DEBUG_YUI` is not defined in core.

### Testing done

Ran a test successfully both with and without `DEBUG_YUI` defined in core.


MarkEWaite (approved):
Thanks!

PR #878, opened 2024-11-25
Provide Hamcrest matchers as alternatives to built-in JenkinsRule run assertions

Vlatombe:
Plays nicely with Awaitability as well as in combination with other matchers.

### Submitter checklist

### DIFF
--- a/src/main/java/jenkins/test/RunMatchers.java
+++ b/src/main/java/jenkins/test/RunMatchers.java
@@ -60,6 +60,13 @@ public final class RunMatchers {
         return new RunLogMatcher(message);
     }
 
+    /**
+     * Creates a matcher checking whether a build has completed.
+     */
+    public static Matcher<Run<?,?>> completed() {
+        return new CompletedRunMatcher();
+    }
+
     private static class RunResultMatcher extends TypeSafeMatcher<Run<?,?>> {
         @NonNull
         private final Result expectedResult;
@@ -114,6 +121,18 @@ public final class RunMatchers {
         @Override
         public void describeTo(Description description) {
             description.appendText("log containing ").appendValue(message);
+        }
+    }
+
+    private static class CompletedRunMatcher extends TypeSafeMatcher<Run<?, ?>> {
+        @Override
+        protected boolean matchesSafely(Run<?, ?> run) {
+            return !run.isLogUpdated();
+        }
+
+        @Override
+        public void describeTo(Description description) {
+            description.appendText("a completed build");
         }
     }
 }
--- a/src/main/java/jenkins/test/RunMatchers.java
+++ b/src/main/java/jenkins/test/RunMatchers.java
@@ -1,0 +1,119 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2024 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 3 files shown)


timja (approved):
❤️

src/main/java/jenkins/test/RunMatchers.java
```diff
@@ -0,0 +1,119 @@
...
+            return run.getResult() == expectedResult;
+        }
+
+        @Override
+        protected void describeMismatchSafely(Run<?, ?> item, Description mismatchDescription) {
+            mismatchDescription.appendText("was ").appendValue(item.getResult());
+        }
+    }
+
+    private static class RunLogMatcher extends TypeSafeMatcher<Run<?, ?>> {
```
jglick, line 87: What about `waitForMessage`? This has special behavior for a completed build; not sure if that can be carried over to Awaitility.

src/test/java/jenkins/test/RunMatchersTest.java
```diff
@@ -0,0 +1,71 @@
...
+public class RunMatchersTest {
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+
+    @Test
+    public void buildSuccessful() throws Exception {
+        var p = j.createFreeStyleProject();
+        p.getBuildersList().add(new SleepBuilder(1000));
+        var b = p.scheduleBuild2(0).waitForStart();
+        assertThat(j.waitForCompletion(b), isSuccessful());
```
jglick, line 52: :thinking: `waitForCompletion` really belongs here too, so that we could write e.g.

```java
await().until(b, allOf(completed(), isSuccessful());
```

or

```java
assertThat(await().until(b, completed()), isSuccessful());
```

jglick (approved):

src/main/java/jenkins/test/RunMatchers.java
Vlatombe: It has a fail-early mechanism in case a build is complete, but I don't see how to replicate that using Awaitibility and matchers.

src/test/java/jenkins/test/RunMatchersTest.java
Vlatombe: Added `completed()`

src/main/java/jenkins/test/RunMatchers.java
Vlatombe: I guess something like
```
await().until(b, completed());
assertThat(b, logContains("foo"));
```

would be equivalent to the current `waitForMessage`

src/main/java/jenkins/test/RunMatchers.java
```diff
@@ -116,4 +123,16 @@ public void describeTo(Description description) {
             description.appendText("log containing ").appendValue(message);
         }
     }
+
+    private static class CompletedRunMatcher extends TypeSafeMatcher<Run<?, ?>> {
+        @Override
+        protected boolean matchesSafely(Run<?, ?> run) {
+            return !run.isLogUpdated();
```
jglick, line 130: So I think https://github.com/jenkinsci/jenkins-test-harness/blob/f2a982bb910fa3fc2998586276a58e5042530e1d/src/main/java/org/jvnet/hudson/test/JenkinsRule.java#L1632 could be deleted and that method could also suggest using this matcher.

src/main/java/jenkins/test/RunMatchers.java
jglick: No, because `waitForMessage` can be used to wait for a line in a running build.

I was just wondering if there was something in Awaitility that a matcher can use to indicate that not only does the condition not match now, we can be sure it will not match later. https://javadoc.io/static/org.awaitility/awaitility/4.2.2/org/awaitility/core/FailFastCondition.html maybe? No Javadoc :cry:

Vlatombe: Ah, https://github.com/awaitility/awaitility/wiki/Usage#fail-fast-conditions

Vlatombe: So something like

```
await().failFast(() -> !b.isLogUpdated()).until(() -> b, logContains("foo"));
```
?

Vlatombe:
Relaunching CI

PR #881, opened 2024-12-02
`XStreamSerializable`

jglick:
Occasionally a `RealJenkinsRule`-based test utility needs to pass and/or return a structured object which is not `Serializable`. This is currently awkward; you can try to move the code to create/assert the object inside the Jenkins JVM, or find some ad-hoc way of passing it. Since XStream does not require objects to implement `Serializable`, and many types used in Jenkins are safe to round-trip via XML but are not marked `Serializable`, it is convenient to be able to work with them directly. (I considered just relaxing the signatures of the various `runRemotely` overloads to tolerate any `Object`, and magically wrap them if not already `Serializable`, but this seemed complicated and prone to abuse. It is better to be conscious of the fact that you are serializing objects _somehow_.)

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -745,7 +745,8 @@ public final class RealJenkinsRule implements TestRule {
     /**
      * One step to run.
      * <p>Since this thunk will be sent to a different JVM, it must be serializable.
-     * The test class will certainly not be serializable, so you cannot use an anonymous inner class.
+     * The test class will certainly not be serializable, so you cannot use an anonymous inner class,
+     * and regular lambdas also risk accidentally capturing non-serializable objects from scope.
      * The friendliest idiom is a static method reference:
      * <pre>
      * &#64;Test public void stuff() throws Throwable {
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -158,10 +158,10 @@ import org.kohsuke.stapler.verb.POST;
  * <p>Known limitations:
  * <ul>
  * <li>Execution is a bit slower due to the overhead of launching a new JVM; and class loading overhead cannot be shared between test cases. More memory is needed.
- * <li>Remote thunks must be serializable. If they need data from the test JVM, you will need to create a {@code static} nested class to package that.
+ * <li>Remote calls must be serializable. Use methods like {@link #runRemotely(RealJenkinsRule.StepWithReturnAndOneArg, Serializable)} and/or {@link XStreamSerializable} as needed.
  * <li>{@code static} state cannot be shared between the top-level test code and test bodies (though the compiler will not catch this mistake).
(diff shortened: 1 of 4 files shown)


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -745,7 +745,8 @@ public String[] getTruststoreJavaOptions() {
     /**
      * One step to run.
      * <p>Since this thunk will be sent to a different JVM, it must be serializable.
-     * The test class will certainly not be serializable, so you cannot use an anonymous inner class.
+     * The test class will certainly not be serializable, so you cannot use an anonymous inner class,
+     * and regular lambdas also risk accidentally capturing non-serializable objects from scope.
```
jglick, line 749: Admittedly you will find that out at runtime with a `NotSerializableException`. My bigger concern is that the serial form of a general lambda (not a method handle) might not carry over from the test JVM to the Jenkins JVM, especially if in the future we support running the Jenkins JVM in a container that might be using a distinct version of Java. Maybe I should stop worrying and trust that https://docs.oracle.com/javase/8/docs/api/serialized-form.html#java.lang.invoke.SerializedLambda will be reliable? The bytecode sent to the Jenkins JVM is guaranteed to be identical to that run in the test JVM, so there is no concern about choice of compiler, version skew as test code is edited over time, etc.; [this post](https://cr.openjdk.org/~briangoetz/eg-attachments/lambda-serialization.html) implies that there is no risk for a given bytecode version.

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
```diff
@@ -389,4 +393,20 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListen
         p.scheduleBuild2(0).waitForStart();
     }
 
+    @Test public void xStreamSerializable() throws Throwable {
+        rr.startJenkins();
+        // Neither ParametersDefinitionProperty nor ParametersAction could be passed directly.
+        // (In this case, ParameterDefinition and ParameterValue could have been used raw.
+        // But even List<ParameterValue> cannot be typed here, only e.g. ArrayList<ParameterValue>.)
+        var a = rr.runRemotely(RealJenkinsRuleTest::_xStreamSerializable, new XStreamSerializable<>(new ParametersDefinitionProperty(new StringParameterDefinition("X", "dflt"))));
```
Vlatombe, line 401: Maybe as static factory to avoid the syntax overhead.

```suggestion
        var a = rr.runRemotely(RealJenkinsRuleTest::_xStreamSerializable, XStreamSerializable.of(new ParametersDefinitionProperty(new StringParameterDefinition("X", "dflt"))));
```

Vlatombe (approved):

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
jglick: Had considered it. I guess it is slightly simpler that way…

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
jglick: ⇒ #882

PR #882, opened 2024-12-02
Advertise lambdas as arguments to `then` / `runRemotely`

jglick:
More concise alternative to use of static method handles, at least for the common case that there is only one call site (or the lambda delegates to a static method with arguments). https://github.com/jenkinsci/jenkins-test-harness/pull/881#discussion_r1866482652 for discussion (see https://github.com/jenkinsci/jenkins-test-harness/pull/882/commits/1f09a1983f262c1e5442a9e62167323a8d8ce600)

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -759,7 +759,7 @@ public final class RealJenkinsRule implements TestRule {
      * try {@link #runRemotely(Step2)} or {@link #runRemotely(StepWithReturnAndOneArg, Serializable)} etc.
      * (using {@link XStreamSerializable} as needed).
      * <p>
-     * Alternately, you could use a lambda (taking care not to capture non-serializable objects from scope):
+     * Alternately, you could use a lambda:
      * <pre>
      * &#64;Test public void stuff() throws Throwable {
      *     rr.then(r -> {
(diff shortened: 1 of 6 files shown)


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -759,6 +758,15 @@ public String[] getTruststoreJavaOptions() {
      * If you need to pass and/or return values, you can still use a static method reference:
      * try {@link #runRemotely(Step2)} or {@link #runRemotely(StepWithReturnAndOneArg, Serializable)} etc.
      * (using {@link XStreamSerializable} as needed).
+     * <p>
+     * Alternately, you could use a lambda (taking care not to capture non-serializable objects from scope):
```
Vlatombe, line 762: In my experience, this is pretty difficult to assess once you experience an issue.

jglick: Could be. Noted in https://github.com/jenkinsci/jenkins-test-harness/pull/881#discussion_r1866482652. The difference is that (anonymous) inner classes will _always_ capture variables in lexical scope, whereas lambdas usually will not. Noted another caveat in b9e40f53ae5faca982a045f7b23a7825ed228ffb.

amuniz (approved):

Vlatombe (approved):

PR #897, opened 2025-01-17
Add LogRecorder utility

strangelookingnerd:
When working on https://github.com/jenkinsci/configuration-as-code-plugin/pull/2623 I was looking for a way to make use of `LoggerRule` without depending on JUnit 4. 
My workaround https://github.com/jenkinsci/configuration-as-code-plugin/pull/2623#discussion_r1918338722 does the trick but I think it should be optimized.

This PR extracts the logic from `LoggerRule` and provides a _JUnit-version-agnostic_ implementation that is more versatile.

I was thinking of adding a JUnit5 extension that is similar to `LoggerRule` however I do not really see the point. 
The only thing that `@Rule public LoggerRule rule = new LoggerRule();` does is to have an automated clean-up in the `TestCase#after`. The way to do this way sleeker _in my opinion_ is to make use of `AutoClosable` like so:

```java
@Test
void myTestCase() {
  // do stuff that is required for my test...
  try (LogRecorder recorder = new LogRecorder().record("Foo", Level.INFO).capture(100)) {
    // do the stuff I want to record...
    assertThat(recorder, LogRecorder.recorded(equalTo("MyMessage")));
  }
}
```

See `LogRecorderTest` for more examples of this.

Ultimately this new utility enables more tests to migrate away from JUnit4.
     
### Testing done

Clean build with `mvn clean verify`

### Submitter checklist

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/LogRecorder.java
+++ b/src/main/java/org/jvnet/hudson/test/LogRecorder.java
@@ -1,0 +1,326 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2016 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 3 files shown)


src/main/java/org/jvnet/hudson/test/LogRecorder.java
```diff
@@ -0,0 +1,326 @@
...
+import java.util.stream.Collectors;
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeMatcher;
+
+/**
+ * A test utility which allows you to easily enable one or more loggers for the duration of a test.
+ * Call {@link #record(Class, Level)} or another overload for the recording to take effect.
+ * <p>By default, messages are merely printed to test output.
+ * If you also want to examine them, call {@link #capture}.
```
timja, line 50: Could you see if possible to add example code in the javadoc?

```java
 try (LogRecorder recorder = new LogRecorder().record("Foo", Level.INFO).capture(100)) {
   FOO_LOGGER.log(Level.INFO, "Entry 2");
   assertThat(logRecorder, LogRecorder.recorded(equalTo("Entry 2")));
}
```

src/test/java/org/jvnet/hudson/test/LogRecorderTest.java
```diff
@@ -0,0 +1,153 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2017 CloudBees, Inc.
```
timja, line 4:
```suggestion
 * Copyright 2025 Jenkins project contributors
```

src/main/java/org/jvnet/hudson/test/LogRecorder.java
```diff
@@ -0,0 +1,335 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2016 CloudBees, Inc.
```
timja, line 4:
```suggestion
 * Copyright 2025 Jenkins project contributors
```

timja (approved):

daniel-beck:
Would be useful if `LoggerRule` had an `@see` or similar in Javadoc when trying to use it and being blocked, and/or deprecation of `LoggerRule` in favor of the new one.

PR #900, opened 2025-01-20
Enable using `@LocalData` for JUnit5 tests

strangelookingnerd:
Right now there seems to be no way to make JUnit5 tests annotated with `@LocalData` to resolve the local data in a convenient way.
This PR adds support by adapting the recipe resolution in `org.jvnet.hudson.test.junit.jupiter.JUnit5JenkinsRule#recipe` as well as the test method resolution in `org.jvnet.hudson.test.recipes.LocalData.RuleRunnerImpl#setup`.

This change enables more tests to migrate to JUnit5

### Testing done

Added 4 new tests that demonstrate the correct resolution of `@LocalData` in JUnit5 tests,

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #907, opened 2025-01-23
Fixes `@LocalData` when being used on a test method in superclass

Vlatombe:
By switching to `getDeclaredMethod` in #900, this broke this use case that is currently exercised in email-ext plugin
(`ScriptContentSecureTest>ScriptContentTest#testTemplateShouldBeLoadedFromTheClosestExistingFolderConfigInTheHierarchyUpToGlobalConfig`)

The provided test breaks without this patch.

### Submitter checklist
- [X] Link to relevant issues in GitHub or Jira
- [X] Link to relevant pull requests, esp. upstream and downstream changes

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/recipes/LocalData.java
+++ b/src/main/java/org/jvnet/hudson/test/recipes/LocalData.java
@@ -102,7 +102,7 @@ public @interface LocalData {
             Method testMethod;
 
             try {
-                testMethod = desc.getTestClass().getDeclaredMethod(desc.getMethodName());
+                testMethod = desc.getTestClass().getMethod(desc.getMethodName());
             } catch (NoSuchMethodException ex) {
                 testMethod = desc.getTestClass().getDeclaredMethod(desc.getMethodName(), JenkinsRule.class);
             }
--- a/src/test/java/org/jvnet/hudson/test/recipes/LocalDataExtendedTest.java
+++ b/src/test/java/org/jvnet/hudson/test/recipes/LocalDataExtendedTest.java
@@ -1,0 +1,8 @@
+package org.jvnet.hudson.test.recipes;
+
+
+/**
+ * Illustrates a case where a test class extends another test class that has test methods using the LocalData recipe.
+ */
+public class LocalDataExtendedTest extends LocalDataZipTest {
+}


src/test/java/org/jvnet/hudson/test/recipes/LocalDataExtendedTest.java
```diff
@@ -0,0 +1,8 @@
+package org.jvnet.hudson.test.recipes;
+
+
+/**
+ * Illustrates a case where a test class extends another test class that has test methods using the LocalData recipe.
+ */
+public class LocalDataExtendedTest extends LocalDataZipTest {
```
timja, line 7: just to confirm this is intended not to have any test inside it and it fails without the change in `src/main`?

Vlatombe: Yes, the purpose is for it to pick up the test method defined in `LocalDataZipTest`.

timja (approved):

src/main/java/org/jvnet/hudson/test/recipes/LocalData.java
```diff
@@ -102,7 +102,7 @@ public void setup(JenkinsRule jenkinsRule, LocalData recipe) throws Exception {
             Method testMethod;
 
             try {
-                testMethod = desc.getTestClass().getDeclaredMethod(desc.getMethodName());
```
strangelookingnerd, line 105: This will not work for JUnit5 tests that do not have a JenkinsRule as parameter (which is a valid option) and are not public (which is also a valid option). 

One remedy would be to use both `getMethod` and `getDeclaredMethod`.

strangelookingnerd (changes requested):
@Vlatombe Sorry for the inconvenience, I was unaware of that use case being used. Please see my comment below.

src/main/java/org/jvnet/hudson/test/recipes/LocalData.java
Vlatombe: > that do not have a JenkinsRule as parameter (which is a valid option)

How is that valid? `@LocalData` is expected to work within a JenkinsRule instance, right?

strangelookingnerd: Yes it is, however you can do something like this:

```Java
@WithJenkins
class TestClass {

  private JenkinsRule rule;

  @BeforeEach
  void before(JenkinsRule rule) {
    this.rule = rule;
  }
  
  @LocalData
  @Test
  void testMethod() {
   // do stuff with rule
  }
}
```

PR #908, opened 2025-01-23
Use stderr rather than stdout for `RealJenkinsRule` & `TailLog`

jglick:
The `junit` plugin displays stdout & stderr separately, following the separated streams in Surefire `TEST-*.xml`. This is just confusing when you are trying to see the sequence of events: you need to mentally collate by timestamp to reconstruct the log you would see from `mvn test -Dtest=…` output (compare https://github.com/jenkinsci/plugin-pom/blob/8561e2e8c77c344d0465ec6c77e2e3eedc927c00/pom.xml#L1070-L1080) particularly if the output includes a mixture of `Logger` lines from the Surefire JVM and the Jenkins JVM.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -563,7 +563,7 @@ public final class RealJenkinsRule implements TestRule {
         this.description = description;
         return new Statement() {
             @Override public void evaluate() throws Throwable {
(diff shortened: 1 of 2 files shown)


timja (approved):

basil:
Required changes to `hudson.security.LDAPEmbeddedFIPSTest.testBlowsUpOnStart` in https://github.com/jenkinsci/ldap-plugin/pull/336.

PR #910, opened 2025-01-27
Raise default permissions of `JUnit5JenkinsRule`

### LINKED ISSUE #901: `JUnit5JenkinsRule` and `JenkinsRule` have different default permissions
I noticed some very odd behavior in https://github.com/jenkinsci/jira-plugin/pull/634 and https://github.com/jenkinsci/config-file-provider-plugin/pull/347 where `CredentialsStore#addCredentials` would suddenly throw `AccessDeniedException3 anonymous is missing the Credentials/Create permission` when invoked in a JUnit5 test case whereas there is no issue when doing the same with JUnit4.

### Reproduction steps

```java
public class ReproducerTest {

    @Rule
    public JenkinsRule junit4rule = new JenkinsRule();

    @org.junit.Test
    public void junit4() {
        MockAuthorizationStrategy strategy = new MockAuthorizationStrategy();
        junit4rule.getInstance().setAuthorizationStrategy(strategy);

        Assert.assertTrue(junit4rule.getInstance().getACL().hasPermission(Permission.READ));
        Assert.assertTrue(junit4rule.getInstance().getACL().hasPermission(Permission.WRITE));
        Assert.assertTrue(junit4rule.getInstance().getACL().hasPermission(Permission.CREATE));
        Assert.assertTrue(junit4rule.getInstance().getACL().hasPermission(Permission.UPDATE));
        Assert.assertTrue(junit4rule.getInstance().getACL().hasPermission(Permission.DELETE));
        Assert.assertTrue(junit4rule.getInstance().getACL().hasPermission(Permission.CONFIGURE));

    }

    @org.junit.jupiter.api.Test
    @WithJenkins
    void junit5(JenkinsRule junit5rule) {
        MockAuthorizationStrategy strategy = new MockAuthorizationStrategy();
        junit5rule.getInstance().setAuthorizationStrategy(strategy);

        Assertions.assertTrue(junit5rule.getInstance().getACL().hasPermission(Permission.READ));
        Assertions.assertTrue(junit5rule.getInstance().getACL().hasPermission(Permission.WRITE));
        Assertions.assertTrue(junit5rule.getInstance().getACL().hasPermission(Permission.CREATE));
        Assertions.assertTrue(junit5rule.getInstance().getACL().hasPermission(Permission.UPDATE));
        Assertions.assertTrue(junit5rule.getInstance().getACL().hasPermission(Permission.DELETE));
        Assertions.assertTrue(junit5rule.getInstance().getACL().hasPermission(Permission.CONFIGURE));
    }
}
```

### Expected Results

Permissions of JUnit4 `JenkinsRule` and `JUnit5JenkinsRule` should be identical.

### Actual Results

Permissions are different.

### Anything else?

I think I found a potential culprit in https://github.com/jenkinsci/jenkins-test-harness/blob/446201c06cb8935880d5b0ae71348345fda18ea1/src/main/java/org/jvnet/hudson/test/JenkinsRule.java#L657 which does not seem to be applied in the `JUnit5JenkinsRule` as well.
However I will need some more time to investigate this further.

### Are you interested in contributing a fix?

I would love to. However I'm unsure what the best solution would be:

#### Option 1

Remove permissions from `JenkinsRule` to align with `JUnit5JenkinsRule` -> Will likely cause a lot of tests to fail

#### Option 2

Add permissions to `JUnit5JenkinsRule` to align with `JenkinsRule` -> Will likely cause _some_ tests to fail

strangelookingnerd:
Fixes #901

This PR raises the default permissions of `JUnit5JenkinsRule` to align with their JUnit4 counterparts.
It is important to notice that this change is breaking and will cause _some_ JUnit5 based plugin tests to fail in case they rely on the fact that the default user / permission is `Unauthenticated`.

The alternative would be to leave the permissions as they are but make users aware of the differences in JUnit4 and JUnit5.

### Migration

JUnit5 based tests may fail due to this change (as they expect to run in an `Unauthenticated` context) - like for example:

```java
@Test
void testUnauthenticated(JenkinsRule r) throws Exception {
    DummySecurityRealm realm = r.createDummySecurityRealm();
    r.jenkins.setSecurityRealm(realm);

    MockAuthorizationStrategy strategy = new MockAuthorizationStrategy();
    r.jenkins.setAuthorizationStrategy(strategy);

    assertThrows(AccessDeniedException.class, () -> somethingUnauthenticatedUsersCanNotDo());
}
```

To restore the previous behavior tests need to explicitly impersonate an `ANONYMOUS` user like so:

```java
@Test
void testUnauthenticated(JenkinsRule r) throws Exception {
    DummySecurityRealm realm = r.createDummySecurityRealm();
    r.jenkins.setSecurityRealm(realm);

    MockAuthorizationStrategy strategy = new MockAuthorizationStrategy();
    r.jenkins.setAuthorizationStrategy(strategy);

    try (ACLContext ignored = ACL.as2(Jenkins.ANONYMOUS2)) {
        assertThrows(AccessDeniedException.class, () -> somethingUnauthenticatedUsersCanNotDo());
    }
}
```

### Testing done

Added `org.jvnet.hudson.test.JenkinsRulePermissionTest` to validate the behavior.

### Submitter checklist

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/junit/jupiter/JUnit5JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/junit/jupiter/JUnit5JenkinsRule.java
@@ -2,6 +2,8 @@ package org.jvnet.hudson.test.junit.jupiter;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.lang.reflect.Method;
+
+import hudson.security.ACL;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.ParameterContext;
 import org.junit.runner.Description;
@@ -26,6 +28,9 @@ class JUnit5JenkinsRule extends JenkinsRule {
 
     @Override
     public void recipe() throws Exception {
+        // so that test code has all the access to the system
+        ACL.as2(ACL.SYSTEM2);
+
         JenkinsRecipe jenkinsRecipe =
                 context.findAnnotation(JenkinsRecipe.class).orElse(null);
         if (jenkinsRecipe != null) {
--- a/src/test/java/org/jvnet/hudson/test/JenkinsRulePermissionTest.java
+++ b/src/test/java/org/jvnet/hudson/test/JenkinsRulePermissionTest.java
@@ -1,0 +1,54 @@
+package org.jvnet.hudson.test;
+
+import hudson.security.Permission;
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.jupiter.api.Assertions;
+import org.jvnet.hudson.test.junit.jupiter.WithJenkins;
+
+/**
+ * Validates that JUnit4 and JUnit5 based {@link JenkinsRule} implementations have the same default authorization and permissions.
+ */
+@SuppressWarnings("deprecation")
+public class JenkinsRulePermissionTest {
+
+    @Rule
+    public JenkinsRule junit4rule = new JenkinsRule();
+
+    @org.junit.Test
+    public void junit4() {
+        MockAuthorizationStrategy strategy = new MockAuthorizationStrategy();
+        junit4rule.getInstance().setAuthorizationStrategy(strategy);
+
+        // basic permissions
+        Assert.assertTrue(junit4rule.getInstance().getACL().hasPermission(Permission.READ));
+        Assert.assertTrue(junit4rule.getInstance().getACL().hasPermission(Permission.WRITE));
+        Assert.assertTrue(junit4rule.getInstance().getACL().hasPermission(Permission.CREATE));
+        Assert.assertTrue(junit4rule.getInstance().getACL().hasPermission(Permission.UPDATE));
+        Assert.assertTrue(junit4rule.getInstance().getACL().hasPermission(Permission.DELETE));
+        Assert.assertTrue(junit4rule.getInstance().getACL().hasPermission(Permission.CONFIGURE));
+
+        // admin permissions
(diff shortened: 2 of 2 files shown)


timja (commented):
Seems sensible, possible to evaluate impact? I guess this could be backported to the default version in PCT or BOM? Or are the numbers so low you could do a quick test?

strangelookingnerd:
While many builds in https://github.com/jenkinsci/bom/pull/4324 failed for reasons I can not really explain, there was only a single plugin that had test failures that are related to this change: https://ci.jenkins.io/job/Tools/job/bom/job/PR-4324/4/pipeline-console/?selected-node=9033

Since I'm the maintainer of that plugin this is not really a suprise to me 😄 
I guess the number of plugins that have already migrated to JUnit5 and relied on the current behavior of permissions is diminishing. With that being said, the impact of this change will likely not be too noticable.

timja:
> I can not really explain

I think to get them to pass you would need to backport this PR to the pre ee9 branch: https://github.com/jenkinsci/jenkins-test-harness/pull/906

I don't think JUnit 5 has been adopted very far so lets go for it, we can revert back if required.

timja (approved):

basil:
@strangelookingnerd In https://github.com/jenkinsci/bom/pull/4689 I am seeing PCT-only failures of `jenkins.plugins.foldericon.PermissionTest` which can be reproduced locally. I only see the failures when `custom-folder-icon` is compiled with its release version of JTH and when the tests are executed with this PR. If the tests are both compiled and executed with the JTH version from this PR, everything passes. Not sure why, but in the meantime can we get a new release of `custom-folder-icon` to work around this?

strangelookingnerd:
@basil Released the plugin as `2.19`

PR #920, opened 2025-02-18
Introduce `RealJenkinsRule.createSyntheticPlugin` to allow synthetic plugins to be installed dynamically and break compatibility for existing callers of `RealJenkinsRule.addSyntheticPlugin`

dwnusbaum:
Needed for https://github.com/jenkinsci/jenkins/pull/10240.

Backwards-incompatible for most existing callers of `RealJenkinsRule.addSyntheticPlugin`.

### Testing done

New automated test case added.

### Submitter checklist

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -30,6 +30,7 @@ import edu.umd.cs.findbugs.annotations.Nullable;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Extension;
 import hudson.ExtensionList;
+import hudson.PluginManager;
 import hudson.model.UnprotectedRootAction;
 import hudson.security.ACL;
 import hudson.security.ACLContext;
@@ -297,9 +298,22 @@ public final class RealJenkinsRule implements TestRule {
      * @return a builder
      */
     public SyntheticPlugin addSyntheticPlugin(Package pkg) {
-        SyntheticPlugin p = new SyntheticPlugin(pkg.getName());
+        SyntheticPlugin p = new SyntheticPlugin(pkg.getName(), this);
         syntheticPlugins.add(p);
         return p;
+    }
+
+    /**
+     * Creates a test-only plugin based on sources defined in this module, but does not install it.
+     * See {@link #addSyntheticPlugin} for more details. Prefer that method if you simply want the
+     * plugin to be installed automatically. Use {@link SyntheticPlugin#writeTo} to package the
+     * plugin, which can then be used as needed, for example by {@link PluginManager#dynamicLoad}.
+     * @see #addSyntheticPlugin
+     * @param pkg the Java package containing any classes and resources you want included
+     * @return a builder
+     */
+    public static SyntheticPlugin createSyntheticPlugin(Package pkg) {
+        return new SyntheticPlugin(pkg.getName(), null);
     }
 
     /**
@@ -1700,14 +1714,16 @@ public final class RealJenkinsRule implements TestRule {
      * Alternative to {@link #addPlugins} or {@link TestExtension} that lets you build a test-only plugin on the fly.
      * ({@link ExtensionList#add(Object)} can also be used for certain cases, but not if you need to define new types.)
      */
-    public final class SyntheticPlugin {
+    public static final class SyntheticPlugin {
         private final String pkg;
+        private final RealJenkinsRule rjr;
         private String shortName;
         private String version = "1-SNAPSHOT";
         private Map<String, String> headers = new HashMap<>();
 
-        SyntheticPlugin(String pkg) {
+        SyntheticPlugin(String pkg, RealJenkinsRule rjr) {
             this.pkg = pkg;
+            this.rjr = rjr;
             shortName = "synthetic-" + pkg.replace('.', '-');
         }
 
@@ -1748,10 +1764,13 @@ public final class RealJenkinsRule implements TestRule {
          * @return back to the rule builder
          */
         public RealJenkinsRule done() {
-            return RealJenkinsRule.this;
+            if (rjr == null) {
+                throw new IllegalStateException("May only be called after RealJenkinsRule.addSyntheticPlugin");
+            }
+            return rjr;
         }
 
-        void writeTo(File jpi, String defaultJenkinsVersion) throws IOException, URISyntaxException {
+        public void writeTo(File jpi, String defaultJenkinsVersion) throws IOException, URISyntaxException {
             var mani = new Manifest();
             var attr = mani.getMainAttributes();
             attr.put(Attributes.Name.MANIFEST_VERSION, "1.0");
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -30,7 +30,6 @@ import edu.umd.cs.findbugs.annotations.Nullable;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Extension;
 import hudson.ExtensionList;
-import hudson.PluginManager;
 import hudson.model.UnprotectedRootAction;
 import hudson.security.ACL;
 import hudson.security.ACLContext;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -297,11 +298,24 @@ public RealJenkinsRule addPlugins(String... plugins) {
...
+     * See {@link #addSyntheticPlugin} for more details. Prefer that method if you simply want the
+     * plugin to be installed automatically. Use {@link SyntheticPlugin#writeTo} to package the
+     * plugin, which can then be used as needed, for example by {@link PluginManager#dynamicLoad}.
+     * @see #addSyntheticPlugin
+     * @param pkg the Java package containing any classes and resources you want included
+     * @return a builder
+     */
+    public static SyntheticPlugin createSyntheticPlugin(Package pkg) {
+        return new SyntheticPlugin(pkg.getName(), null);
+    }
```
dwnusbaum, line 317: IDK what approach is best. It is somewhat awkward to preserve the `done` method for existing callers. I also considered moving this to `JenkinsRule` instead as a non-static method. I also thought about adding a `File SyntheticPlugin.create(JenkinsRule)` or `SyntheticPlugin.dynamicLoad(JenkinsRule)` method instead of exposing `writeTo(File, String)`.

If we do not care about compatibility I think it would be clearest to get rid of `SyntheticPlugin.done`, expose the constructor for `SyntheticPlugin`, change `SyntheticPlugin addSyntheticPlugin` to `RealJenkinsRule addSyntheticPlugin(SyntheticPlugin)`, and then  adding either `File JenkinsRule.createSyntheticPlugin(SyntheticPlugin)` or `void JenkinsRule.dynamicLoad(SyntheticPlugin)`.

jglick: (did you mean `s/JenkinsRule/RealJenkinsRule/g` ^^^?)

jglick: > If we do not care about compatibility…

So far `addSyntheticPlugin` is just used by one CloudBees CI plugin, so if it would be clearer to make an incompatible change I think that would be OK. We would just make a minor edit to those tests when bumping the JTH version.

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -1748,10 +1764,13 @@ public SyntheticPlugin header(String key, String value) {
...
         public RealJenkinsRule done() {
-            return RealJenkinsRule.this;
+            if (rjr == null) {
+                throw new IllegalStateException("May only be called after RealJenkinsRule.addSyntheticPlugin");
+            }
+            return rjr;
         }
 
-        void writeTo(File jpi, String defaultJenkinsVersion) throws IOException, URISyntaxException {
+        public void writeTo(File jpi, String defaultJenkinsVersion) throws IOException, URISyntaxException {
```
jglick, line 1773: Better introduce Javadoc especially for the nonobvious `defaultJenkinsVersion` param (which we should likely rename to just `jenkinsVersion`).

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleSyntheticPluginTest.java
```diff
@@ -58,4 +60,15 @@ private static void _smokes(JenkinsRule r) throws Throwable {
...
 
+    @Test public void dynamicLoad() throws Throwable {
+        var file = temp.newFile("synthetic-stuff.jpi");
+        rr.then(r -> {
+            RealJenkinsRule.createSyntheticPlugin(Stuff.class.getPackage())
+                    .shortName("synthetic-stuff")
+                    .writeTo(file, Jenkins.getVersion().toString());
+            r.jenkins.pluginManager.dynamicLoad(file);
+            assertThat(r.createWebClient().goTo("stuff", "text/plain").getWebResponse().getContentAsString(),
+            is(Jenkins.get().getLegacyInstanceId()));
```
jglick, line 71:
```suggestion
                is(Jenkins.get().getLegacyInstanceId()));
```
(enabling Spotless would be a good idea, the flux of PRs in this repo is just pretty high)

jglick (dismissed):
(pending discussion of `done`)

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
dwnusbaum: > (did you mean s/JenkinsRule/RealJenkinsRule/g ^^^?)

No. If you want to install the plugin dynamically, you need a `JenkinsRule`, and non-static methods on `RealJenkinsRule` cannot be used from the thunk sent to the controller. You can create the synthetic plugin via RJR outside of the scope of `then`/`runRemotely` and just write it to a temp file that gets sent to the controller for a call to `dynamicLoad`, but then you have to use `runRemotely` once first just to get the correct Jenkins version, which is awkward. Methods on `JenkinsRule` are able to call `Jenkins.getVersion()` (why is there not a non-static equivalent?) and use `JenkinsRule.env` for temp files.

Here is an example of what the alternate approaches I described would look like in the test here:

With `JenkinsRule.dynamicLoad`:

```java
    @Test public void dynamicLoad() throws Throwable {
        rr.then(r -> {
            r.dynamicLoad(new SyntheticPlugin(Stuff.class.getPackage()).shortName("synthetic-stuff"));
            assertThat(r.createWebClient().goTo("stuff", "text/plain").getWebResponse().getContentAsString(),
                is(Jenkins.get().getLegacyInstanceId()));
        });
    }
```

With `JenkinsRule.createSyntheticPlugin`:

```java
    @Test public void dynamicLoad() throws Throwable {
        rr.then(r -> {
            File pluginJpi = r.createSyntheticPlugin(new SyntheticPlugin(Stuff.class.getPackage()).shortName("synthetic-stuff"));
            r.jenkins.pluginManager.dynamicLoad(pluginJpi);
            assertThat(r.createWebClient().goTo("stuff", "text/plain").getWebResponse().getContentAsString(),
                is(Jenkins.get().getLegacyInstanceId()));
        });
    }
```

jglick: > If you want to install the plugin dynamically, you need a `JenkinsRule`

Why?

jglick: > You can create the synthetic plugin via RJR outside of the scope of `then`/`runRemotely` and just write it to a temp file that gets sent to the controller for a call to `dynamicLoad`

Right, that is what I was expecting here.

> then you have to use `runRemotely` once first just to get the correct Jenkins version, which is awkward

```java
var jenkinsVersion = rjr.call(r -> Jenkins.getVersion().toString());
```

is awkward?

Anyway we can compute the `targetJenkinsVersion` from the WAR without starting Jenkins, so we should just do that, without asking the caller to pass in a version. (If they specifically want to declare an older version, to test detached plugins, `header` already lets them do so.)

dwnusbaum: See [056e148](https://github.com/jenkinsci/jenkins-test-harness/pull/920/commits/056e148972f94c931cf409cfceb5bba1a9d77c03).

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
dwnusbaum: [056e148](https://github.com/jenkinsci/jenkins-test-harness/pull/920/commits/056e148972f94c931cf409cfceb5bba1a9d77c03)

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleSyntheticPluginTest.java
dwnusbaum: Indented in [056e148](https://github.com/jenkinsci/jenkins-test-harness/pull/920/commits/056e148972f94c931cf409cfceb5bba1a9d77c03)

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleSyntheticPluginTest.java
```diff
@@ -40,7 +41,7 @@ public final class RealJenkinsRuleSyntheticPluginTest {
     @Rule public RealJenkinsRule rr = new RealJenkinsRule().prepareHomeLazily(true);
 
     @Test public void smokes() throws Throwable {
-        rr.addSyntheticPlugin(Stuff.class.getPackage()).done();
+        rr.addSyntheticPlugin(new SyntheticPlugin(Stuff.class.getPackage()));
```
dwnusbaum, line 44: The change is no longer backwards compatible, but this is what needs to be done to adapt.

jglick (approved):

jglick:
(Merge at will, if the revised API is being used successfully downstream.)

PR #921, opened 2025-02-18
Let `SimpleCommandLauncher` kill the agent JVM more promptly when disconnected

jglick:
https://github.com/jenkinsci/support-core-plugin/pull/624#issuecomment-2666863457

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/SimpleCommandLauncher.java
+++ b/src/main/java/org/jvnet/hudson/test/SimpleCommandLauncher.java
@@ -86,7 +86,10 @@ public class SimpleCommandLauncher extends ComputerLauncher {
             	pb.environment().putAll(env);
             }
             final Process proc = pb.start();
-            ExtensionList.lookupSingleton(FasterKill.class).procs.put(computer, new WeakReference<>(proc));
+            var impls = ExtensionList.lookup(FasterKill.class);
+            if (!impls.isEmpty()) { // not available in RJR
+                impls.get(0).procs.put(computer, new WeakReference<>(proc));
+            }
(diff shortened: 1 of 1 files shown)


dwnusbaum (approved):
Seems fine. Do you think it is worth running through the PCT to check if it causes any problems for agent disconnection/restart related tests?

jglick:
> Do you think it is worth running through the PCT

That does sound prudent. I will have to look up how to test a JTH bump in PCT.

jglick:
Turns out it is pretty complex due to a history of core changes: https://github.com/jenkinsci/plugin-compat-tester/pull/746

timja:
> Turns out it is pretty complex due to a history of core changes: [jenkinsci/plugin-compat-tester#746](https://github.com/jenkinsci/plugin-compat-tester/pull/746)

yeah you need to backport this change to the pre jetty 12 line of the test harness and then make a PR to plugin-compat-tester updating the minimum versions of JTH, then PR to bom with the updated PCT version

jglick:
I think I am not going to bother with all that. Will try this in a couple of plugins, and if later it turns out to cause any issues, we can just revert/amend.

jglick:
Breaks `ExecutorStepTest.nodeDisconnectMissingContextVariableException` though I am not yet sure why. Also does not seem to suffice in every single case in `support-core` or in `workflow-cps`, though again I am unsure why. I was able to confirm that this does kill the agent process before `JenkinsRule` test cleanup begins, so what else is holding the file handle? Unfortunately Windows file lock error messages do not give details, and I have been unable to reproduce any of these failures locally.

jglick:
https://github.com/jenkinsci/support-core-plugin/pull/624 still not stable. I think I actually meant to use `afterDisconnect`.

PR #924, opened 2025-02-20
Add support for custom prefix

### LINKED ISSUE #917: Allow specifying custom context path in `RealJenkinsRule`
### What feature do you want to see added?

`RealJenkinsRule` always uses the `--prefix=/jenkins`, and trying to override it in `#jenkinsOptions(String)` doesn't seem to work.

Some tests may want to test with a different (typically empty) prefix. For `JenkinsRule`, this is possible, see e.g. https://github.com/jenkinsci/jenkins/blob/f294f55ea0e8becdd75e9b9ed5e9c15b8f372674/test/src/test/java/org/kohsuke/stapler/BindTest.java#L33-L40, but not for `RealJenkinsRule`.

### Upstream changes

_No response_

### Are you interested in contributing this feature?

_No response_

Vlatombe:
Fixes #917

### Submitter checklist

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -30,6 +30,7 @@ import edu.umd.cs.findbugs.annotations.Nullable;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Extension;
 import hudson.ExtensionList;
+import hudson.Util;
 import hudson.model.UnprotectedRootAction;
 import hudson.security.ACL;
 import hudson.security.ACLContext;
@@ -243,6 +244,8 @@ public final class RealJenkinsRule implements TestRule {
     private SSLSocketFactory sslSocketFactory;
     private X509Certificate rootCA;
 
+    private String prefix = "/jenkins";
+
     public RealJenkinsRule() {
         home = new AtomicReference<>();
     }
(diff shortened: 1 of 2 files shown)


daniel-beck (approved):
Looks good, thanks! (Untested.)

src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java
```diff
@@ -103,6 +105,16 @@ private static void _smokes(JenkinsRule r) throws Throwable {
...
     }
 
+    @Test public void customPrefix() throws Throwable {
+        rr.withPrefix("/foo").startJenkins();
+        assertThat(rr.getUrl().getPath(), equalTo("/foo/"));
+    }
+
+    @Test public void noPrefix() throws Throwable {
+        rr.noPrefix().startJenkins();
+        assertThat(rr.getUrl().getPath(), emptyString());
```
jglick, line 115: Should this not be `equalTo("/")`?

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -825,7 +859,7 @@ public URL getUrl() throws MalformedURLException {
         if (port == 0) {
             throw new IllegalStateException("This method must be called after calling #startJenkins.");
         }
-        return new URL(https ? "https" : "http", host, port, "/jenkins/");
+        return new URL(https ? "https" : "http", host, port, prefix == null ? "" : prefix + "/");
```
jglick, line 862:
```suggestion
        return new URL(https ? "https" : "http", host, port, prefix == null ? "/" : prefix + "/");
```

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -390,6 +393,37 @@ public RealJenkinsRule withHost(String host) {
...
+     * <p>
+     * By default, the prefix defaults to {@code /jenkins}.
+     * <p>
+     * Use {@link #noPrefix()} to remove the prefix, or provide another prefix of your choice as long as it starts with / and doesn't end with /.
+     */
+    public RealJenkinsRule withPrefix(@NonNull String prefix) {
+        var sanitized = Util.fixEmpty(prefix);
+        if (sanitized == null) {
+            throw new IllegalStateException("Use noPrefix() to remove the prefix.");
+        }
```
jglick, line 416: To my mind it would be clearer to delete `noPrefix` and just let `.withPrefix("")` mean to request `http://localhost:12345/` which would be consistent with `.withPrefix("/xxx")` meaning `http://localhost:12345/xxx/`. The root URL is always expected to end with a slash.

jglick (changes requested):
Inconsistent treatment.

jglick (approved):

PR #926, opened 2025-02-21
Deprecating `RestartableJenkinsRule`

jglick:
I found that some plugin developers were unaware that this class had long since been superseded by `JenkinsSessionRule`.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
@@ -1,7 +1,6 @@
 package org.jvnet.hudson.test;
 
 import groovy.lang.Closure;
-import hudson.PluginManager;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
@@ -21,6 +20,7 @@ import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import org.junit.After;
 import org.junit.Assert;
 import org.junit.rules.MethodRule;
 import org.junit.runner.Description;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
```diff
@@ -21,34 +20,26 @@
...
- * {@link #then} is called.
- * <p>
- * If your test requires disabling of a plugin then the default {@link PluginManager} ({@link TestPluginManager}) used for tests
- * will need to be changed to {@link UnitTestSupportingPluginManager}.
- * This can be accomplished by annotating the test with {@code @WithPluginManager(UnitTestSupportingPluginManager.class)}.
- * 
  * @author Kohsuke Kawaguchi
- * @see JenkinsRule
  * @since 1.567
+ * @deprecated New code should use {@link JenkinsSessionRule}.
```
timja, line 33: hmm shouldn't new code use `@WithJenkins` (JUnit 5)? with which you can use `.restart()` directly on the jenkinsrule instance?

jglick: I have no experience with the JUnit 5 facilities and see no reason to ever switch from 4.

Regarding `JenkinsRule.restart`, I would not recommend it. It seems to make a copy of `$JENKINS_HOME`, which `JenkinsSessionRule` does not need to do. Also it does not support performing test steps in between Jenkins sessions, as `JenkinsSessionRule` does.

timja: fyi @strangelookingnerd  ^^

timja (approved):

src/main/java/org/jvnet/hudson/test/RestartableJenkinsRule.java
strangelookingnerd: I‘m all for deprecating the rule. Overall I still think the transition towards JUnit5 is something we should promote rather than relying on a framework that is no longer maintained. I understand @jglick point that as of now there is little reason to switch, but that may change in the future.

jglick: > something we should promote

You should add Javadoc to various 4-specific utilities noting their replacements.

PR #936, opened 2025-03-14
`JenkinsRule#restart` preserves the original root path

strangelookingnerd:
My initial implementation of `JenkinsRule#restart` in https://github.com/jenkinsci/jenkins-test-harness/pull/716 seems to be flawed in a sense that the underlying Jenkins instance is restored with a copy of the original `JENKINS_HOME` under a *new* path.

In https://github.com/jenkinsci/throttle-concurrent-builds-plugin/pull/302 in became clear that this impacts some corner cases where this path is actually relevant for some plugins (see https://github.com/jenkinsci/throttle-concurrent-builds-plugin/pull/302#issuecomment-2718877688).

This PR aims to remove this flaw. Now `JenkinsRule#restart` will start a new instance with a copy of the original `JENKINS_HOME` under the *same* path as the previous instance. This is achieved by creating a `backup.zip` from the current `JENKINS_HOME` before shutting it down and starting a new instance from the unzipped backup that has the same path as before.

I do not assume that this change will cause any incompatibilities as I can not find evidence of a tests validating / building upon the flaws of the initial implementation.

Ping @basil @jglick Your input would be very welcome.

### Testing done

Adapted the existing tests. 

### Submitter checklist

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -3097,35 +3097,26 @@ public class JenkinsRule implements TestRule, MethodRule, RootAction {
      * Restart the current instance with the same port and a copy of its {@code JENKINS_HOME}.
      */
     public void restart() throws Throwable {
-        // create backup of current instance in new home
-        URL source = jenkins.getRootDir().toURI().toURL();
-        File copy = new TemporaryDirectoryAllocator().allocate();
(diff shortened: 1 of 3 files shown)


timja (approved):

jglick:
> creating a `backup.zip` from the current `JENKINS_HOME` before shutting it down and starting a new instance from the unzipped backup

Sounds better, though still inferior to `JenkinsSessionRule` which does not need to touch files at all.

PR #939, opened 2025-03-19
Fix RealJenkinsRule to properly set contextPath when using withPrefix…

### LINKED ISSUE #933: Custom prefix not forwarded to `CustomJenkinsRule`
### Jenkins and plugins versions report

n/a

### What Operating System are you using (both controller, and any agents involved in the problem)?

n/a

### Reproduction steps

```java
import static org.hamcrest.CoreMatchers.is;
import static org.hamcrest.MatcherAssert.assertThat;

import org.junit.Rule;
import org.junit.Test;
import org.jvnet.hudson.test.JenkinsRule;
import org.jvnet.hudson.test.RealJenkinsRule;

public class ContextPathTest {
    @Rule
    public RealJenkinsRule rjr = new RealJenkinsRule().withPrefix("");

    @Test
    public void test() throws Throwable {
        rjr.then(ContextPathTest::_test);
    }

    public static void _test(JenkinsRule jr) throws Exception {
        assertThat(jr.contextPath, is(""));
    }
}
```

### Expected Results

Test passes

### Actual Results

```
[ERROR] Errors: 
[ERROR]   ContextPathTest.test:17 » Step Remote step threw an exception: java.lang.AssertionError: 
Expected: is ""
     but: was "/jenkins"
[INFO] 
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0
```

### Anything else?

Previously: #917 being addressed (incompletely I guess) by #924.

The test looks kinda pointless, but many tests involving custom context paths/prefixes are parameterized (empty/nonempty prefix), and then having access to the actual prefix at that point is useful.

### Are you interested in contributing a fix?

_No response_

AhmedFatthy1040:
This PR fixes an issue where `RealJenkinsRule` doesn't properly set the `contextPath` field in `CustomJenkinsRule` when `withPrefix("")` is used. The current implementation doesn't propagate the empty prefix to the `contextPath` field of `JenkinsRule`. Closes #933 

### Testing done

A test case `ContextPathTest` has been added that verifies the fix works correctly. The test:
1. Creates a `RealJenkinsRule` instance with `withPrefix("")`
2. Asserts that the resulting `JenkinsRule` instance has an empty string in its `contextPath` field

The test passes with the fix, but would fail without it.

### Submitter checklist

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
@@ -1651,6 +1651,8 @@ public final class RealJenkinsRule implements TestRule {
         public CustomJenkinsRule(URL url) throws Exception {
             this.jenkins = Jenkins.get();
             this.url = url;
+            // Set the contextPath to match the URL path
+            this.contextPath = url.getPath().replaceAll("/$", "");
             if (jenkins.isUsageStatisticsCollected()) {
                 jenkins.setNoUsageStatistics(true); // cannot use JenkinsRule._configureJenkinsForTest earlier because it tries to save config before loaded
             }
--- a/src/test/java/org/jvnet/hudson/test/ContextPathTest.java
+++ b/src/test/java/org/jvnet/hudson/test/ContextPathTest.java
@@ -1,0 +1,23 @@
+package org.jvnet.hudson.test;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.RealJenkinsRule;
+
+public class ContextPathTest {
+    @Rule
+    public RealJenkinsRule rjr = new RealJenkinsRule().withPrefix("");
+
+    @Test
+    public void test() throws Throwable {
+        rjr.then(ContextPathTest::_test);
+    }
+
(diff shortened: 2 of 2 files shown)


Vlatombe:
Hi @AhmedFatthy1040, thank you for the pull request !

For the testing, I think you could amend existing tests (https://github.com/jenkinsci/jenkins-test-harness/blob/82aaf75f27b32d43d603560a70b24a03ecce4265/src/test/java/org/jvnet/hudson/test/RealJenkinsRuleTest.java#L108-L124) to check for context path in addition to url.

The context path itself could be added to `InputPayload` instead of being parsed every time.

AhmedFatthy1040:
Hi @Vlatombe , thanks for your review! I’ve made the changes, tested them, and pushed the updates. Let me know if you have any feedback!

src/main/java/org/jvnet/hudson/test/RealJenkinsRule.java
```diff
@@ -1684,9 +1685,10 @@ public void doTimeout(@QueryParameter String token) {
     public static final class CustomJenkinsRule extends JenkinsRule implements AutoCloseable {
         private final URL url;
 
-        public CustomJenkinsRule(URL url) throws Exception {
+        public CustomJenkinsRule(URL url, String contextPath) throws Exception {
```
Vlatombe, line 1688: Since the class and constructor are public, I think you should keep the old constructor signature and mark it deprecated.

Vlatombe (approved):

PR #205, opened 2020-03-31
Add support for customization on the server

Wadeck:
Used in https://github.com/jenkinsci/jenkins/pull/4615. 

With this modification one will be able to customize the context before the test server is started.

PR #941, opened 2025-03-24
[JENKINS-75572] Add support for Jakarta EE 10

basil:
### Context

Part of the following series of PRs:

- https://github.com/jenkinsci/jelly/pull/140
- https://github.com/jenkinsci/stapler/pull/617
- https://github.com/jenkinsci/winstone/pull/422
- https://github.com/jenkinsci/jenkins/pull/10461
- https://github.com/jenkinsci/jenkins-test-harness/pull/941

### Problem

The test harness only supports EE 9, precluding an eventual upgrade to EE 10.

### Solution

Support both EE 9 and EE 10 in the test harness, dynamically selecting the version of Jakarta EE based on the Jenkins core under test. This is similar at a high level to a previous PR that added support for both EE 8 and EE 9, though the details are different: in this case, the same `jakarta.servlet` types are in use, and the implementation is much cleaner.

#### Servlet API

Normally, the test harness is consumed in the plugin parent POM, which also [adds a `provided` dependency on the Jakarta EE servlet API](https://github.com/jenkinsci/plugin-pom/blob/2c1afb0bbc95879e4288967c457e1c40d3e5d633/pom.xml#L260-L265). The plugin parent POM does not declare a version, so the version comes from the Jenkins core BOM.

Prior to this PR, the test harness also declared a (transitive) dependency on the Jakarta EE servlet API via Jetty. When we supported both EE 8 and EE 9 concurrently in this repository, there was no conflict because the two Maven packages used different coordinates. In this PR, `jakarta.servlet:jakarta.servlet-api` and its close relative `org.eclipse.jetty.toolchain:jetty-jakarta-servlet-api` exist for both EE 9 and EE 10 with different version numbers, so a conflict has to be resolved. We have resolved it by excluding `jakarta.servlet:jakarta.servlet-api` and `org.eclipse.jetty.toolchain:jetty-jakarta-servlet-api` from being pulled in transitively by Jetty and instead declaring a direct `provided` dependency on `jakarta.servlet:jakarta.servlet-api`, as does the plugin parent POM. Jenkins core's test suite is not implemented like the plugin parent POM, so we also have to add a `provided` dependency there, as in https://github.com/jenkinsci/jenkins/pull/10554. For the Jenkins test harness's own test suite, we need to support different core versions, so we also pull in the core BOM here so that when testing the test harness we get the correct Jakarta EE JAR for the Jenkins core under test. The end result of all of this is:

- `jakarta.servlet:jakarta.servlet-api` is provided by core at the Jakarta EE version matching Winstone
- The test harness uses the above to figure out which Jetty web application context to load (EE 9 vs EE 10)
- The dependency pattern is consistent between the plugin parent POM, the Jenkins core test suite, and the Jenkins test harness test suite.

#### Refactoring

There were two copies of this code already, one for `HudsonTestCase` and one for `JenkinsRule`. Rather than multiply that into four copies (as we did when adding support for EE 8 and EE 9 concurrently), we took this opportunity to refactor the code into a more sustainable pattern.

There is now a `JettyProvider` class that contains a single copy of the common code. `JenkinsRule` and `HudsonTestCase` delegate to this. The two implementations are in `Jetty12EE9Provider` and `Jetty12EE10Provider` which contain the EE 9/10 specific bits, calling up to their parent class for common code. The concrete classes are indexed in `META-INF/services` and loaded by `ServiceLoader`; the constructor's job is to test whether the class is applicable to the given EE version and to throw an exception if it is not.

### Testing done

- Core EE 9: https://ci.jenkins.io/job/Core/job/jenkins/job/PR-10554/1/
- Core EE 10: https://ci.jenkins.io/job/Core/job/jenkins/job/PR-10461/19/ (single failure was a flake)
- Plugins EE 9: https://ci.jenkins.io/job/Tools/job/bom/job/PR-4900/1/
- Plugins EE 10: https://ci.jenkins.io/job/Tools/job/bom/job/PR-4779/13/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,9 +74,24 @@ THE SOFTWARE.
         <scope>import</scope>
       </dependency>
       <dependency>
+        <groupId>org.eclipse.jetty.ee10</groupId>
+        <artifactId>jetty-ee10-bom</artifactId>
+        <version>${jetty.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+      <dependency>
         <groupId>org.eclipse.jetty.ee9</groupId>
         <artifactId>jetty-ee9-bom</artifactId>
         <version>${jetty.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+      <dependency>
+        <!-- Pull in servlet API based on Jenkins core in use -->
+        <groupId>org.jenkins-ci.main</groupId>
+        <artifactId>jenkins-bom</artifactId>
+        <version>${jenkins.version}</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
@@ -126,9 +141,70 @@ THE SOFTWARE.
       </exclusions>
     </dependency>
     <dependency>
+      <groupId>org.eclipse.jetty.ee10</groupId>
+      <artifactId>jetty-ee10-webapp</artifactId>
+      <exclusions>
+        <!-- Provided by plugin parent POM at Jenkins core version -->
+        <exclusion>
+          <groupId>jakarta.servlet</groupId>
+          <artifactId>jakarta.servlet-api</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.eclipse.jetty.toolchain</groupId>
+          <artifactId>jetty-jakarta-servlet-api</artifactId>
+        </exclusion>
+        <!-- Provided by Jenkins core -->
+        <exclusion>
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-api</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.jetty.ee10.websocket</groupId>
+      <artifactId>jetty-ee10-websocket-jetty-server</artifactId>
+      <!-- or jetty-ee10-websocket-javax-server -->
+      <exclusions>
+        <!-- Provided by Jenkins core -->
+        <exclusion>
+          <groupId>jakarta.annotation</groupId>
+          <artifactId>jakarta.annotation-api</artifactId>
+        </exclusion>
+        <!-- Provided by plugin parent POM at Jenkins core version -->
+        <exclusion>
+          <groupId>jakarta.servlet</groupId>
+          <artifactId>jakarta.servlet-api</artifactId>
+        </exclusion>
+        <!-- Not shipped in Winstone either and causes unnecessary log spam -->
+        <exclusion>
+          <groupId>org.eclipse.jetty.ee10</groupId>
+          <artifactId>jetty-ee10-annotations</artifactId>
+        </exclusion>
+        <!-- Provided by plugin parent POM at Jenkins core version -->
+        <exclusion>
+          <groupId>org.eclipse.jetty.toolchain</groupId>
+          <artifactId>jetty-jakarta-servlet-api</artifactId>
+        </exclusion>
+        <!-- Provided by Jenkins core -->
+        <exclusion>
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-api</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
       <groupId>org.eclipse.jetty.ee9</groupId>
       <artifactId>jetty-ee9-webapp</artifactId>
       <exclusions>
+        <!-- Provided by plugin parent POM at Jenkins core version -->
+        <exclusion>
+          <groupId>jakarta.servlet</groupId>
+          <artifactId>jakarta.servlet-api</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.eclipse.jetty.toolchain</groupId>
+          <artifactId>jetty-jakarta-servlet-api</artifactId>
+        </exclusion>
         <!-- Provided by Jenkins core -->
         <exclusion>
           <groupId>org.slf4j</groupId>
@@ -146,6 +222,16 @@ THE SOFTWARE.
           <groupId>jakarta.annotation</groupId>
           <artifactId>jakarta.annotation-api</artifactId>
         </exclusion>
+        <!-- Provided by plugin parent POM at Jenkins core version -->
+        <exclusion>
+          <groupId>jakarta.servlet</groupId>
+          <artifactId>jakarta.servlet-api</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.eclipse.jetty.toolchain</groupId>
+          <artifactId>jetty-jakarta-servlet-api</artifactId>
+        </exclusion>
+        <!-- Provided by Jenkins core -->
         <exclusion>
           <groupId>org.slf4j</groupId>
           <artifactId>slf4j-api</artifactId>
@@ -198,6 +284,11 @@ THE SOFTWARE.
       <artifactId>junit-jupiter-api</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.kohsuke.metainf-services</groupId>
+      <artifactId>metainf-services</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
       <groupId>org.netbeans.modules</groupId>
       <artifactId>org-netbeans-insane</artifactId>
       <version>RELEASE200</version>
@@ -211,6 +302,11 @@ THE SOFTWARE.
       <groupId>org.openjdk.jmh</groupId>
       <artifactId>jmh-generator-annprocess</artifactId>
       <version>${jmh.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>jakarta.servlet</groupId>
+      <artifactId>jakarta.servlet-api</artifactId>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci</groupId>
--- a/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
+++ b/src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
@@ -117,6 +117,8 @@ import java.util.Enumeration;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
+import java.util.ServiceConfigurationError;
+import java.util.ServiceLoader;
 import java.util.Timer;
 import java.util.TimerTask;
 import java.util.UUID;
@@ -140,21 +142,11 @@ import org.acegisecurity.GrantedAuthority;
 import org.acegisecurity.userdetails.UserDetails;
 import org.acegisecurity.userdetails.UsernameNotFoundException;
 import org.apache.commons.beanutils.PropertyUtils;
-import org.eclipse.jetty.ee9.webapp.WebAppContext;
-import org.eclipse.jetty.ee9.websocket.server.config.JettyWebSocketServletContainerInitializer;
-import org.eclipse.jetty.http.HttpCompliance;
-import org.eclipse.jetty.http.UriCompliance;
 import org.eclipse.jetty.security.HashLoginService;
 import org.eclipse.jetty.security.LoginService;
 import org.eclipse.jetty.security.UserStore;
-import org.eclipse.jetty.server.HttpConfiguration;
-import org.eclipse.jetty.server.HttpConnectionFactory;
 import org.eclipse.jetty.server.Server;
-import org.eclipse.jetty.server.ServerConnector;
-import org.eclipse.jetty.server.handler.gzip.GzipHandler;
-import org.eclipse.jetty.util.resource.ResourceFactory;
 import org.eclipse.jetty.util.security.Password;
-import org.eclipse.jetty.util.thread.QueuedThreadPool;
 import org.htmlunit.AjaxController;
 import org.htmlunit.AlertHandler;
 import org.htmlunit.BrowserVersion;
@@ -185,6 +177,7 @@ import org.htmlunit.javascript.host.xml.XMLHttpRequest;
 import org.htmlunit.util.NameValuePair;
 import org.htmlunit.xml.XmlPage;
 import org.jvnet.hudson.test.HudsonHomeLoader.CopyExisting;
+import org.jvnet.hudson.test.jetty.JettyProvider;
 import org.jvnet.hudson.test.recipes.Recipe;
 import org.jvnet.hudson.test.recipes.Recipe.Runner;
 import org.jvnet.hudson.test.recipes.WithPlugin;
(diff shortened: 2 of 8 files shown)


pom.xml
```diff
@@ -73,13 +73,28 @@ THE SOFTWARE.
...
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <!-- Pull in servlet API based on Jenkins core in use -->
+        <groupId>org.jenkins-ci.main</groupId>
+        <artifactId>jenkins-bom</artifactId>
+        <version>${jenkins.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
```
basil, line 97: Matching the plugin parent POM and Jenkins core test suite. This is needed so that we get the version of `jakarta.servlet:jakarta.servlet-api` (added a few lines below) matching the Jenkins core under test (given by `jenkins.version`) when running the Jenkins test harness test suite.

pom.xml
```diff
@@ -125,10 +140,71 @@ THE SOFTWARE.
...
+      <exclusions>
+        <!-- Provided by plugin parent POM at Jenkins core version -->
+        <exclusion>
+          <groupId>jakarta.servlet</groupId>
+          <artifactId>jakarta.servlet-api</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.eclipse.jetty.toolchain</groupId>
+          <artifactId>jetty-jakarta-servlet-api</artifactId>
+        </exclusion>
```
basil, line 155: Provided below by the `provided`-scoped dependency on `jakarta.servlet:jakarta.servlet-api`.

pom.xml
```diff
@@ -125,10 +140,71 @@ THE SOFTWARE.
...
+      <groupId>org.eclipse.jetty.ee10.websocket</groupId>
+      <artifactId>jetty-ee10-websocket-jetty-server</artifactId>
+      <!-- or jetty-ee10-websocket-javax-server -->
+      <exclusions>
+        <!-- Provided by Jenkins core -->
+        <exclusion>
+          <groupId>jakarta.annotation</groupId>
+          <artifactId>jakarta.annotation-api</artifactId>
+        </exclusion>
+        <!-- Provided by plugin parent POM at Jenkins core version -->
```
basil, line 173: Ditto

pom.xml
```diff
@@ -125,10 +140,71 @@ THE SOFTWARE.
...
+        <exclusion>
+          <groupId>jakarta.servlet</groupId>
+          <artifactId>jakarta.servlet-api</artifactId>
+        </exclusion>
+        <!-- Not shipped in Winstone either and causes unnecessary log spam -->
+        <exclusion>
+          <groupId>org.eclipse.jetty.ee10</groupId>
+          <artifactId>jetty-ee10-annotations</artifactId>
+        </exclusion>
```
basil, line 182: Jetty 12 EE 10 appears to re-introduce an old Jetty bug fixed a long time ago: the WebSocket package unnecessarily depends on the Jetty Annotations library. And in this case, it wreaks havoc, finding the EE 9 WebSocket initializer on the class path and trying to load it into an EE 10 web application context, which fails. Both here and in https://github.com/jenkinsci/winstone/pull/422, excluding this package to retain the same behavior as EE 9.

pom.xml
```diff
@@ -197,6 +283,11 @@ THE SOFTWARE.
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.kohsuke.metainf-services</groupId>
+      <artifactId>metainf-services</artifactId>
+      <optional>true</optional>
+    </dependency>
```
basil, line 290: Needed to create the `META-INF/services` index for the two `JettyProvider` implementations below.

pom.xml
```diff
@@ -212,6 +303,11 @@ THE SOFTWARE.
       <artifactId>jmh-generator-annprocess</artifactId>
       <version>${jmh.version}</version>
     </dependency>
+    <dependency>
+      <groupId>jakarta.servlet</groupId>
+      <artifactId>jakarta.servlet-api</artifactId>
+      <scope>provided</scope>
+    </dependency>
```
basil, line 310: Matching the plugin parent POM. Also requires the same addition to the Jenkins core test suite, as in https://github.com/jenkinsci/jenkins/pull/10554.

src/main/java/org/jvnet/hudson/test/HudsonTestCase.java
```diff
@@ -547,51 +540,25 @@ public void setPluginManager(PluginManager pluginManager) {
      * that we need for testing.
      */
     protected ServletContext createWebServer2() throws Exception {
-        QueuedThreadPool qtp = new QueuedThreadPool();
```
basil, line 550: All of the deleted code in this file was previously duplicated with `JenkinsRule`. It is now consolidated in:

- `JettyProvider`
- `Jetty12EE9Provider`
- `Jetty12EE10Provider`

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -792,121 +780,24 @@ public File getWebAppRoot() throws Exception {
      * that we need for testing.
      */
     protected ServletContext createWebServer2() throws Exception {
```
basil, line 782: This method is still `protected`, but subclasses are highly discouraged. The only remaining subclasses are in Jenkins core's test suite as well as a single CloudBees proprietary plugin. For the most part, they represent technical debt and should be replaced with `RealJenkinsRule`.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -792,121 +780,24 @@ public File getWebAppRoot() throws Exception {
...
-    }
-
-    /**
-     * Prepares a webapp hosting environment to get {@link jakarta.servlet.ServletContext} implementation
-     * that we need for testing.
-     * 
-     * @param contextAndServerConsumer configures the {@link WebAppContext} and the {@link Server} for the instance, before they are started
-     * @since 2.63
-     */
-    protected ServletContext createWebServer2(@CheckForNull BiConsumer<WebAppContext, Server> contextAndServerConsumer)
```
basil, line 805: This was added in https://github.com/jenkinsci/jenkins-test-harness/pull/205 and is only consumed by https://github.com/jenkinsci/jenkins/pull/4615, itself a "more correct" alternative to https://github.com/jenkinsci/jenkins/pull/4611. While I don't doubt that https://github.com/jenkinsci/jenkins/pull/4615 was "more correct", it is also more fragile, and it requires https://github.com/jenkinsci/jenkins-test-harness/pull/205 which exposes a particular EE version (in this case EE 9) in a public test harness API, something which is going to be inherently fragile in the future as we upgrade to future EE versions. Rather than deal with this problem, I elected to revive the alternative "less correct" (but also less fragile) approach from https://github.com/jenkinsci/jenkins/pull/4611 in https://github.com/jenkinsci/jenkins/pull/10554. With this (only) consumer out of the way, we can delete this public API that hard-codes a particular EE version in the `WebAppContext` type. I consider this a medium-term solution; ultimately, in the long term `BootFailureTest` should be rewritten to avoid `JenkinsRule` entirely in favor of `RealJenkinsRule`, but that requires a bit more time than I have right now.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -792,121 +780,24 @@ public File getWebAppRoot() throws Exception {
...
-     *
-     * @param contextPath          the context path at which to put Jenkins
-     * @param portSetter           the port on which the server runs will be set using this function
-     * @param classLoader          the class loader for the {@link WebAppContext}
-     * @param localPort            port on which the server runs
-     * @param loginServiceSupplier configures the {@link LoginService} for the instance
-     * @return                     the {@link Server}
-     * @since 2.50
-     */
-    public static WebAppContext _createWebAppContext2(
```
basil, line 830: Public API, but only consumed within this repository (I checked CloudBees, too), so safe to delete.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -792,121 +780,24 @@ public File getWebAppRoot() throws Exception {
...
-     * @param contextPath              the context path at which to put Jenkins
-     * @param portSetter               the port on which the server runs will be set using this function
-     * @param classLoader              the class loader for the {@link WebAppContext}
-     * @param localPort                port on which the server runs
-     * @param loginServiceSupplier     configures the {@link LoginService} for the instance
-     * @param contextAndServerConsumer configures the {@link WebAppContext} and the {@link Server} for the instance, before they are started
-     * @return                         the {@link Server}
-     * @since 2.50
-     */
-    public static WebAppContext _createWebAppContext2(
```
basil, line 851: Public API, but only consumed within this repository (I checked CloudBees, too), so safe to delete.

src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -792,121 +780,24 @@ public File getWebAppRoot() throws Exception {
...
-            String contextPath,
-            Consumer<Integer> portSetter,
-            ClassLoader classLoader,
-            int localPort,
-            Supplier<LoginService> loginServiceSupplier,
-            @CheckForNull BiConsumer<WebAppContext, Server> contextAndServerConsumer)
-            throws Exception {
-        QueuedThreadPool qtp = new QueuedThreadPool();
```
basil, line 859: All of the deleted code in this file was previously duplicated with `HudsonTestCase`. It is now consolidated in:

- `JettyProvider`
- `Jetty12EE9Provider`
- `Jetty12EE10Provider`

src/main/java/org/jvnet/hudson/test/jetty/Jetty12EE10Provider.java
```diff
@@ -0,0 +1,84 @@
...
+import org.jvnet.hudson.test.NoListenerConfiguration2;
+import org.jvnet.hudson.test.WarExploder;
+import org.kohsuke.MetaInfServices;
+
+@MetaInfServices(JettyProvider.class)
+public final class Jetty12EE10Provider extends JettyProvider {
+
+    public Jetty12EE10Provider() {
+        try {
+            ServletRequest.class.getDeclaredMethod("getRequestId");
```
basil, line 44: This method was newly introduced in EE 10, so it will throw an exception in EE 9 and prevent the `ServiceLoader` from finding this implementation.

src/main/java/org/jvnet/hudson/test/jetty/Jetty12EE10Provider.java
```diff
@@ -0,0 +1,84 @@
...
+            throws Exception {
+        createServer();
+
+        File explodedWarDir = WarExploder.getExplodedDir();
+        WebAppContext context = new WebAppContext(explodedWarDir.getPath(), contextPath) {
+            @Override
+            public void preConfigure() throws Exception {
+                super.preConfigure();
+                getServletHandler().setDecodeAmbiguousURIs(true);
+            }
```
basil, line 63: Matching https://github.com/jenkinsci/winstone/pull/422, this hunk is new for EE 10 and was recommended by the Jetty developers in order to support backslashes in legacy mode.

src/main/java/org/jvnet/hudson/test/jetty/Jetty12EE10Provider.java
```diff
@@ -0,0 +1,84 @@
...
+                return loader;
+            }
+        };
+        context.setBaseResource(ResourceFactory.of(context).newResource(explodedWarDir.getPath()));
+        context.setClassLoader(getClass().getClassLoader());
+        context.setConfigurationDiscovered(true);
+        context.addBean(new NoListenerConfiguration2(context));
+        context.setServer(server);
+        context.getSecurityHandler().setLoginService(loginServiceSupplier.get());
+        JettyWebSocketServletContainerInitializer.configure(context, null);
```
basil, line 77: While it looks like this is duplicate code with `Jetty12EE9Provider`, it actually uses different types (with no common ancestor), so the apparent symbolic duplication can't be eliminated.

src/main/java/org/jvnet/hudson/test/jetty/Jetty12EE9Provider.java
```diff
@@ -0,0 +1,78 @@
...
+import org.jvnet.hudson.test.NoListenerConfiguration2;
+import org.jvnet.hudson.test.WarExploder;
+import org.kohsuke.MetaInfServices;
+
+@MetaInfServices(JettyProvider.class)
+public final class Jetty12EE9Provider extends JettyProvider {
+
+    public Jetty12EE9Provider() {
+        try {
+            ServletRequest.class.getDeclaredMethod("getRealPath", String.class);
```
basil, line 44: This method was deleted in EE 10, so this lookup will fail in EE 10 with an exception and prevent the `ServiceLoader` from finding this implementation.

src/main/java/org/jvnet/hudson/test/jetty/JettyProvider.java
```diff
@@ -0,0 +1,98 @@
...
+     * Prepares a webapp hosting environment and returns a context containing the {@link ServletContext}, port, and
+     * server for testing.
+     */
+    public abstract Context createWebServer(
+            int localPort, String contextPath, Supplier<LoginService> loginServiceSupplier) throws Exception;
+
+    protected final void createServer() {
+        QueuedThreadPool qtp = new QueuedThreadPool();
+        qtp.setName("Jetty (JenkinsRule)");
+        server = new Server(qtp);
```
basil, line 65: Moved from the original in `JenkinsRule`.

src/main/java/org/jvnet/hudson/test/jetty/Jetty12EE9Provider.java
```diff
@@ -0,0 +1,78 @@
...
+                return loader;
+            }
+        };
+        context.setBaseResource(ResourceFactory.of(context).newResource(explodedWarDir.getPath()));
+        context.setClassLoader(getClass().getClassLoader());
+        context.setConfigurationDiscovered(true);
+        context.addBean(new NoListenerConfiguration2(context));
+        context.setServer(server);
+        context.getSecurityHandler().setLoginService(loginServiceSupplier.get());
+        JettyWebSocketServletContainerInitializer.configure(context, null);
```
basil, line 71: Moved from the original in `JenkinsRule`.

src/main/java/org/jvnet/hudson/test/jetty/JettyProvider.java
```diff
@@ -0,0 +1,98 @@
...
+        config.setHttpCompliance(HttpCompliance.RFC7230);
+        config.setUriCompliance(UriCompliance.LEGACY);
+        connector.setHost("localhost");
+        if (System.getProperty("port") != null) {
+            connector.setPort(Integer.parseInt(System.getProperty("port")));
+        } else if (localPort != 0) {
+            connector.setPort(localPort);
+        }
+        server.addConnector(connector);
+        return connector;
```
basil, line 83: Moved from the original in `JenkinsRule`.

src/main/java/org/jvnet/hudson/test/jetty/JettyProvider.java
```diff
@@ -0,0 +1,98 @@
...
+        if (compression.equals("gzip")) {
+            GzipHandler gzipHandler = new GzipHandler();
+            gzipHandler.setHandler(handler);
+            server.setHandler(gzipHandler);
+        } else if (compression.equals("none")) {
+            server.setHandler(handler);
+        } else {
+            throw new IllegalArgumentException("Unexpected compression scheme: " + compression);
+        }
+    }
```
basil, line 97: Moved from the original in `JenkinsRule`.

alecharp (approved):
I do hope the method to detect EE 9 won't be added back in future update.

Thank you for https://github.com/jenkinsci/jenkins-test-harness/pull/941/files#r2045575779 because I was a bit confused in GH page as I could see class hierarchy.

PR #963, opened 2025-05-01
`InboundAgentRule.AgentArguments` should not refer to deprecated `*.jnlp`

jglick:
Cleanup from #670.


timja (approved):

jglick:
Actually there was one CloudBees CI test which relied on passing `-jnlpUrl` even when using a new agent JAR, so may need to rework this.

PR #988, opened 2025-06-27
Add JUnit5 equivalents of JUnit4 Rule implementations

strangelookingnerd:
In the light of https://github.com/jenkinsci/throttle-concurrent-builds-plugin/pull/302#pullrequestreview-2936124170 and https://github.com/jenkinsci/text-finder-plugin/pull/255#issuecomment-3010017986 I want to create JUnit5 equivalents for the most commonly used `Rule` implementations:

`BuildWatcher` -> `BuildWatcherExtension` 
`InboundAgentRule` -> `InboundAgentExtension` 
`JenkinsSessionRule` -> `JenkinsSessionExtension` 
`RealJenkinsRule` -> `RealJenkinsExtension` 

Feedback would be highly appreciated (ping @basil @timja @jglick)

The current state mostly copies the existing `*Rule` into an `*Extension`, implementing the corresponding wrapper-methods of the respective framework. 

I don't like that this will likely cause the implementations to diverge at some point. However I failed to come up with a better solution that does not force me to refactor the entire thing. Something that would already help is to allow the classes to be compared more easily. Right now the formatting is all over the place - `spotless` could help.

Overall there is some very questionable coding, hacks and workarounds in both implementations, existing and new. I already removed some `deprecations` as well as `TODO` that are no longer relevant. Another pair of eyes would however help here.

As always, give me a ping if there any questions.

### Testing done

I copied the existing tests for the rules above and adapted them to use JUnit5 and my `*Extension` implementations.
My main concern right now is that even the existing tests are only covering a small percentage of the functionallity provided, meaning there is a certain risk that some things simply do not work (yet).

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed


timja (commented):
From a cursory glance it looks fine, best would be to try migrate a few plugins and see how it does.

timja:
> From a cursory glance it looks fine, best would be to try migrate a few plugins and see how it does.

What plugins have you migrated for each extension?

strangelookingnerd:
> What plugins have you migrated for each extension?

I have some prepared, will create & link the PR.

strangelookingnerd:
Overall my approach seems to work just fine. 

I found a tiny issue with meta-annotations such as `LocalData` in `JenkinsSessionExtension` that was fixed in [14ec804](https://github.com/jenkinsci/jenkins-test-harness/pull/988/commits/14ec80488532ea5380feb4ddcd4a8f4605340d8b).

Should I add some more migrations or would the four linked above suffice?

jglick:
> I don't like that this will likely cause the implementations to diverge at some point. However I failed to come up with a better solution that does not force me to refactor the entire thing.

Belated opinion: it would have been better to pause the JUnit 5 transition until all relevant classes have been refactored into non-framework-specific utilities, and then both JUnit 4 rules & JUnit 5 extensions introduced which are thin wrappers over those utilities which you could use if you found it more legible to do so than to call the utilities directly.

PR #1020, opened 2025-09-06
Replace 'slave' and 'master with 'agent' and 'controller' in comments/…

Rohit-62018:
# This PR request update comments and log messages to replace the term 'slave' and 'master' with agent and controller
for consistency with current Jenkins new terminology.

Notes:
- Only comments and log messages were updated.
- References to actual classes and methods such as `PretendSlave` and `createSlave` have been left unchanged 
  to avoid breaking Javadoc generation or code references.
- This change is purely cosmetic and does not affect functionality.

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
+++ b/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
@@ -71,7 +71,7 @@ import org.junit.rules.ExternalResource;
  * <p>To avoid flakiness when tearing down the test, ensure that the agent has gone offline with:
  *
  * <pre>
- * Slave agent = inboundAgents.createAgent(r, […]);
+ * agent agent = inboundAgents.createAgent(r, […]);
  * try {
  *     […]
  * } finally {
@@ -662,7 +662,7 @@ public final class InboundAgentRule extends ExternalResource {
             throw new AssertionError("no such agent: " + name);
         }
         if (!(node instanceof Slave)) {
-            throw new AssertionError("agent is not a Slave: " + name);
+            throw new AssertionError("agent is not a agent: " + name);
         }
         r.waitOnline((Slave) node);
         if (!loggers.isEmpty()) {
@@ -701,7 +701,7 @@ public final class InboundAgentRule extends ExternalResource {
         s.setLabelString(options.getLabel());
         s.setRetentionStrategy(RetentionStrategy.NOOP);
         r.jenkins.addNode(s);
-        // SlaveComputer#_connect runs asynchronously. Wait for it to finish for a more deterministic test.
+        // AgentComputer#_connect runs asynchronously. Wait for it to finish for a more deterministic test.
         Computer computer = s.toComputer();
         while (computer == null || computer.getOfflineCause() == null) {
             Thread.sleep(100);
--- a/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
+++ b/src/main/java/org/jvnet/hudson/test/JenkinsRule.java
@@ -905,7 +905,7 @@ public class JenkinsRule implements TestRule, MethodRule, RootAction {
             long start = System.currentTimeMillis();
             while (slave.getChannel() == null) {
                 if (System.currentTimeMillis() > (start + 10000)) {
-                    throw new IllegalStateException("Timed out waiting on DumbSlave channel to connect.");
+                    throw new IllegalStateException("Timed out waiting on DumbAgent channel to connect.");
                 }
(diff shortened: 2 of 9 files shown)


src/main/java/org/jvnet/hudson/test/JenkinsRule.java
```diff
@@ -1234,7 +1234,7 @@ public void close() throws SecurityException {}
                 logger.addHandler(handler);
                 loggerReferences.add(logger);
             });
-            DeltaSupportLogFormatter.start = start; // match clock time on master
+            DeltaSupportLogFormatter.start = start; // match clock time on Controller
```
MarkEWaite, line 1237: Should not be upper case.  Not a proper noun.
```suggestion
            DeltaSupportLogFormatter.start = start; // match clock time on controller
```

src/main/java/org/jvnet/hudson/test/junit/jupiter/InboundAgentExtension.java
```diff
@@ -630,7 +630,7 @@ private static void waitForAgentOnline(JenkinsRule r, String name, Map<String, L
             throw new AssertionError("no such agent: " + name);
         }
         if (!(node instanceof Slave)) {
-            throw new AssertionError("agent is not a Slave: " + name);
+            throw new AssertionError("agent is not a Agent: " + name);
```
MarkEWaite, line 633: Capitalization mistake in the original string.
```suggestion
            throw new AssertionError("agent is not an agent: " + name);
```

src/main/java/org/jvnet/hudson/test/junit/jupiter/InboundAgentExtension.java
```diff
@@ -669,7 +669,7 @@ private static Slave createAgentJR(JenkinsRule r, Options options)
         s.setLabelString(options.getLabel());
         s.setRetentionStrategy(RetentionStrategy.NOOP);
         r.jenkins.addNode(s);
-        // SlaveComputer#_connect runs asynchronously. Wait for it to finish for a more deterministic test.
+        // AgentComputer#_connect runs asynchronously. Wait for it to finish for a more deterministic test.
```
MarkEWaite, line 672: No Java class named `AgentComputer`
```suggestion
        // SlaveComputer#_connect runs asynchronously. Wait for it to finish for a more deterministic test.
```

Rohit-62018:
Should I move the code to the local system and fix it again or accept your suggestions and request a review again? I want to keep the changes small because as you have noticed, I am making a lot of mistakes.

MarkEWaite:
> Should I move the code to the local system and fix it again or accept your suggestions and request a review again? I want to keep the changes small because as you have noticed, I am making a lot of mistakes.

I think you'll be more successful if you work with it locally and assure that it compiles locally.

src/main/java/org/jvnet/hudson/test/InboundAgentRule.java
```diff
@@ -662,7 +662,7 @@ private static void waitForAgentOnline(JenkinsRule r, String name, Map<String, L
             throw new AssertionError("no such agent: " + name);
         }
         if (!(node instanceof Slave)) {
-            throw new AssertionError("agent is not a Slave: " + name);
+            throw new AssertionError("agent is not an agent: " + name);
```
MarkEWaite, line 665: I am sorry that I missed this in my first review, but the previous message is the more correct message because it is reporting that `node` has the wrong type.  It is expecting an instance of type `Slave`
```suggestion
            throw new AssertionError("agent is not a Slave: " + name);
```

src/main/java/org/jvnet/hudson/test/junit/jupiter/InboundAgentExtension.java
```diff
@@ -630,7 +630,7 @@ private static void waitForAgentOnline(JenkinsRule r, String name, Map<String, L
             throw new AssertionError("no such agent: " + name);
         }
         if (!(node instanceof Slave)) {
-            throw new AssertionError("agent is not a Slave: " + name);
+            throw new AssertionError("agent is not an agent: " + name);
```
MarkEWaite, line 633: I made the same mistake here, not reading the surrounding code while reviewing.
```suggestion
            throw new AssertionError("agent is not a Slave: " + name);
```

MarkEWaite (changes requested):
I made two mistakes on my earlier review.  So sorry about that.  Two of the changes need to be removed because the assertion message is specifically referring to the `Slave` object type.

MarkEWaite (approved):

PR #1022, opened 2025-09-12
Replace ‘slave’ by ‘agent’ in some internal variable names

Rohit-62018:
This PR replaces all occurrences of the term ‘slave’ with ‘agent’ in variable names across the codebase.
No functional logic was changed, only variable names were updated for consistency.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,5 +6,5 @@
 buildPlugin(useContainerAgent: true, forkCount: '1C', configurations: [
   [platform: 'linux', jdk: 17],
   [platform: 'windows', jdk: 17],
-  [platform: 'linux', jdk: 21, jenkins: '2.527']
+  [platform: 'linux', jdk: 21, jenkins: '2.528']
 ])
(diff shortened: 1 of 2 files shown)


MarkEWaite:
> I want to rename functions that have the name slave, but I can't rename them directly because I don't know where those functions are being used. How can I rename them safely?

It is a long process to change a method name in Jenkins or any other system that values long term compatibility.  The process looks something like this:

* Create new API methods with the new names and call the new API methods from the old API methods, annotate the old API methods as deprecated
* Release a new version of the library that provides both new and old API methods
* Propose pull requests to all public Jenkins repositories that use the old API methods to replace old API usage with new API usage (after finding those usages, usually with the "usage in plugins" tool)
* Monitor those pull requests until they are merged and released
* Create a pull request that deletes the old API methods and document the removal through a blog post or an update to the Jenkins.io documentation site

This can be a multi-year project.

MarkEWaite (approved):

PR #1134, opened 2025-12-17
Improve injected tests

strangelookingnerd:
Addressing https://github.com/jenkinsci/maven-hpi-plugin/pull/846#discussion_r2626945138 and https://github.com/jenkinsci/jenkins-test-harness/pull/1132#discussion_r2626969032

* Consolidate tests to reduce runtime
* Fix null configuration parameters
* Only include required config parameters
* Use sub-classing rather than `@Suite`

Enables https://github.com/jenkinsci/maven-hpi-plugin/pull/848

### Testing done
Tested in https://github.com/jenkinsci/custom-folder-icon-plugin/pull/506

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/main/java/org/jvnet/hudson/test/injected/InjectedTestBase.java
+++ b/src/main/java/org/jvnet/hudson/test/injected/InjectedTestBase.java
@@ -54,10 +54,10 @@ import org.kohsuke.stapler.jelly.JellyClassLoaderTearOff;
  */
 @WithJenkins
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
-public abstract class InjectedTest {
+public abstract class InjectedTestBase {
 
     private static final JellyClassLoaderTearOff JCT =
-            new MetaClassLoader(InjectedTest.class.getClassLoader()).loadTearOff(JellyClassLoaderTearOff.class);
+            new MetaClassLoader(InjectedTestBase.class.getClassLoader()).loadTearOff(JellyClassLoaderTearOff.class);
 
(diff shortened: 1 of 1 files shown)


jglick:
> Still trying to figure out how to reproduce outside of CI.

I _think_ this has to do with use of https://plugins.jenkins.io/parallel-test-executor/ and thus an include/exclude list, rather than the existence of submodules as such. Will try to confirm, unless this gets merged and release soon in which case I will just see if this fixes it.

strangelookingnerd:
> I think https://docs.junit.org/5.0.1/api/org/junit/jupiter/api/Assertions.html#assertAll-org.junit.jupiter.api.function.Executable...- is what I was looking for. Probably unnecessary given the subclassing scheme but I need to confirm (once the source generator is fixed) that the `TEST-*.xml` in fact works properly.

That would be the last remedy I think. I don't really like the idea of cramping everything into a single test.

jglick:
For some reason this did not produce an incremental build. Checking with a local snapshot…

jglick (approved):
Seems to work better now, thanks! (Will not know for sure until this gets picked up in a POM update and used in a plugin running `parallel-test-executor`.)

```xml
<testsuite … name="org.jenkins_ci.plugins.durable_task.InjectedTest" …>
…
  <testcase name="testCliSanity" classname="org.jenkins_ci.plugins.durable_task.InjectedTest" time="0.007"/>
…
```

PR #1029, opened 2025-09-24
Make `MemoryAssert.assertGC` stricter when soft references are disallowed

dwnusbaum:
While testing https://github.com/jenkinsci/pipeline-groovy-lib-plugin/pull/199, I ran into cases where `MemoryAssert.assertGC(..., false)` was passing, but very slowly, because although there _were_ soft references, they were cleared out prior to an `OutOfMemoryError` occurring, and so we skipped the code that failed when `softErr != null`.

This PR adjusts the code to fail if `softErr != null` after existing the initial loop, so that we also fail in this kind of case.

I have not attempted to test this against other callers of this method, so this is just a draft PR.

### Testing done

See the new test in https://github.com/jenkinsci/pipeline-groovy-lib-plugin/pull/199.

### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

PR #1168, opened 2026-04-15
Fix assertGC passing when only soft references prevent collection

RanjitCS:
`softErr` was only checked inside `catch(OutOfMemoryError)`, but the JVM clears weak refs before throwing OOM, so the while loop exits normally and `softErr` is never evaluated. To prevent this we check `if (softErr != null) fail(softErr)` after the while loop so soft reference detection is checked regardless of how the loop exits   

### Testing done
Added a unit test with a softly-reachable-only object that fails without the fix  

### Submitter checklist
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/test/java/org/jvnet/hudson/test/MemoryAssertTest.java
+++ b/src/test/java/org/jvnet/hudson/test/MemoryAssertTest.java
@@ -33,6 +33,8 @@ import static org.junit.Assume.assumeTrue;
 import static org.jvnet.hudson.test.MemoryAssert.assertGC;
 import static org.jvnet.hudson.test.MemoryAssert.assertHeapUsage;
 
+import java.lang.ref.Reference;
+import java.lang.ref.SoftReference;
 import java.lang.ref.WeakReference;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


RanjitCS:
https://github.com/jenkinsci/jenkins-test-harness/pull/1029 already contains the same fix. we should merge either that pr or this one, asked in the comment there. i should have checked before raising this PR

src/test/java/org/jvnet/hudson/test/MemoryAssertTest.java
```diff
@@ -76,4 +78,20 @@ public void gc() {
                         + "]: {}; apparent weak references: {}",
                 actual.getMessage());
     }
+
+    @Test
+    public void gcFailsForSoftReferenceWithoutStrongRef() {
```
jglick, line 83: Would be good to port to #1029 right?

jglick: I guess after that merges, you could update this from base branch, resolve merge conflict in favor of trunk, and reopen this as a test-only PR.

src/test/java/org/jvnet/hudson/test/MemoryAssertTest.java
```diff
@@ -76,4 +78,20 @@ public void gc() {
...
+        Runtime.Version runtimeVersion = Runtime.version();
+        assumeTrue("TODO JENKINS-67974 works on Java 17 but not 11", runtimeVersion.feature() >= 17);
```
jglick: Is this necessary? https://github.com/jenkinsci/jenkins-test-harness/blob/65f48845f485a0d0c1871b0576cdf4cf8d298a7b/Jenkinsfile#L7-L9

src/test/java/org/jvnet/hudson/test/MemoryAssertTest.java
```diff
@@ -76,4 +78,20 @@ public void gc() {
...
+        // object held only by a soft reference (no strong ref), so the weak ref
+        // will be collected under memory pressure, but allowSoft=false should still fail
+        Object obj = new Object();
+        SoftReference<Object> soft = new SoftReference<>(obj);
+        WeakReference<Object> weak = new WeakReference<>(obj);
+        // clear strong stack root so object is only reachable via soft ref
+        obj = null;
+        assertThrows(AssertionError.class, () -> assertGC(weak, false));
+        // prevent JIT from collecting soft before assertGC finishes
+        Reference.reachabilityFence(soft);
```
jglick, line 93: TIL https://docs.oracle.com/en/java/javase/23/docs/api/java.base/java/lang/ref/Reference.html#reachabilityFence(java.lang.Object)

src/test/java/org/jvnet/hudson/test/MemoryAssertTest.java
RanjitCS: updated to be a test only PR

src/test/java/org/jvnet/hudson/test/MemoryAssertTest.java
RanjitCS: just copied from the other tests, but not needed you are right, removed the assert

jglick (approved):

PR #1189, opened 2026-08-31
test(harness): detect Jenkins-booting tests via PostDiscoveryFilter

rantoniuk:
Adds a JUnit Platform PostDiscoveryFilter to detect tests that boot a real Jenkins instance
(that are annotated with `@WithJenkins` or `@WithJenkinsConfiguredWithCode`) and exclude them when
`-Djenkins.tests.excludeJenkins=true` is set. The filter is registered via `META-INF/services` so
it is auto-discovered by the JUnit Platform. This is opt-in and does not change default test runs.

Addresses request to provide a harness-side helper so plugins can opt-in to fast-tests.

--- 

Real benefit of JenkinsBootTestFilter, at two scales:

#### Org-wide (jenkinsci top-level CI)

The Jenkins ecosystem has 1,800+ plugin repos, most of which pull in jenkins-test-harness and use @WithJenkins/JenkinsRule for integration tests. That population shares CI infrastructure (ci.jenkins.io agents, GitHub Actions minutes) and, more importantly, a shared pipeline template most plugins build from (the jenkins-infra "buildplan"/reusable workflow). That's the leverage point:

- **Zero-migration, org-wide rollout.** 
 
Because the classification is derived from an annotation every plugin already uses for functional reasons - not a tag/exclude list every maintainer has to hand-write and keep in sync - the shared pipeline template could add a "fast tests" pre-stage once, centrally, and every consuming plugin gets it for free on their next jenkins-test-harness bump. No 1,800-repo migration effort, no opt-in campaign, no risk of individual repos drifting out of sync with a tagging convention.

- **A cheap, universal PR-feedback tier**. 
 
Org infra could standardize on: fast-tests stage runs on every push (seconds, no Jenkins boot, immediate signal on logic bugs), full matrix (multi-JDK, Jenkins-booting ITs) runs afterward or gates merge. That's a latency win applied uniformly across the ecosystem rather than plugin-by-plugin reinvention.

- **Ecosystem-level visibility.** 
 
Since the filter's decision is mechanical and consistent, infra tooling could aggregate "fraction of test time spent booting Jenkins" across all plugin repos, useful data for prioritizing infra investment (e.g., is it worth funding faster JenkinsRule boot / container reuse, or parallelizing agents) instead of guessing.

- **Consistency of meaning.** 
 
"Fast tests" means _the same thing in every plugin repo that adopts it_, because the source of truth is the harness annotation, not each repo's bespoke Surefire <excludes>/tag convention (which today differ or don't exist at all in most plugins).

#### Per-plugin CI / local dev loop

- **Faster inner loop for contributors.** 
 
JenkinsRule boot (Jetty + PluginManager + full extension list init) is usually the dominant per-test cost. A contributor fixing a pure logic bug (validators, model classes, parsers) can run -Djenkins.tests.excludeJenkins=true locally and get feedback in seconds instead of minutes, without needing to know which tests are "safe" to skip.

- **A pre-merge fast-fail gate without rewriting the test suite.** 
 
A plugin's own CI can add a quick "fast tests" job ahead of the full build, catching obvious breakage before spending agent time on the expensive Jenkins-booting suite —pline for maintainers to adopt or forget.

- **No drift risk.** 

The alternative (manually tagging/excluding tests) requires maintainers to remember to tag every new Jenkins-booting test going forward; miss one and the "fast" run either silently boots Jenkins anyway or wrongly  it's automatically and permanently correct as long as the annotationis used correctly.

- **Test-pyramid nudge.** 

Visibility into the fast/Jenkins-booting split output) gives plugin authors a concrete signal to push more logic into pure unit tests rather than always reaching for @WithJenkins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,3 +14,9 @@ See https://javadoc.jenkins.io/component/jenkins-test-harness/
 
 * See [GitHub Releases](https://github.com/jenkinsci/jenkins-test-harness/releases).
 * For releases before `2.49`, see [this archive](./docs/CHANGELOG-OLD.md)
+
+## Fast-tests opt-in filter
+
+A JUnit Platform PostDiscoveryFilter is provided to allow excluding tests that boot a real Jenkins instance
+(i.e., tests annotated with `@WithJenkins` or `@WithJenkinsConfiguredWithCode`). To enable exclusion set
+`-Djenkins.tests.excludeJenkins=true` in Surefire/Failsafe when running tests (opt-in). By default the filter is disabled.
--- a/README.md
+++ b/README.md
@@ -18,5 +18,6 @@ See https://javadoc.jenkins.io/component/jenkins-test-harness/
 ## Fast-tests opt-in filter
 
 A JUnit Platform PostDiscoveryFilter is provided to allow excluding tests that boot a real Jenkins instance
-(i.e., tests annotated with `@WithJenkins` or `@WithJenkinsConfiguredWithCode`). To enable exclusion set
-`-Djenkins.tests.excludeJenkins=true` in Surefire/Failsafe when running tests (opt-in). By default the filter is disabled.
+(i.e., tests annotated with `@WithJenkins` or `@WithJenkinsConfiguredWithCode` whose method declares the
+corresponding rule parameter). To enable exclusion set `-Djenkins.tests.excludeJenkins=true` in Surefire/Failsafe
+when running tests (opt-in). By default the filter is disabled.
--- a/pom.xml
+++ b/pom.xml
@@ -292,6 +292,10 @@ THE SOFTWARE.
     </dependency>
     <dependency>
       <groupId>org.junit.platform</groupId>
+      <artifactId>junit-platform-launcher</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.platform</groupId>
       <artifactId>junit-platform-suite</artifactId>
     </dependency>
     <dependency>
--- a/src/main/java/org/jenkinsci/test/JenkinsBootTestFilter.java
+++ b/src/main/java/org/jenkinsci/test/JenkinsBootTestFilter.java
@@ -1,16 +1,15 @@
 package org.jenkinsci.test;
 
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.util.Optional;
+import org.junit.platform.commons.support.AnnotationSupport;
 import org.junit.platform.engine.FilterResult;
 import org.junit.platform.engine.TestDescriptor;
 import org.junit.platform.engine.TestSource;
 import org.junit.platform.engine.support.descriptor.ClassSource;
 import org.junit.platform.engine.support.descriptor.MethodSource;
-import org.junit.platform.commons.support.AnnotationSupport;
 import org.junit.platform.launcher.PostDiscoveryFilter;
-
-import java.lang.annotation.Annotation;
-import java.lang.reflect.Method;
-import java.util.Optional;
 
 /**
  * PostDiscoveryFilter that excludes tests (when enabled) which are annotated with
--- a/src/main/java/org/jenkinsci/test/JenkinsBootTestFilter.java
+++ b/src/main/java/org/jenkinsci/test/JenkinsBootTestFilter.java
@@ -1,0 +1,117 @@
+package org.jenkinsci.test;
+
+import org.junit.platform.engine.FilterResult;
+import org.junit.platform.engine.TestDescriptor;
+import org.junit.platform.engine.TestSource;
+import org.junit.platform.engine.support.descriptor.ClassSource;
+import org.junit.platform.engine.support.descriptor.MethodSource;
+import org.junit.platform.commons.support.AnnotationSupport;
+import org.junit.platform.launcher.PostDiscoveryFilter;
+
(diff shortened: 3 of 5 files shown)


strangelookingnerd (commented):
While the implementation seems solid I am some what questioning the use case here. There is already plenty of ways to exclude test classes or single tests from execution. You can also enable/disable tests based on properties, environment variables and the likes. Why would you not use the existing options?

rantoniuk:
Fair question! Even within a single plugin repo, the practical difference is that `@WithJenkins`/`@WithJenkinsConfiguredWithCode` are already the source of truth for "this test boots a real Jenkins instance", that's literally what triggers the JenkinsRule bootstrap. Tag-based exclusion, Surefire `<excludes>`, or category markers require a maintainer to maintain a *second*, parallel piece of bookkeeping: tagging every existing Jenkins-booting test up front, and then remembering to tag every new one going forward. Miss a tag and your "fast" test run silently either boots Jenkins anyway or excludes a test that shouldn't be excluded — the two markers can drift out of sync over time.

With this filter there's nothing to keep in sync: a maintainer adds the dependency, sets `-Djenkins.tests.excludeJenkins=true` in a fast-tests profile/CI stage, and it's derived automatically and correctly from the annotation that's already there for functional reasons. New tests just get classified correctly with zero extra effort, and there's no risk of a test being mis-tagged. That's the main win over rolling your own tag/exclude setup per plugin, on top of it also generalizing painlessly across the whole ecosystem (meaning entire jenkinsci org applicable for all plugins, allowing e.g. for groupped-test runs).

strangelookingnerd:
> Fair question! Even within a single plugin repo, the practical difference is that `@WithJenkins`/`@WithJenkinsConfiguredWithCode` are already the source of truth for "this test boots a real Jenkins instance", that's literally what triggers the JenkinsRule bootstrap. 

We should not forget about `RealJenkinsExtension` - might be worth to include it as well. Also I am wondering how this would affect usage of `@WithoutJenkins`. 

> Tag-based exclusion, Surefire `<excludes>`, or category markers require a maintainer to maintain a _second_, parallel piece of bookkeeping: tagging every existing Jenkins-booting test up front, and then remembering to tag every new one going forward. Miss a tag and your "fast" test run silently either boots Jenkins anyway or excludes a test that shouldn't be excluded — the two markers can drift out of sync over time.

On that note I’m wondering if it would be better / possible to automatically add a tag when using these annotations. Or add a `@DisabledIf` annotation on-the-fly to check for the proposed property flag. This way one could achieve the same results in a more convenient or standardised manner. If I’m not mistaken skipped tests would at least be reported, while excluded tests are silently ignored.

rantoniuk:
### `RealJenkinsExtension`

`RealJenkinsExtension` is used via `@RegisterExtension` on a field, not an annotation, and it boots a genuinely separate, out-of-process Jenkins JVM (as opposed to `@WithJenkins`'s in-process `JenkinsRule`). A `PostDiscoveryFilter` doing class/method annotation reflection structurally can't pick that up the same way - it would need to scan declared fields for a `RealJenkinsExtension` type instead, which is a different (and a bit less reliable, since a declared field isn't proof it's actually exercised) detection path. I'd treat this as a legitimate follow-up rather than a blocker for the annotation-based case.
If you wish I can open an issue about this as a follow-up.

### `@WithoutJenkins`

This one turned out to be a bit of a landmine, so thanks for raising it. `@WithoutJenkins` is a **JUnit 3/4-only** annotation (`org.jvnet.hudson.test.WithoutJenkins`), honored only by `JenkinsRule.apply(...)` (the `@Rule` path) and `HudsonTestCase`. `JenkinsExtension` (the JUnit5 extension behind `@WithJenkins`) never looks at it at all - for JUnit5, the idiom for "this method doesn't need Jenkins" is simply *not requesting a `JenkinsRule` parameter*, per `JenkinsExtension`'s own javadoc.

That does expose a real gap in the current filter, though: it excludes at the class level whenever `@WithJenkins` is present, without checking whether an individual method actually asks for the `JenkinsRule` parameter. So today a method in a `@WithJenkins` annotated class that doesn't take a `JenkinsRule` argument would still get excluded even though it never boots Jenkins. 

- [x] I'll follow up with a fix so the method-level check mirrors `JenkinsExtension`'s own logic instead of just relying on the class-level annotation.

### Silent exclusion vs. `@DisabledIf`

You're right that this is a real trade-off, not a misunderstanding: a `PostDiscoveryFilter` exclusion removes the test from the discovered tree entirely, so it never shows up as skipped anywhere, whereas a `@DisabledIf`/tag-based approach would report it as skipped and stay visible in CI/IDE output.

I went with `PostDiscoveryFilter` deliberately:
- It's genuinely zero-cost and zero-touch when disabled (default) — no tagging, no per-test annotation to add, nothing to keep in sync with the harness annotations that already exist for functional reasons.
- Applying `@DisabledIf` "on the fly" from a filter would be roughly equivalent mechanically (still a discovery-time hook), but I do like your point that reporting beats silence.

Given that, maybe there's a middle ground: keep the `PostDiscoveryFilter` (so it stays a single drop-in dependency, no per-test annotations needed), but have it log a summary at the end of the run (count of excluded tests, maybe which classes) so the exclusion isn't completely invisible even if individual tests don't show up as "skipped" in the report. Open to other ideas if visibility is a hard requirement for you.

- [x] I will also update the PR description itself to reflect the _Why_ I think this can be useful.

